PR #5, opened 2017-04-28
Added elasticsearch client and test. DSDEGP-1231

kshakir:
### Description

Elasticsearch status wired into the /health endpoint.
Config values, including the elasticsearch hosts, may be specified via environment variables.
Using elastic4s-http for the elasticsearch client.
Using testcontainers-scala for the elasticsearch client tests.
Step towards the nightly tests against a dev elasticsearch for DSDEGP-1231. A following PR will contain the python code to exercise these endpoints.
Includes workaround for docker-java issue related to reading macOS configs.

----

* Review cycle:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -14,6 +14,7 @@ val ClioVersion = "0.0.1"
 // Common version numbers of dependencies
 val AkkaHttpVersion = "10.0.5"
 val CirceVersion = "0.7.0"
+val Elastic4sVersion = "5.3.2"
 
 // Dependencies
 libraryDependencies ++= Seq(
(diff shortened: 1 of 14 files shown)


tbl3rd (approved):
sbt assembly and test failed looking for a running docker image, which I also couldn't build because of the same exception, but I assume I'm just doing something wrong.

kshakir:
@tbl3rd Grab the latest and give it another shot please? I traced down a workaround that fixes the problem for now.

tbl3rd:
That looks good.
+    sys.props += "DOCKER_CONFIG" -> "/dev/null/workaround/docker-java/issues/806"
looks bizarre.  Does this just cleverly "break" DOCKER_CONFIG so it gets a default?

kshakir:
Yep. See: https://github.com/docker-java/docker-java/blob/f43f174f64e88cdffa7259e3075671681f57c1fa/src/main/java/com/github/dockerjava/core/AuthConfigFile.java#L104-L159
1. If the config file doesn't exist, `AuthConfigFile.loadConfig` just returns an empty config.
2. Otherwise, it tries to parse the file using a Jackson mapping. In the case of issue 806, this is where Jackson fails because of an "unexpected" json element.
3. But `AuthConfigFile.loadConfig` assumes any exceptions are due to the wrong file type, eventually then trying to parse the config file with a really old `name=value` parser.

The "workaround" is to just change where the config file is being searched. As the config doesn't exist, option 1. above is triggered and the docker pull runs just fine.

tbl3rd:
... and type-safe AF. ;-0

kshakir:
Heh, I'm so used to reading scala now, that I forget that `… += … -> …` may not mean anything to most folks.

Squashing and merging unless there are other comments.

PR #7, opened 2017-05-24
First pass at creating an Elasticsearch index. DSDEGP-1232

kshakir:
### Description

Added creation of a readgroups bam index.
While adding a spec for creation of the ES index, added specs for the services.
The HTTP server now waits to startup until the ES indexes are populated.
The server startup and shutdown are recorded internally with additional statuses beyond "OK".
Bumped up dependency versions while addressing an Akka-HTTP warning with 10.0.0.5.
While updating dependency versions, broke dependencies and other sections of build.sbt into separate project files.
DSDEGP-1232

----

* Review cycle:

### DIFF
--- a/src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
+++ b/src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
@@ -1,0 +1,47 @@
+package org.broadinstitute.clio.dataaccess
+
+import org.broadinstitute.clio.model.ServerStatusInfo
+
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+  * Stores the server status in a synchronized mutable variable.
+  */
+class CachedServerStatusDAO private() extends ServerStatusDAO {
+  private val mutex = new Object()
+  private var currentStatus: ServerStatusInfo = ServerStatusInfo.NotStarted
+
+  override def setStatus(status: ServerStatusInfo): Future[Unit] = Future.fromTry(
+    Try {
+      mutex synchronized {
+        import CachedServerStatusDAO.level
+        if (level(currentStatus) < level(status))
+          currentStatus = status
+      }
+    }
+  )
+
+  override def getStatus: Future[ServerStatusInfo] = Future.fromTry(
+    Try {
+      mutex synchronized currentStatus
+    }
+  )
+}
+
+object CachedServerStatusDAO {
+  def apply(): CachedServerStatusDAO = {
+    new CachedServerStatusDAO()
+  }
+
+  private def level(status: ServerStatusInfo): Int = {
+    status match {
+      case ServerStatusInfo.NotStarted => 0
+      case ServerStatusInfo.Starting => 1
+      case ServerStatusInfo.Started => 2
+      case ServerStatusInfo.ShuttingDown => 3
+      case ServerStatusInfo.ShutDown => 4
+      case unknown => throw new RuntimeException(s"unknown status $unknown")
+    }
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -1,131 +1,29 @@
+import org.broadinstitute.clio.sbt._
+
 // ## sbt settings
 // For more info, see: https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509
 
-// Set the scala version used by the project. sbt version set in build.properties.
-scalaVersion := "2.12.2"
-
-// Project name
-name := "clio"
-// Project organization, stored in the assembly jar's META-INF/MANIFEST.MF
-organization := "org.broadinstitute"
-// This version number if this is a release commit, or the upcoming version number if this is a snapshot.
-val ClioVersion = "0.0.1"
-
-// Common version numbers of dependencies
-val AkkaHttpVersion = "10.0.5"
-val CirceVersion = "0.7.0"
-val Elastic4sVersion = "5.4.0"
-
-// Dependencies
-libraryDependencies ++= Seq(
-  "com.iheart" %% "ficus" % "1.4.0",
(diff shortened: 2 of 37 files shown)


src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import org.broadinstitute.clio.model.ServerStatusInfo
+
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+  * Stores the server status in a synchronized mutable variable.
+  */
+class CachedServerStatusDAO private() extends ServerStatusDAO {
```
pshapiro4broad, line 11: @rtitle is currently adding (or done adding?) server status to rawls. It might be worth looking at what he's done, to avoid work duplication, even if we don't have a way to share code across projects.

src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
```diff
@@ -0,0 +1,47 @@
...
+  private val mutex = new Object()
+  private var currentStatus: ServerStatusInfo = ServerStatusInfo.NotStarted
+
+  override def setStatus(status: ServerStatusInfo): Future[Unit] = Future.fromTry(
+    Try {
+      mutex synchronized {
+        import CachedServerStatusDAO.level
+        if (level(currentStatus) < level(status))
```
pshapiro4broad, line 19: It seems like you'd want an abstraction here, e.g. `canTransitionTo()` on `ServerStatusInfo`, in case the rule for transitioning status change or become more complicated.

src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
```diff
@@ -0,0 +1,47 @@
...
+}
+
+object CachedServerStatusDAO {
+  def apply(): CachedServerStatusDAO = {
+    new CachedServerStatusDAO()
+  }
+
+  private def level(status: ServerStatusInfo): Int = {
+    status match {
+      case ServerStatusInfo.NotStarted => 0
```
pshapiro4broad, line 39: Would it be possible to store this information directly in `ServerStatusInfo`, e.g. as per-instance state? Doing that avoids the need to handle the "unknown" case and prevents the situation where a user adds a new status and forgets to add it here.

pshapiro4broad (commented):
Would it be worth creating an abstraction layer that separates elasticsearch from our code? That way you can separate the functionality that elasticsearch provides (which is the abstraction layer), and elasticsearch itself (which is the implementation layer). Part of the reason for doing this is so we can provide a name for the class that describes its function in our code, instead of using the name of a product, which by itself may not have that meaning.

src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
kshakir: I'll ping Rob. Thanks!

src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
kshakir: I wrestled with this one a bit. Ultimately I decided to follow the firecloud-orchestration [DEVNOTES](https://github.com/broadinstitute/firecloud-orchestration/blob/63dcb0be65b2fe69234b0e139c4ce6e57bf90087/DEVNOTES.md), that are copied into the clio [wiki](https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509#Clio(MetadataManager)-PackageStructureandSingleResponsibilities).

> - model
>     -  model classes. Case classes if possible. Little-to-no logic implementations; things like toString allowed

Thus the `ServerStatusInfo` model objects passed around do not contain any business logic, not even the ability to transition from one "level" to another.

kshakir:
Re:

> Would it be worth creating an abstraction layer that separates elasticsearch from our code?

As of the latest commit, all ES specific implementation details are _only_ in the HttpElasticsearchDAO. The abstract interface ElasticsearchDAO uses concepts from ES, but these could easily be implemented for any other another data storage type, in theory. In memory, RDBMS, filesystems, etc. would all work.

However, there are still concepts in the model of an "Elasticsearch specific index". For example, in addition to the index-name, there is also an [index-type](https://www.elastic.co/blog/index-vs-type).

That said, all this could be more clearly defined in Clio. I am notoriously bad an naming, and happy to discuss alternate naming strategies. FYI- I chose the names peeking at the Firecloud Orchestration code. They seem to name the [abstract DAOs](https://github.com/broadinstitute/firecloud-orchestration/tree/b42c1bf37f162c469fc38cc5d69cbbd401610663/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess) based on the type of client, not the function of the DAO.

src/main/scala/org/broadinstitute/clio/dataaccess/CachedServerStatusDAO.scala
rtitle: Hi - I haven't fully looked at this PR (or even know what this repo is for :)), but my [Rawls PR](https://github.com/broadinstitute/rawls/pull/703) is more about asynchronous monitoring of subsystems to provide health checks. It seems like this is more about lifecycle management? So maybe slightly different use cases, though I'd love to discuss the possibility of sharing code.

As an aside, I've used [Guava service](https://github.com/google/guava/wiki/ServiceExplained) before as a way to manage services which have a lifecycle/state and found it pretty nice to use (granted this was in a Java project).

kshakir:
Verbal approval received. Leaving commits separated by what they contribute to the codebase.

PR #11, opened 2017-06-09
DSDEGP-1275 stub in an /authorization endpoint.

tbl3rd:
### Description

----

- [X] **Submitter**: Include the JIRA issue number in the PR description
- [X] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [X] **Submitter**: Ensure that you have added or updated tests
- [X] **Submitter**: If you're adding/switching libraries or otherwise changing the design, update the [design documentation](https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509).
* Review cycle:


tbl3rd:
This PR does not authorize against all requests against a whitelist.
It just implements an authorizer to pass to the authorize() directive
and wraps a stub /authorization endpoint with it, then adds accept
and reject tests to validate the OIDC headers.

Auth.org
```diff
@@ -0,0 +1,66 @@
+#+TITLE: Clio Authentication and Authorization
```
pshapiro4broad, line 1: What is this file? It looks like documentation but if so I'm not sure why it's called `Auth.org` and is stored at the root of the clio repo.

src/main/scala/org/broadinstitute/clio/model/authorization.scala
```diff
@@ -0,0 +1,65 @@
+package org.broadinstitute.clio.model
+
+import akka.http.scaladsl.model.headers.{ModeledCustomHeader, ModeledCustomHeaderCompanion}
+
+import scala.util.Try
+
+
+trait OidcHeader {
```
pshapiro4broad: I'm not sure what our conventions are in Scala, but I didn't expect to see a class named `OidcHeader` in a file called `authorization.scala`. I don't know why `Oidc` is; is that something specific to this implementation or a general concept about authorization?

There should be comments here too.

src/main/scala/org/broadinstitute/clio/service/AuthorizationService.scala
```diff
@@ -0,0 +1,34 @@
...
+
+
+case class AuthorizationInfo(token: String, expiration: String, email: String, id: String)
+
+class AuthorizationService() {
+
+  def authorize(context: RequestContext): Boolean = {
+    val pairs = for (header <- context.request.headers) yield
+      header match {
+        case OidcAccessToken(token)      => 0 -> token
```
pshapiro4broad: It's not clear to me what the numbers here represent. Is it just a way to allow the code to allow the API to provide either `user id` or `sub` in the header? Is this a common idiom in Akka code?

IMO it might be clearer to have the code make a map using strings instead of numbers, e.g. `"token" -> token` and then you don't need `Vector` below to unpack them, you can pull them out directly using a map lookup.

src/main/scala/org/broadinstitute/clio/service/AuthorizationService.scala
```diff
@@ -0,0 +1,34 @@
...
+        case OidcClaimExpiresIn(expires) => 1 -> expires
+        case OidcClaimEmail(email)       => 2 -> email
+        case OidcClaimUserId(id)         => 3 -> id
+        case OidcClaimSub(id)            => 3 -> id
+      }
+    (pairs.size == 4) && (Vector.tabulate(4) { pairs.toMap } match {
+      case Vector(token, expires, email, id) =>
+        AuthorizationService.mock == AuthorizationInfo(token, expires, email, id)
```
pshapiro4broad: I'm unclear what `.mock` is used for here. I would expect it to be a mocked service API for testing but it looks like it contains the actual values from the HTTP response header.

It looks like this is changing the value in `AuthorizationService.mock` but I don't see how that could work since it's declared as a `val`.

src/main/scala/org/broadinstitute/clio/ClioServer.scala
```diff
@@ -32,7 +33,7 @@ object ClioServer
   private val wrapperDirectives: Directive0 = {
     auditRequest & auditResult & completeWithInternalErrorJson & auditException & mapRejectionsToJson
   }
-  private val innerRoutes: Route = concat(statusRoutes)
+  private val innerRoutes: Route = concat(authorizationRoute, statusRoutes)
```
pshapiro4broad, line 36: This adds `/authorization` as new endpoint but it doesn't make authorization required by any other APIs, correct?

src/test/scala/org/broadinstitute/clio/webservice/AuthorizationWebServiceSpec.scala
```diff
@@ -0,0 +1,67 @@
+package org.broadinstitute.clio.webservice
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import akka.http.scaladsl.server.AuthorizationFailedRejection
+import org.scalatest.{FlatSpec, Matchers}
+import org.broadinstitute.clio.model._
+import org.broadinstitute.clio.service._
+
+class AuthWebServiceSpec extends FlatSpec
```
pshapiro4broad: Should the code also test the case where both `sub` and `user id` are provided, since it supports that case?

src/test/scala/org/broadinstitute/clio/webservice/AuthorizationWebServiceSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+    with Matchers with AuthorizationWebService with ScalatestRouteTest {
+
+  behavior of "AuthWebService"
+
+  it should "authorizationRoute" in {
+
+    Get("/authorization").
+      addHeader(OidcClaimExpiresIn(AuthorizationService.mock.expiration)).
```
pshapiro4broad: Why is this code using `AuthorizationService.mock` to provide the value for `expiration`? It looks like any value would work here.

pshapiro4broad (approved):
Many comments but seems OK overall

👍

src/main/scala/org/broadinstitute/clio/model/authorization.scala
kshakir: Re the naming _only_: from the [clio wiki](https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509#Clio(MetadataManager)-Filenaming) and borrowed from the [scala style guide](http://docs.scala-lang.org/style/files.html):

> As a rule, files should contain a single logical compilation unit. … One exception to this guideline is for classes or traits which have companion objects. … These files should be named according to the class, trait or object they contain …
> All multi-unit files should be given camelCase names with a lower-case first letter. This is a very important convention. It differentiates multi- from single-unit files, greatly easing the process of finding declarations.

src/test/scala/org/broadinstitute/clio/webservice/AuthorizationWebServiceSpec.scala
tbl3rd: Any value would work here now.
But "AuthWebService" should be "AuthorizationWebService".

src/test/scala/org/broadinstitute/clio/webservice/AuthorizationWebServiceSpec.scala
tbl3rd: I don't know.  Maybe.

src/main/scala/org/broadinstitute/clio/service/AuthorizationService.scala
tbl3rd: We don't have an authentication service yet.
This is really just plumbing now.

src/main/scala/org/broadinstitute/clio/service/AuthorizationService.scala
tbl3rd: The numbers just represent parameters to AuthorizationInfo.
I hope there is a solution better than yours (and mine).

src/main/scala/org/broadinstitute/clio/model/authorization.scala
tbl3rd: It used to have more authy named things in it.
So I should rename it 'oidcHeaders.scala'?

src/main/scala/org/broadinstitute/clio/ClioServer.scala
tbl3rd: Correct.

Auth.org
tbl3rd: It just documents my current understanding.
It will wind up in the README, Confluence,
or deleted eventually.

src/main/scala/org/broadinstitute/clio/ClioServer.scala
tbl3rd: The /authorization route is there for testing.
It isn't even exported as an endpoint yet --
though more by default than by design.

src/test/scala/org/broadinstitute/clio/webservice/AuthorizationWebServiceSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    Get("/authorization").
+      addHeader(OidcAccessToken(AuthorizationService.mock.token)).
+      addHeader(OidcClaimExpiresIn(AuthorizationService.mock.expiration)).
+      addHeader(OidcClaimEmail(AuthorizationService.mock.email)).
+      addHeader(OidcClaimUserId(AuthorizationService.mock.id)) ~>
+    authorizationRoute ~> check {
+      status shouldEqual StatusCodes.OK
+    }
+  }
```
tbl3rd: It seems to me that the AuthorizationService is more like the AuditService
than like the StatusService, and it should probably be tested similarly.

PR #17, opened 2017-07-06
Better json handling and cleanup. DSDEGP-1070

kshakir:
### Description

Better json handling.
Using instances of enumeratum.Enum instead of String constants.
Json decoders no longer allow extra object keys from web service requests and elasticsearch responses.
Json printers no longer output scala.None as null in web service responses and elasticsearch requests.
Json printers print object keys with snake_case names.
Moved existing models to server.
Implicit parameters have full names instead of abbrs.
Renamed ElasticsearchDAO to SearchDAO.
Refactored the HttpElasticsearcDAO implementation.

The current models are server only. There are forthcoming DTOs that may be shared between the existing clio-server and a future clio-client.

DSDEGP-1070

----

* Review cycle:

### DIFF
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
@@ -1,0 +1,112 @@
+package org.broadinstitute.clio.server.dataaccess
+
+import akka.pattern.after
+import org.broadinstitute.clio.server.ClioServerConfig
+import org.broadinstitute.clio.server.dataaccess.elasticsearch.ElasticsearchIndex
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+trait SystemElasticsearchDAO extends SearchDAO { this: HttpElasticsearchDAO =>
+
+  override def checkOk: Future[Unit] = {
+    getClusterHealth transform {
+      case Success(health) if health.status == "green" => Success(())
+      case Success(health) =>
+        val message = s"Health status was not green: $health"
+        logger.error(message)
+        Failure(new RuntimeException(message))
+      case Failure(exception) =>
+        logger.error(
+          s"Error while getting Elasticsearch cluster status from $hostsString",
+          exception
+        )
+        Failure(exception)
+    }
+  }
+
+  override def initialize(): Future[Unit] = {
+    for {
+      _ <- waitForSearchReady()
+      _ <- createOrUpdateIndex(ElasticsearchIndex.ReadGroup)
+    } yield ()
+  }
+
+  override def close(): Future[Unit] = {
+    Future {
+      closeClient()
+    }
+  }
+
+  /** If an index with the name doesn't exist then create it, and then separately update the fields on the index. */
+  private def createOrUpdateIndex(index: ElasticsearchIndex[_]): Future[Unit] = {
+    for {
+      exists <- existsIndexType(index)
+      _ <- if (exists) Future.successful(()) else createIndexType(index)
+      _ <- updateFieldDefinitions(index)
+    } yield ()
+  }
+
+  private def isReady: Future[Boolean] = {
+    getClusterHealth transform {
+      case Success(health) =>
+        if (ClioServerConfig.Elasticsearch.readinessColors.contains(
+              health.status
+            )) {
+          Success(true)
+        } else {
+          logger.debug(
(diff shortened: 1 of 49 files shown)


broadinst:
Can one of the admins verify this patch?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
```diff
@@ -0,0 +1,111 @@
...
+          exception
+        )
+        Failure(exception)
+    }
+  }
+
+  override def initialize(): Future[Unit] = {
+    for {
+      _ <- waitForSearchReady()
+      _ <- createOrUpdateIndex(ElasticsearchIndex.ReadGroup)
```
tbl3rd, line 31: Verify at startup time that all the index types supported are available?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -62,8 +58,7 @@ class ServerService private (
   private[service] def startup(): Future[Unit] = {
     for {
       _ <- serverStatusDAO.setStatus(ServerStatusInfo.Starting)
-      _ <- waitForElasticsearchReady()
-      _ <- createOrUpdateIndices()
+      _ <- searchDAO.initialize()
```
tbl3rd, line 61: Why isn't initialize() also named startup(), or the other way round?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
```diff
@@ -0,0 +1,111 @@
+package org.broadinstitute.clio.server.dataaccess
+
+import akka.pattern.after
+import org.broadinstitute.clio.server.ClioServerConfig
+import org.broadinstitute.clio.server.dataaccess.elasticsearch.ElasticsearchIndex
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+trait SystemElasticsearchDAO extends SearchDAO { this: HttpElasticsearchDAO =>
```
tbl3rd, line 10: I don't understand this at all.
(I don't really understand ElasticSearchContainer either.)
http://docs.scala-lang.org/tutorials/tour/self-types.html
http://www.cakesolutions.net/teamblogs/2011/12/19/cake-pattern-in-depth
http://jonasboner.com/real-world-scala-dependency-injection-di/
It allows SystemElasticsearchDAO to call getClusterHealth and so on?
How did you arrive at this design?

tbl3rd (approved):
Have not got through all of it yet.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
kshakir: Hopefully discussed in person, still:

> It allows SystemElasticsearchDAO to call getClusterHealth and so on?

Yes, it specifies that an SystemElasticsearchDAO may only be used in a place where the trait is mixed in with an HttpElasticsearchDAO.

Meanwhile, HttpElasticsearchDAO mixes in SystemElasticsearchDAO, so the contract is satisfied.

> How did you arrive at this design?

I arrived at this pattern because I wanted to start separating different parts of a massive HttpElasticsearchDAO into separate traits. The underlying SearchDAO interface/trait will eventually contain dozens of abstract methods. Instead of having all the concrete methods implemented in one massive class, I'd rather have them grouped into different mixins.

H/T also to the Firecloud-Orchestration [design](https://broadinstitute.atlassian.net/wiki/spaces/DSDEGP/pages/114531509/Clio+Metadata+Manager#Clio(MetadataManager)-PackageStructureandSingleResponsibilities).

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
kshakir: Yep

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
kshakir: The ES server must already be running, but may be empty. `initialize()` will create or update the mappings. Meanwhile the akka-http server `startup()` actually starts listening for connections to the server.

kshakir:
Minor doc patches being squashed in, then merging after verbal ok from TBL.

PR #55, opened 2017-08-06
Upgrade scalafmt so it doesn't barf on trailing commas.

danxmoran:
### Description

Scalafmt 1.1.0 has changed the default dialect for its parser to be "lenient" and accept [pretty much everything](https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala#L43), including trailing commas. We weren't able to use 1.1.0 when we first enabled scalafmt because of a bug in the SBT plugin that we use to pull in the formatter. The bug's been fixed in v1.9 of the plugin, so we can bump up both versions and hopefully never see the cryptic scalafmt parsing errors on trailing commas again.

Caveat: trailing commas were only made valid Scala syntax in 2.12.2, and they're only allowed in certain situations, so we're still going to see compile errors for any trailing commas in our SBT code (which runs on 2.10 until SBT 1.0) and a lot of trailing commas in our app code. The errors will be coming from the compiler instead of scalafmt, though, so they'll be tagged with the path of the file causing the error instead of the totally unhelpful `<input>:1234` errors we get now.

Scalafmt 1.1.0 also fixed an off-by-one error in its maximum-line-length calculation, so a few of our files that were living on the edge got reformatted from the version bump.

----

* Review cycle:


jacarey (approved):
Looks good!

PR #57, opened 2017-08-07
DSDEGP-1469: Use `/api/v1/index/action/...` instead of `/api/index/action/v1/...`

tbl3rd:
### Description

Interesting changes in ReadGroupWebService.scala and ReadGroupWebServiceSpec.scala.
Other scala code changes should be just scalafmt fallout.

clio_test.py isolates the new versioning to readgroup_v1_url().

The clio_test.py changes isolate URL construction to 3 functions: elasticsearch_url(), clio_url(), and readgroup_v1_url().

The rest of the changes just use those and factor associations as in test_read_group_mapping().

----

* Review cycle:


clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/ReadGroupWebServiceSpec.scala
```diff
@@ -20,7 +20,7 @@ class ReadGroupWebServiceSpec
   it should "postMetadata with OnPrem location" in {
     val webService = new MockReadGroupWebService()
     Post(
-      "/metadata/v1/barcodeOnPrem/3/libraryOnPrem/OnPrem",
+      "/metadata/barcodeOnPrem/3/libraryOnPrem/OnPrem",
```
pshapiro4broad, line 23: Why isn't this `/v1/metadata/...`?

clio-integration-test/src/main/python/clio_test.py
```diff
@@ -64,68 +71,63 @@ def test_health(allowYellowHealth=False):
...
+                'run_name',
+                'sample_alias',
+                'sample_gender',
+                'sample_id',
+                'sample_lsid',
+                'sample_type',
+                'target_intervals',
+                'ubam_md5',
+                'ubam_path']
+    expected = {k: 'keyword' for k in keywords}
```
pshapiro4broad, line 119: This is an improvement. If the tests were written in Scala than we'd have a hope of getting rid of this list of keywords from the test.

pshapiro4broad (approved):

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/ReadGroupWebService.scala
```diff
@@ -29,12 +31,10 @@ trait ReadGroupWebService {
 
   private[webservice] val postMetadata: Route = {
     pathPrefix("metadata") {
-      pathPrefix("v1") {
-        pathPrefixKeyV1 { key =>
-          post {
-            entity(as[TransferReadGroupV1Metadata]) { metadata =>
-              complete(readGroupService.upsertMetadata(key, metadata))
-            }
+      pathPrefixKeyV1 { key =>
```
danxmoran: Should we rename `pathPrefixKeyV1` to just be `pathPrefixKey`?

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/ReadGroupWebServiceSpec.scala
danxmoran: The `v1` path segment got pulled out of `webService.postMetadata` and into `webService.readGroupRoutes`. Should we use the read group routes value in all of these tests instead of the individual sub-routes?

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/ReadGroupWebService.scala
tbl3rd: Sure.

clio-integration-test/src/main/python/clio_test.py
tbl3rd: I mostly wanted to distinguish the mapping types.
The special cases get lost in the sea of 'string' types.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/ReadGroupWebService.scala
tbl3rd: I vacillated on both of these comments.  It is still a V1 key, matching up with the V1 transfer cases. We could build the routes up in the server like I did in the Python test, but I think it is easier to read the tests with the explicit subroutes. I don't have strong opinions either way unless I missed your point.

danxmoran: Ah, I missed that it's still tied to the v1 transfer class. I think it's OK as-is then.

pshapiro4broad: Yeah I wrote the same comment as @danxmoran but then deleted it after reading the code a bit more.

tbl3rd: Too late! But I'll leave the test code as is.

PR #62, opened 2017-08-10
DSDEGP-1261: clio integration test in jenkins

pshapiro4broad:
### Description

DSDEGP-1261: modify clio integration test to run on jenkins and to use cloud elastic search (not docker elastic search)

This change adds a new sbt task `testDockerJenkins` which runs the same tests as `testDocker` but uses `elasticsearch1.gotc-dev.broadinstitute.org` instead of the docker elastic search host.

I've tested this out by configuring a Jenkins job to build this new task, the output of that build is here: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-integration/11/console

----

* Review cycle:


clio-integration-test/src/test/resources/docker-compose/docker-compose-jenkins.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  clio-jenkins-integration-test:
+    image: broadinstitute/clio-integration-test:${CLIO_DOCKER_TAG}
+    environment:
+      ELASTICSEARCH_HTTP_HOST: elasticsearch1.gotc-dev.broadinstitute.org
+    network_mode: "host"
+  clio-server:
+    image: broadinstitute/clio-server:${CLIO_DOCKER_TAG}
+    environment:
+      clio.server.elasticsearch.http-hosts.0.hostname: elasticsearch1.gotc-dev.broadinstitute.org
```
danxmoran: I'm surprised this works with the dots, we had to change it to [pass through JAVA_OPTS](https://github.com/broadinstitute/clio/blob/develop/clio-integration-test/src/test/resources/docker-compose/docker-compose.yml#L12) in the other docker-compose file to get around the dots-in-name problem. Maybe a difference in how shell expansion works between our local Docker versions and whatever runs on Jenkins.

Can you change it to use the JAVA_OPTS style, so we're consistent across both setups and guarded against the shell behavior changing out from under us?

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestRunner.scala
```diff
@@ -76,6 +78,35 @@ class ClioIntegrationTestRunner(testClassesDirectory: File,
     *
     * @return The exit code of the integration test.
     */
+  private def tryDockerComposeJenkinsTest(): Try[Int] = Try {
+    runCommandAsync(
```
danxmoran: I think we can make this work with a single docker-compose call:
```
$ docker-compose --file docker-compose-jenkins.yml up
```
You'll need to update the compose file to declare that the integration-test container depends on the clio container, so it gets the setup ordering right.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestRunner.scala
```diff
@@ -132,6 +163,25 @@ class ClioIntegrationTestRunner(testClassesDirectory: File,
...
+  private def runCommandAsync(command: Seq[String],
+                              cwd: File,
+                              extraEnv: Seq[(String, String)],
+                              logFunction: String => Unit): Unit = {
+    import scala.sys.process._
+    val process = Process(command, cwd, extraEnv: _*)
+    val processLogger = ProcessLogger(logFunction)
+    process run processLogger
+    ()
+  }
```
danxmoran: If we do keep separate docker-compose calls, I think it'd be better to have this function return a `Future` instead of `Unit`. That way we can detect when the command fails, vs this which is more fire-and-forget.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
```diff
@@ -13,20 +13,30 @@ object ClioIntegrationTestSettings extends AutoPlugin {
...
-    new ClioIntegrationTestRunner(testClassesDirectory, clioVersion, log).run()
+      // Wait for task test:compile, and the task to create the logs
+      (compile in Test).value
+      val log = streams.value.log
+
+      new ClioIntegrationTestRunner(testClassesDirectory,
+                                    clioVersion,
+                                    jenkinsTest,
+                                    log)
+        .run()
```
danxmoran, line 34: `scalafmt` why

project/src/main/scala/org/broadinstitute/clio/sbt/Docker.scala
```diff
@@ -58,7 +58,7 @@ object Docker {
       workDir("/app")
       copy(pythonSource / "requirements.txt", "./")
       run("pip", "install", "-r", "requirements.txt")
-      entryPoint("pytest", "-vv", "-s")
+      entryPoint("pytest", "-s")
```
danxmoran, line 61: Hooray for less noisy output :tada:

clio-integration-test/src/test/resources/docker-compose/docker-compose-jenkins.yml
```diff
@@ -0,0 +1,14 @@
+version: '2'
+
+services:
+  clio-jenkins-integration-test:
+    image: broadinstitute/clio-integration-test:${CLIO_DOCKER_TAG}
+    environment:
+      ELASTICSEARCH_HTTP_HOST: elasticsearch1.gotc-dev.broadinstitute.org
+    network_mode: "host"
+  clio-server:
+    image: broadinstitute/clio-server:${CLIO_DOCKER_TAG}
```
danxmoran: Rather than declaring all the parameters for the server container here in addition to in the other docker-compose file, could we bust out the definition to a separate YAML and then import it here [like we do for elasticsearch](https://github.com/broadinstitute/clio/blob/develop/clio-integration-test/src/test/resources/docker-compose/docker-compose.yml#L15)? Same for the integration-test-runner container too.

danxmoran (changes requested):
Some requests / suggestions! Also a higher-level question for my own understanding: Why not also hit the clio-server running in dev in addition to the elasticsearch cluster running there, rather than spin up a local clio-server container?

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestRunner.scala
danxmoran: ^ our deploy config uses this pattern, so it's a good example

clio-integration-test/src/test/resources/docker-compose/docker-compose-jenkins.yml
pshapiro4broad: This worked for me both locally and on the server, before I change it let's try to figure out why it's working for me, and maybe see if it can work for others too.

danxmoran: Want to huddle after the DSDE meetup? I'm curious to get to the bottom of this too.

pshapiro4broad:
> Why not also hit the clio-server running in dev in addition to the elasticsearch cluster running there, rather than spin up a local clio-server container?

This way we can test clio on a branch without having to deploy it first. This seems useful to me but maybe in practice we don't need to do this?

danxmoran:
> This seems useful to me but maybe in practice we don't need to do this?

They both seem useful to me :smile: When I think of testing a branch I just think of running everything in isolation, vs. running in an environment when I'd want as many components as possible to be running in that environment. Getting the tests to work against both local and deployed Clio will be pretty complicated because of the authentication layer on the deployed machines, so no worries for now.

clio-integration-test/src/test/resources/docker-compose/docker-compose-jenkins.yml
pshapiro4broad: The configs are different so I don't think that will work. E.g. this integration-test service has a different ELASTICSEARCH_HTTP_HOST and requires `network_mode: "host"`.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestRunner.scala
pshapiro4broad: Using `up` doesn't work here, we can talk in person about this more.

danxmoran: Thanks for the in-person explanation. Can you add a comment explaining why we have to do it this way?

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestRunner.scala
danxmoran: From in-person: returning `Process` is good too

danxmoran (commented):
Deleted comments made outdated by the huddle. Thanks for talking me through it!

danxmoran (approved):
Works for me locally now :shipit:

PR #74, opened 2017-08-18
DSDEGP-1473 Split out elasticsearch models into a sub-project, with fallout

danxmoran:
### Description

Part two of porting our integration tests to Scala (part one is #73), this restructures our SBT sub-projects so that the integration-test project will be able to access our elasticsearch index definitions. The index definitions use our shapeless mappers, so rather than have the dataaccess-model project depend on transfer-model, I also split out the shapeless stuff into the truly generic clio-util.

----

### Checklist (never delete this)

Never delete this, it is our record that procedure was followed. If you find that for whatever reason one of the checklist points doesn't apply to your PR, you can leave it unchecked but please add an explanation below.

- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: Ensure that you have added or updated tests
- [x] **Submitter**: If you're adding/switching libraries or otherwise changing the design, update the [design documentation](https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509).
- [x] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * Add notes on what you've tested
* Review cycle:
  * Team may comment on PR at will
  * **Reviewer assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to reviewer** for further feedback
- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Squash commits and merge to develop
- [ ] **Submitter**: Delete branch after merge
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [ ] **Submitter**: Move the JIRA issue to QA once this checklist is completed

PR #75, opened 2017-08-18
DSDEGP-1473 Rewrite ITs in Scala using clio-client

danxmoran:
### Description

Final part of the Python -> Scala integration test rewrite! Not for merging until after #74.

----

* Review cycle:


danxmoran:
Crucible version of the review is here: http://crucible.gotc-dev.broadinstitute.org:8060/cru/CR-17

danxmoran:
Jenkins is hitting permissions errors when trying to docker-pull images from testcontainers, so this is on hold until I can figure it out.

danxmoran:
Got Jenkins working! Successful run of ITs is here: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-integration/72/console

build.sbt
```diff
@@ -114,14 +114,14 @@ lazy val `clio-client` = project
...
 
 /**
-  * Integration tests for the clio-server (and eventually the clio-client).
+  * Integration tests for the clio-server and the clio-client.
   */
 lazy val `clio-integration-test` = project
-  .enablePlugins(DockerPlugin)
-  .settings(commonDockerSettings)
-  .settings(dockerfile in docker := Docker.integrationTestDockerFile.value)
+  .dependsOn(`clio-client`, `clio-dataaccess-model`)
```
pshapiro4broad, line 120: not part of this PR but shouldn't that be spelled `clio-data-access-model`?

clio-integration-test/src/it/resources/logback-test.xml
```diff
@@ -0,0 +1,24 @@
+<configuration>
```
pshapiro4broad, line 1: Is there any way to do this without duplicating the logback.xml files? If not, is there a good way to keep all the logback files in sync across the different sbt projects?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+  * The integration spec that tests Dockerized Clio using a local Elasticsearch
+  * cluster also running in Docker.
+  */
+class FullDockerIntegrationSpec
+    extends DockerIntegrationSpec("docker-compose.yml", "local") {
+
+  private val esName = "elasticsearch1_1"
```
danxmoran: Note to self, as found by Trevyn: this needs to be lazy to avoid an initialization error.

pshapiro4broad (approved):
Looks good but am not sure about not using an environment variable for dev/staging/etc. So each different platform needs to be added to clio, vs having jenkins provide this.

clio-integration-test/src/it/resources/logback-test.xml
danxmoran: I think there's a way to do it, will investigate!

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+    val restClient = RestClient.builder(host).build()
+    HttpClient.fromRestClient(restClient)
+  }
+
+  /**
+    * Get a bearer token for hitting the /api route of Clio.
+    *
+    * For now, just calls out to `gcloud` in an external process.
+    */
+  def bearerToken: String = "gcloud auth print-access-token".!!.stripLineEnd
```
tbl3rd, line 76: I'm surprised that shelling out to gcloud in a docker container works.

tbl3rd (approved):
Instructive.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
pshapiro4broad: I believe as part of this checkin the tests no longer run inside a docker container.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+  * @see `ClioIntegrationTestSettings` in the build
+  */
+abstract class DockerIntegrationSpec(composeFile: String, esDescription: String)
+    extends BaseIntegrationSpec(
+      s"Containerized Clio - connecting to $esDescription Elasticsearch"
+    )
+    with ForAllTestContainer
+    with IntegrationSuite {
+
+  private val clioName = "clio-server_1"
```
pshapiro4broad: This introduces a dependency on the name in a docker-compose.yml file, correct? If there's no way to avoid this kind of cross-source dependency, it needs to be commented so it's clear what needs to be changed if it's renamed.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
```diff
@@ -1,42 +1,90 @@
...
+  lazy val resetItLogs: Initialize[Task[File]] = Def.task {
+    val logDir = itLogDir.value
+    IO.delete(logDir)
+    IO.createDirectory(logDir)
+    /*
+     * Our Dockerized integration tests tail the clio log to check
+     * for the "started" message before starting tests, so we ensure
+     * it exists here.
+     */
+    val clioLog = logDir / "clio-server" / "clio.log"
```
pshapiro4broad: It looks like this path is a magic string that's repeated in more than once place. If so, it should be in a constant to avoid duplication.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
```diff
@@ -1,42 +1,90 @@
...
-      new ClioIntegrationTestRunner(testClassesDirectory,
-                                    clioVersion,
-                                    jenkinsTest,
-                                    log)
-        .run()
-    }
+object ClioIntegrationTestSettings extends ClioIntegrationTestKeys {
+
+  /** Directory to which all container logs will be sent during Dockerized integration tests. */
+  lazy val itLogDir: Initialize[File] = Def.setting {
```
pshapiro4broad: Since these fields are already scoped inside `ClioIntegrationTestSettings` it doesn't seem like they need an `it` name as well.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
```diff
@@ -1,42 +1,90 @@
...
+         *
+         * Integration tests should be explicitly run with the keys defined below, or "it:test".
+         */
+        test := {},
+        /*
+         * Define aliases for "it:testOnly XYZ" to make running integration tests against
+         * different setups / environments more convenient.
+         */
+        testDocker := (testOnly in IntegrationTest)
+          .toTask(" *FullDockerIntegrationSpec")
```
pshapiro4broad: Does this depend on a class name to work correctly? If so, it should use a type reference to look up the name, instead of using a hardcoded string.

pshapiro4broad:
Why is the source path `src/it/resources` and `src/it/scala`? The root `it` seems unnecessary here; I would expect to see `src/test/resources` and `src/test/scala` (or `src/main`, `src/resources`).

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
danxmoran: It does depend on the class name, but I don't think it can use a type reference (if I understand what you mean). The build code is compiled as an entirely separate project from the app code within SBT, and the two run binary incompatible versions of Scala (SBT is stuck on 2.10). If we really wanted to lock down one location for the spec names across the build and app code I think we'd need to auto-generate the source files from SBT with the expected names, and I'm not sure we want to head down that road.

At the least, I'll add big warning comments above the test classes saying that if they're renamed, these task definitions also need to be modified.

Alternatively I could just scrap these tasks. Currently, they're just serving as aliases for the `it:testOnly` input task that SBT provides, so anyone who's gotten the muscle memory of typing `testDocker` doesn't have to change their workflow. If we don't think they're worth it I can toss them out, but I think they're a nice convenience.

danxmoran:
> but am not sure about not using an environment variable for dev/staging/etc. So each different
> platform needs to be added to clio, vs having jenkins provide this.

Talked in person, but for any future archeologists: SBT makes it hard to set environment variables once you've dropped into its shell, so I went with separate tasks for each test platform instead of one task + an env var. Having separate suites for each platform also lets us run the test suites of multiple platforms with one `it:test` or `it:testOnly` command if we want, but I'm not sure if we'll ever want to do such a thing.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
pshapiro4broad: I don't think it's worth having them just as aliases if it means referencing a class by string literal. Be sure to update `Readme.md` with the new sbt tasks.

danxmoran:
>Why is the source path src/it/resources and src/it/scala? The root it seems unnecessary here; I
> would expect to see src/test/resources and src/test/scala (or src/main, src/resources).

`src/it/scala` is the default source root for the `IntegrationTest` configuration in SBT. If I had used the default `Test` configuration instead of `IntegrationTest`, then every time somebody would run `sbt test` it would run both unit and integration tests because of how SBT's task aggregation works. Putting the integration tests in a separate configuration prevents their aggregation with the unit tests of the other sub-projects. I could have chosen another approach to have the same effect, but this was the way that fell out of SBT's built-in task scoping mechanisms most cleanly.

As for `src/main/scala`, I tried it out but couldn't make it immediately work. SBT provides a bunch of default logic for running the Scalatest runner within the `Test` and `IntegrationTest` configurations. I didn't see the point in copying over all the settings to the default `Compile` configuration.

pshapiro4broad:
OK, didn't know about the `src/it` sbt convention.

clio-integration-test/src/it/resources/logback-test.xml
danxmoran: I couldn't find a quick-and-easy solution for this, so I'm going to file a separate ticket for general configuration cleanup.

build.sbt
danxmoran: Not going to change this now because to do it properly we'd need to rename the directory too, and I don't want to cause more merge conflicts for people working in the elasticsearch code.

PR #76, opened 2017-08-20
Reorg of commands per Phil's review suggestion.

jacarey:
Reorganized commands to be more strongly typed and constructed via argument parser.

----

* Review cycle:


clio-client/src/main/resources/logback.xml
```diff
@@ -2,14 +2,14 @@
 <!--suppress ALL -->
 <configuration>
 
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <appender name="STDERR" class="ch.qos.logback.core.ConsoleAppender">
```
danxmoran, line 5: Does this need a `<target>System.err</target>` as well?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -2,39 +2,104 @@ package org.broadinstitute.clio.client.commands
...
+                s" Response code: ${response.status}"
+            )
+          }
+          isSuccess
+        },
+        ClioClientConfig.responseTimeout
+      )
+    }
+  }
+  case class AddReadGroupBam()(implicit ec: ExecutionContext)
```
danxmoran: Rather than bundling the `execute` functionality up inside the case classes, could we:
1. Make `Command` be an enum using enumeratum
2. Move the logic for executing commands inside `CommandDispatch`

I think it'd be an improvement because:
1. `Command` is sealed, so as we move forward with adding commands this file is going to get pretty big if we keep the `execute` logic in here too; moving the logic into `CommandDispatch` would let us break the functionality up into traits / typeclasses that the dispatcher uses
2. With the sub-commands as enums, we won't need to do the weird `apply()` calls in the parser to get their names

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -2,39 +2,104 @@ package org.broadinstitute.clio.client.commands
...
+      extends Command(commandName = "addReadGroupBam") {
+    override def execute(clioWebClient: ClioWebClient,
+                         config: BaseArgs): Boolean = {
+      //parse metadata to validate inputs
+      val json = parse(IoUtil.readMetadata(config.metadataLocation)) match {
```
danxmoran: I think it'd be more idiomatic to roll up `json` and `decoded` using `flatMap`:
```scala
val decodedOrErr = parse(...).flatMap(_.as[TransferReadGroupV1Metadata])
```
Then instead of throwing, we can transform the `Either` into a `Future`:
```scala
decodedOrErr.fold(Future.failed, { decoded => /* call to clioWebClient */ })
```
If we hit a failure, it'll still throw up at the `Await.result` layer.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -2,39 +2,104 @@ package org.broadinstitute.clio.client.commands
...
+          location = config.location.map(s => Location.pathMatcher(s)),
+          lcSet = config.lcSet,
+          project = config.project,
+          sampleAlias = config.sampleAlias,
+          runDateEnd = config.runDateEnd,
+          runDateStart = config.runDateStart
+        )
+      )
+
+      checkResponse(responseFuture)
```
danxmoran: Rather than ending every call to `execute` with a call to `checkResponse`, could we have `execute` return a `Future[HttpResponse]` and then have it be `CommandDispatch` who calls something like `checkResponse(command.execute)`

clio-client/src/main/scala/org/broadinstitute/clio/client/parser/BaseArgs.scala
```diff
@@ -28,8 +31,8 @@ class BaseParser extends scopt.OptionParser[BaseArgs]("clio-client") {
     .action((token, conf) => conf.copy(bearerToken = token))
     .text("A valid bearer token for authentication.")
 
-  cmd(Commands.addReadGroupBam)
-    .action((_, c) => c.copy(command = Some(Commands.addReadGroupBam)))
+  cmd(Commands.AddReadGroupBam.apply().commandName)
```
danxmoran: Here's the weird `apply` usage I was talking about. If our commands were case objects we could just to `AddReadGroupBam.commandName`

clio-client/src/main/scala/org/broadinstitute/clio/client/parser/BaseArgs.scala
```diff
@@ -5,7 +5,9 @@ import java.time.OffsetDateTime
 import org.broadinstitute.clio.client.ClioClientConfig
 import org.broadinstitute.clio.client.commands.Commands
 
-case class BaseArgs(command: Option[String] = None,
+import scala.concurrent.ExecutionContext
+
+case class BaseArgs(command: Option[Commands.Command] = None,
                     bearerToken: String = "",
                     metadataLocation: String = "",
```
danxmoran: Should `metadataLocation` be an `Option` too, since this class is serving as the bucket-of-all-args that might or might not be needed?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
danxmoran: This would also help the integration tests cover more code. In #75 I wrote them to use the web client instead of the top-level clio client so I could get ahold of the `HttpResponse` objects.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -2,39 +2,104 @@ package org.broadinstitute.clio.client.commands
...
+          libraryName = config.libraryName,
+          location = config.location.map(s => Location.pathMatcher(s)),
+          lcSet = config.lcSet,
+          project = config.project,
+          sampleAlias = config.sampleAlias,
+          runDateEnd = config.runDateEnd,
+          runDateStart = config.runDateStart
+        )
+      )
+
```
danxmoran: Between running and checking, I think this command should unmarshal and print the contents of `responseFuture`. Otherwise people who run queries will see a log saying it succeeded, but not the actual query response.

danxmoran (changes requested):
Mainly one architectural question, but looks like the right way to go overall since scopt makes it such a pain to work with subclass instances nested in a case class.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -25,10 +25,9 @@ object ClioClient extends App {
 
 class ClioClient(val webClient: ClioWebClient) {
 
-  val parser = new BaseParser
-
   def execute(args: Array[String])(implicit ec: ExecutionContext): Int = {
-    val success = parser.parse(args, BaseArgs()) match {
+    val parser: BaseParser = new BaseParser
```
pshapiro4broad, line 29: So you have to create a new parser every time you want to call `parse()`? It looks like the way you implemented `BaseParser` results in an immutable scopt parser.

jacarey: When would you need to parse the args more than once ?

pshapiro4broad: We don't, except perhaps in a test. I was worried that the change was required due to a test, in which case I wanted to know if the parser wasn't actually immutable.

I guess on the other hand, this limits the scope of the object to area where it's needed.

jacarey:
@danxmoran Do you mind taking another look over this?

clio-client/src/test/scala/org/broadinstitute/client/commands/QueryReadGroupBamSpec.scala
```diff
@@ -2,41 +2,33 @@ package org.broadinstitute.client.commands
...
-      libraryName = Some(testLibName),
-      location = Some(testLocation),
-      lcSet = Some(testLcSet),
-      project = Some(testProject),
-      sampleAlias = Some(testSampleAlias),
-      runDateEnd = Some(testRunDateEnd),
-      runDateStart = Some(testRunDateStart),
-      bearerToken = testBearer
-    ).execute should be(false)
+    CommandDispatch.checkResponse(QueryWgsUbamCommand.execute(MockClioWebClient.returningInternalError, config)) should be(
```
danxmoran: scalafmt's line-length checking has failed us

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,40 +1,30 @@
...
+import scala.collection.immutable.IndexedSeq
+import scala.concurrent.{ExecutionContext, Future}
+
+sealed trait CommandType extends EnumEntry
+
+object Commands extends Enum[CommandType] {
+  override val values: IndexedSeq[CommandType] = findValues
+
+  case object AddWgsUbam extends CommandType
+  case object QueryWgsUbam extends CommandType
```
danxmoran, line 17: Do we care that these will be PascalCase instead of camelCase on the command line?

danxmoran (approved):
LGTM, with one question

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
jacarey: Yeah I debated this. Not sure I feel strongly one way or the other so I just went with scala naming conventions instead.

jacarey:
ugh conflict on name changes :(

tlangs (approved):
Looks good!

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
tlangs: I'm almost leaning towards snake_case or gnu-case for command line arguments

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

PR #77, opened 2017-08-20
Jc swagger

jacarey:
Added swagger yaml and swagger UI.
I haven't tested with open idc proxy for auth but you should be able to hit it locally by going to /swagger and not authenticating.

----

* Review cycle:


clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/SwaggerDirectives.scala
```diff
@@ -0,0 +1,35 @@
+package org.broadinstitute.clio.server.webservice
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+
+trait SwaggerDirectives {
+
+  lazy val swaggerRoutes: Route = concat(swaggerUiService, swaggerApiDocs)
```
tbl3rd, line 9: Why lazy?

tbl3rd (approved):

clio-server/src/main/resources/swagger/api-docs.yaml
```diff
@@ -0,0 +1,383 @@
...
+    type: "oauth2"
+    authorizationUrl: 'https://accounts.google.com/o/oauth2/auth'
+    flow: "implicit"
+    scopes:
+      openid: "open id authorization"
+      email: "email authorization"
+      profile: "profile authorization"
+
+definitions:
+  TransferReadGroupV1Key:
```
pshapiro4broad: Is there any way to avoid the code duplication with the data model classes?

danxmoran:
Needs updating now that @gbggrant has renamed "read group" -> "wgs ubam" everywhere. Long-term we probably want to use something like [this](https://github.com/swagger-akka-http/swagger-akka-http) to generate the yaml and avoid having to chase the API / types.

jacarey:
I agree with @danxmoran we need a better solution to auto generate these docs, through we ran into issue with annotations when we were using those on early DSDE projects.

clio-server/src/main/resources/swagger/api-docs.yaml
jacarey: Not sure about this one I haven't touched the models yet. It did seem odd that we have a key, metadata and another that is the key + the metadata.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/SwaggerDirectives.scala
jacarey: It seems that all our routes are lazy so I figured someone knew something I didn't. Doesn't seem obviously necessary since I doubt concat take that long to execute.

clio-server/src/main/resources/swagger/api-docs.yaml
```diff
@@ -0,0 +1,383 @@
...
+  description: "This is the API specification for Clio."
+  version: "1.0.0"
+  title: "Clio"
+  contact:
+    email: "greenteam@broadinstitute.org"
+  license:
+    name: "BSD 3-clause"
+    url: "https://opensource.org/licenses/BSD-3-Clause"
+
+host: "localhost:8080"
```
jacarey, line 13: need to pull this from env and make a ctmpl for releases

jacarey: actually looking at firecloud it seems like the sed in server values on deploy

pshapiro4broad (approved):
Yes!

PR #80, opened 2017-08-22
Tl dsdegp 1481 uuids

tlangs:
### Description

Made Clio assign V1 UUIDs to every upsert.

----

* Review cycle:


project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -139,4 +141,8 @@ object Dependencies {
     "org.scalatest" %% "scalatest" % ScalaTestVersion,
     "org.slf4j" % "slf4j-api" % Slf4jVersion
   ).map(_ % IntegrationTest)
+
+  val ThirdPartyDependencies: Seq[MavenRepository] = Seq(
```
danxmoran: Naming nitpick: `ThirdPartyResolvers` would be better.

danxmoran (commented):
Existing code looks good. One request and one question:
1. Can you add a new integration test that adds a bunch of ubams and asserts their returned IDs are all unique?
2. Instead of returning the raw Long, would it be better to make a new transfer model for add-outputs, so the returned number can have a label?

tlangs:
@danxmoran I like the transfer model idea. I just didn't know if it was worth it for just one Long.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -49,7 +50,8 @@ object Dependencies {
     "io.circe" %% "circe-generic" % CirceVersion,
     "io.circe" %% "circe-generic-extras" % CirceVersion,
     "io.circe" %% "circe-parser" % CirceVersion,
-    "org.slf4j" % "slf4j-api" % Slf4jVersion
+    "org.slf4j" % "slf4j-api" % Slf4jVersion,
+    "com.eaio.uuid" % "uuid" % UUIDVersion
```
tbl3rd, line 54: It looks like this shells out to '/sbin/ifconfig -a' for the MAC address,
which is ... not cool.  See .../com/eaio/uuid/UUIDGen.java

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentWgsUbam.scala
```diff
@@ -8,6 +8,7 @@ case class DocumentWgsUbam(flowcellBarcode: String,
                            lane: Int,
                            libraryName: String,
                            location: Location,
+                           clioId: Option[Long],
```
tbl3rd: This should have a not_analyzed string field mapping.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/WgsUbamService.scala
```diff
@@ -18,19 +19,23 @@ class WgsUbamService private (
   def upsertMetadata(
     transferKey: TransferWgsUbamV1Key,
     transferMetadata: TransferWgsUbamV1Metadata
-  ): Future[Unit] = {
+  ): Future[Long] = {
+    val clioId = new UUID().getClockSeqAndNode
```
tbl3rd: This is just the MAC and the seeded sequence number.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
tbl3rd: Ugh.

tbl3rd (changes requested):
The plumbing looks OK, but we need to find another generator for clioId. Use the entire UUID clioId and map it as a not_analyzed string ... while ... we find a better ID implementation.

PR #81, opened 2017-08-23
Fix timeout override in WgsUbamWebServiceSpec (take two)

danxmoran:
### Description

This test is still intermittently failing on Jenkins even with the new implicit timeout, so I dug into it some more and found I was introducing the implicit into the wrong scope / with the wrong type.

Investigation recap:
- The failure we're seeing is coming from [this line](https://github.com/akka/akka-http/blob/40bd7a2d14ecd92eb1527d43a3f56d51e2c17e23/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestResultComponent.scala#L93) within Akka-HTTP-testkit's `RouteTestResult`
- `RouteTestResult` takes a `FiniteDuration` as an explicit parameter.
- The parameter is set at [this line](https://github.com/akka/akka-http/blob/40bd7a2d14ecd92eb1527d43a3f56d51e2c17e23/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala#L154), pulled from an implicit `RouteTestTimeout` parameter to `injectIntoRoute`
- `injectIntoRoute` outputs an implicit `TildeArrow[RequestContext, Future[RouteResult]]` based on its implicit inputs
- The [`~>` method](https://github.com/akka/akka-http/blob/40bd7a2d14ecd92eb1527d43a3f56d51e2c17e23/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTest.scala#L125) takes an implicit `TildeArrow[A, B]`
- So:
  - When we do `Get("/schema") ~> webService.getSchema ~> check`, the `~>` calls will search for an implicit `TildeArrow`
  - Since `injectIntoRoute` is implicit, it'll automatically kick in to produce the required `TildeArrow` instance, pulling its own implicit parameters (including a `RouteTestTimeout`) from the void in the process
  - Putting an implicit `RouteTestTimeout` in the local scope of the `~>` calls will make the implicit calls to `injectIntoRoute` use the larger timeout, avoiding the intermittent 1-second timeout failures

----

* Review cycle:

### DIFF
--- a/clio-server/src/test/resources/application.conf
+++ b/clio-server/src/test/resources/application.conf
@@ -5,3 +5,10 @@ clio.server {
     replicate-indices: false
   }
 }
+
+akka {
+  test {
+    # Scale default timeouts by 3 to avoid intermittent errors in Jenkins
+    timefactor = 3.0
+  }
+}
--- a/clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
+++ b/clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.clio.server.webservice
 
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Route
-import akka.http.scaladsl.testkit.{RouteTestTimeout, ScalatestRouteTest}
+import akka.http.scaladsl.testkit.ScalatestRouteTest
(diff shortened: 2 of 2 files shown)


pshapiro4broad:
If it works it's 👍 by me.

I see that another approach is use the config to apply this change, presumably since you may want to fix this in many tests. What do you think of doing it that way? https://stackoverflow.com/a/42523392

pshapiro4broad (approved):

tbl3rd (approved):

PR #84, opened 2017-08-24
Switch command line parsing to case-app.

jacarey:
Changing our command parsing to use case-app. This allows us to directly parse command line args into our transfer model and generally makes things much cleaner/clearer (IMO)

----

* Review cycle:


jacarey:
This is a bit of a WIP but I wanted to get early feedback.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,14 +1,25 @@
...
+  CommonOptions
+}
+import org.broadinstitute.clio.client.webclient.ClioWebClient
 
 import scala.concurrent.ExecutionContext
 
-object ClioClient extends App {
+object ClioClient extends CommandAppWithPreCommand[CommonOptions, Command] {
+
+  //added to stop Intellij from removing import
```
danxmoran: 😬  ouch

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,84 @@
...
 import scala.concurrent.{ExecutionContext, Future}
 
-sealed trait CommandType extends EnumEntry
-
-object Commands extends Enum[CommandType] {
-  override val values: IndexedSeq[CommandType] = findValues
+object CustomArgParsers {
+  implicit def locationParser: ArgParser[Location] = {
+    ArgParser.instance[Location]("location") { location =>
+      Right(Location.pathMatcher(location))
```
danxmoran: If `location` isn't a valid stringified `Location`, will this become a `Left[SomeException]` when `Location.pathMatcher(location)` throws?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,84 @@
...
+    }
+  }
+  implicit def documentStatusParser: ArgParser[DocumentStatus] = {
+    ArgParser.instance[DocumentStatus]("documentStatus") { documentStatus =>
+      Right(
+        DocumentStatus.namesToValuesMap
+          .getOrElse(documentStatus, DocumentStatus.Normal)
```
danxmoran: Do we want this `getOrElse` here? I think it'll make all invalid documentStatus arguments map to Normal, instead of raising an exception.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,84 @@
...
+object CustomArgParsers {
+  implicit def locationParser: ArgParser[Location] = {
+    ArgParser.instance[Location]("location") { location =>
+      Right(Location.pathMatcher(location))
+    }
+  }
+  implicit def documentStatusParser: ArgParser[DocumentStatus] = {
+    ArgParser.instance[DocumentStatus]("documentStatus") { documentStatus =>
+      Right(
+        DocumentStatus.namesToValuesMap
```
danxmoran: Rather than having two separate parsers for `Location` and `DocumentStatus`, could we have a single `ArgParser[T <: EnumEntry]` that could handle both (and any others we might add)?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,14 +1,25 @@
...
+  CommandDispatch,
+  CommonOptions
+}
+import org.broadinstitute.clio.client.webclient.ClioWebClient
 
 import scala.concurrent.ExecutionContext
 
-object ClioClient extends App {
+object ClioClient extends CommandAppWithPreCommand[CommonOptions, Command] {
+
```
danxmoran: I think we lost the version-reporting. You should be able to `override val appVersion = <however we read the version>`

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,84 @@
...
+          .getOrElse(documentStatus, DocumentStatus.Normal)
+      )
+    }
+  }
+  implicit def offsetDateTimeParser: ArgParser[OffsetDateTime] = {
+    ArgParser.instance[OffsetDateTime]("date") { offsetDateAndTime =>
+      Right(OffsetDateTime.parse(offsetDateAndTime))
```
danxmoran: Same here, if parsing fails does this become a `Left`?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,84 @@
...
-  )
+sealed abstract class Command extends Product with Serializable {
+  def execute(webClient: ClioWebClient, bearerToken: String)(
+    implicit ec: ExecutionContext
+  ): Future[HttpResponse]
+}
 
+final case class QueryWgsUbam(
+  @Recurse
+  transferWgsUbamV1QueryInput: TransferWgsUbamV1QueryInput,
```
jacarey: Since documentation is handled by annotations I'm not entirely sure the best way to add documentation for options coming from the transfer model

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,84 @@
...
+@AppName("Clio Client")
+@ProgName("clio-client")
+final case class CommonOptions(bearerToken: String = "")
 
-  val pathMatcher = Map(
-    AddWgsUbam.toString -> AddWgsUbam,
-    QueryWgsUbam.toString -> QueryWgsUbam
-  )
+sealed abstract class Command extends Product with Serializable {
+  def execute(webClient: ClioWebClient, bearerToken: String)(
```
danxmoran: I'd still prefer to keep the `execute` definitions separated from the `Command` instance definitions, to avoid being forced by `sealed` to put all the implementations in this file.

danxmoran (commented):
Am a fan of the overall direction, I think it'll save us a lot of boilerplate in the long run 👍 

Did you notice any slowdown in compile times?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
jacarey: I can move execute. Though case-app had issues if this base command wasn't sealed for some reason.. I'll poke around a bit.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
jacarey: Yup just added positive cases to see if I could get it working.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -18,20 +29,27 @@ object ClioClient extends App {
     ClioClientConfig.ClioServer.clioServerPort,
```
pshapiro4broad, line 44: Why is `ClioWebClient` created here instead of in the `ClioClient` constructor below?

jacarey: So that we can pass in mock web clients for testing.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -18,20 +28,28 @@ object ClioClient extends App {
...
     ClioClientConfig.ClioServer.clioServerUseHttps
   )
-  val client = new ClioClient(webClient)
 
-  System.exit(client.execute(args))
+  override def run(command: CommandType, remainingArgs: RemainingArgs): Unit = {
```
danxmoran: Should we check that `remainingArgs` is empty here?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -18,20 +28,28 @@ object ClioClient extends App {
...
+    sys.exit(
+      new ClioClient(webClient, command, commonOptions.bearerToken).execute()
+    )
+  }
+
+  override def beforeCommand(options: CommonOptions,
```
danxmoran: Nitpick: it might be clearer how things work to have this method defined right under the `var commonOptions = _` line

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
```diff
@@ -3,31 +3,30 @@ package org.broadinstitute.clio.client.commands
...
-  ): Future[HttpResponse] = {
+  def execute(
+    command: CommandType,
+    webClient: ClioWebClient,
+    bearerToken: String
+  )(implicit ec: ExecutionContext): Future[HttpResponse] = {
     command match {
-      case AddWgsUbam   => AddWgsUbamCommand.execute(webClient, config)
-      case QueryWgsUbam => QueryWgsUbamCommand.execute(webClient, config)
+      case commandType: AddWgsUbam =>
```
danxmoran: Naming nitpick: rather than `commandType`, could it be `add` and `query`?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,54 @@
...
+import caseapp.Recurse
+import caseapp.core.ArgParser
+import org.broadinstitute.clio.transfer.model.{TransferWgsUbamV1Key, TransferWgsUbamV1QueryInput}
+import org.broadinstitute.clio.util.model.{DocumentStatus, Location}
+
+import scala.util.{Failure, Success, Try}
+
+object CustomArgParsers {
+  implicit def locationParser: ArgParser[Location] = {
+    ArgParser.instance[Location]("location") { location =>
```
danxmoran: Same question as before: rather than having separate parsers for every `EnumEntry` type, could we have one general `ArgParser[T <: EnumEntry]`?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,30 +1,54 @@
...
+                            transferWgsUbamV1Key: TransferWgsUbamV1Key)
+  extends CommandType
 
-trait Command {
-  def execute(webClient: ClioWebClient, config: BaseArgs)(
-    implicit ec: ExecutionContext
-  ): Future[HttpResponse]
-}
+final case class QueryWgsUbam(
+                               @Recurse
```
danxmoran: scalafmt why

danxmoran (commented):
I really like it :tada: though I notice when I run it locally (through SBT) without any arguments it seems to hang

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
jacarey: We would still need to have an instance of check in order to know which enum to query. All we get from the arg parser is the string. I can try it though and you can let me know what you think.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
```diff
@@ -3,31 +3,30 @@ package org.broadinstitute.clio.client.commands
...
+        new AddWgsUbamExecutor(commandType)
+          .execute(webClient, bearerToken)
+      case commandType: QueryWgsUbam =>
```
pshapiro4broad: Is a way to make `CommandType` return the correct constructor to call, to avoid using `match` here? Or maybe we aren't trying to avoid using match in our scala code?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Executor.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.clio.client.commands
+
+import akka.http.scaladsl.model.HttpResponse
+import org.broadinstitute.clio.client.webclient.ClioWebClient
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait Executor {
+  def execute(webClient: ClioWebClient, bearerToken: String)(
```
pshapiro4broad: Could we have a more restricted type for `bearerToken`? It seems like we don't need to perform string operations on it between its creation and its usage.

Is there a way to use an implicit to avoid having to pass it around? It will be the same for all method calls made during the lifetime of JVM, correct?

clio-client/src/test/resources/testdata/metadata
```diff
@@ -1,33 +1,33 @@
...
-    "sampleGender" : "sampleGender",
-    "sampleId" : "sampleId",
-    "sampleLsid" : "sampleLsid",
-    "sampleType" : "sampleType",
-    "targetIntervals" : "targetIntervals",
-    "ubamMd5" : "ubamMd5",
-    "ubamPath" : "ubamPath",
-    "ubamSize" : 1
+  "analysis_type" : "analysisType",
+  "bait_intervals" : "analysisType",
```
pshapiro4broad: Before the values were different for each field, now they're all the same. Why was this changed?

pshapiro4broad (approved):
A few comments but looks like an improvement overall.

It seems like we should have at least one test that verifies that command line arguments are interpreted correctly (vs tests that create the structure that contains parsed argument data). Is the intent that the integration tests would be the place for that?

clio-client/src/test/resources/testdata/metadata
jacarey: I think this had to do with how the json was being unmarshalled. I was getting leftover keys on any value that had camel case instead of snake case

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
jacarey: It was suggested that all execute code is handled by the dispatcher since CommandType is a sealed trait and we didn't want the commands themselves to get too big.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Executor.scala
jacarey: yup

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
pshapiro4broad: OK, I remember from the scala class that using `match` for `trait`s was the scala way. Am not sure I agree that it's a good way...

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -39,16 +45,18 @@ object CustomArgParsers {
...
 final case class AddWgsUbam(metadataLocation: String,
                             @Recurse
                             transferWgsUbamV1Key: TransferWgsUbamV1Key)
-  extends CommandType
+    extends CommandType
 
 final case class QueryWgsUbam(
-                               @Recurse
-                               transferWgsUbamV1QueryInput: TransferWgsUbamV1QueryInput,
-                             ) extends CommandType
```
jacarey: scalafmt can't decide what it wants to do here.

jacarey:
not in love with how we are doing DI for testing... any suggestions?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,47 +1,74 @@
...
+      sys.exit(1)
+    }
+  }
+
   implicit val system: ActorSystem = ActorSystem("clio-client")
   import system.dispatcher
   sys.addShutdownHook({ val _ = system.terminate() })
 
-  val webClient = new ClioWebClient(
+  var webClient: ClioWebClient = new ClioWebClient(
```
danxmoran: Accidental `val` -> `var`?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,47 +1,74 @@
...
+    extends CommandAppWithPreCommand[CommonOptions, CommandType]
+    with LazyLogging {
+  override val progName = "clio-client"
+  override val appName = "Clio Client"
+
+  override val appVersion: String = ClioClientConfig.Version.value
+
+  //not sure this is the best way to pass common opts but case-app docs are spotty at best
+  var commonOptions: CommonOptions = _
+  UUIDParser
```
danxmoran: Needing this makes me sad...

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,33 +1,79 @@
...
+    }
+  }
 }
 
-trait Command {
-  def execute(webClient: ClioWebClient, config: BaseArgs, ioUtil: IoUtil)(
-    implicit ec: ExecutionContext,
-  ): Future[HttpResponse]
-}
+final case class CommonOptions(
```
danxmoran: Could weeeeee move these class / type definitions into the `CustomArgParsers` object (and maybe rename it to be more generic)? Then when we import `CustomArgParsers._` and use `CommandType`, the import will be marked as used and we won't need the weird no-op `UUIDParser` statement.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -77,6 +78,8 @@ object Dependencies {
 
   val ClientMainDependencies: Seq[ModuleID] = Seq(
     "ch.qos.logback" % "logback-classic" % LogbackClassicVersion,
+    "com.github.alexarchambault" %% "case-app" % CaseAppVersion,
+    "com.iheart" %% "ficus" % FicusVersion,
```
danxmoran: Nitpick: can you move this below `scopt` to keep the ABC order?

jacarey: scopt shouldn't be needed any more

jacarey: also g comes before i

danxmoran: Yup was asking to move the `com.iheart` below, but moot point with scopt going away

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,47 +1,69 @@
...
+    val client = new ClioClient(commandDispatch)
+
+    client.execute(command) onComplete {
+      case Success(_) => sys.exit(0)
+      case Failure(_) => sys.exit(1)
+    }
   }
 }
 
 class ClioClient(commandDispatch: CommandDispatch) {
```
danxmoran: Not much point to this class now.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -1,33 +1,80 @@
...
+    core.ArgParser.instance[OffsetDateTime]("date") { offsetDateAndTime =>
+      Try(OffsetDateTime.parse(offsetDateAndTime)) match {
+        case Success(value)     => Right(value)
+        case Failure(exception) => Left(exception.getMessage)
+      }
+    }
+  }
 }
+
+sealed trait CommandType
```
danxmoran, line 68: Should the rest of these be moved into `Commands` to mirror the move of `CommonOptions`?

clio-client/src/test/resources/testdata/metadata
```diff
@@ -1,33 +1,33 @@
...
-    "sample_gender" : "sampleGender",
-    "sample_id" : "sampleId",
-    "sample_lsid" : "sampleLsid",
-    "sample_type" : "sampleType",
-    "target_intervals" : "targetIntervals",
-    "ubam_md5" : "ubamMd5",
-    "ubam_path" : "gs://testProject/testSample/ubamPath1.unmapped.bam",
-    "ubam_size" : 1
+  "analysis_type" : "analysisType",
+  "bait_intervals" : "analysisType",
```
danxmoran: This might have been asked in an outdated comment, but were these all set to the same value on purpose?

danxmoran (approved):
Some final questions, but :shipit: 🎉  overall

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
jacarey: aside from giving tests a call execute without working about sys.exit I agree. I'll try to factor out the tests that call execute directly.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
jacarey: I tried that and case-app got mad... I'll see if I can fiddle around with it more

clio-client/src/test/resources/testdata/metadata
jacarey: Nope. Copy/paste... I'll set them to more relevant values.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,47 +1,69 @@
...
+  override def beforeCommand(options: CommonOptions,
+                             remainingArgs: Seq[String]): Unit = {
+    commonOptions = options
+    if (remainingArgs.nonEmpty) {
+      logger.error(s"Found extra arguments: ${remainingArgs.mkString(" ")}")
+      sys.exit(1)
+    }
+  }
+
   implicit val system: ActorSystem = ActorSystem("clio-client")
```
tbl3rd: ActorSystem(progName) ?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,47 +1,54 @@
...
-  val webClient = new ClioWebClient(
+  val webClient: ClioWebClient = new ClioWebClient(
     ClioClientConfig.ClioServer.clioServerHostName,
     ClioClientConfig.ClioServer.clioServerPort,
     ClioClientConfig.ClioServer.clioServerUseHttps
   )
-  val commandDispatch = new CommandDispatch(webClient, IoUtil)
 
-  val client = new ClioClient(commandDispatch)
+  override def run(command: CommandType, remainingArgs: RemainingArgs): Unit = {
```
tbl3rd, line 48: Weird. Where is remainingArgs used, and why doesn't the compiler complain?

jacarey: used now.. but it definitely didn't complain

tbl3rd:
FWIW: I QA this here. https://github.com/broadinstitute/tbl/blob/master/qa/qa-clio-PR-84-jc.org
Posting in the PR because I could not find a Jira ticket.
Oh, I guess this is it. https://broadinstitute.atlassian.net/browse/DSDEGP-1517
Never mind.

PR #85, opened 2017-08-24
Tl dsdegp 1376 move read group bam client

tlangs:
Added a command to move a WgsUbam in clio. I had to rejigger the testing setup and some of the client code so that ScalaTest wouldn't error out when testing asynchronous stuff. I fixed all the tests and everything passes now. Thanks to @danxmoran for all his help with the scala future weirdness.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  override def execute(webClient: ClioWebClient, config: BaseArgs)(
+    implicit ec: ExecutionContext
+  ): Future[HttpResponse] = {
+    for {
+      wgsUbamsResponse <- queryForWgsUbam(webClient, config) withErrorMsg
+        "Could not query the WgsUbam. No files have been moved"
+      firstJson <- getFirstObject(wgsUbamsResponse) withErrorMsg
+        "Could not get only 1 WgsUbam from Clio. No files have been moved"
+      moveWgsUbam <- moveGoogleObject(
```
pshapiro4broad: It would be safer to implement 'move' in three steps:

- copy
- update location in ES
- delete original

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -38,20 +38,62 @@ object IoUtil {
       .map[File](path => path.toFile)
       .forEach(file => file.deleteOnExit())
   }
+
+  def moveGoogleObject(from: String, to: String): Int = {
+    val gs = new GsUtil(None)
+    gs.mv(from, to)
```
pshapiro4broad: Didn't we talk about using the java bindings here instead of shelling out to call gsutil, to avoid the various issues we've run into with gsutil elsewhere?

tlangs: We have, but I think thats best left for another ticket.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
```diff
@@ -2,57 +2,60 @@ package org.broadinstitute.clio.client.commands
...
-    )
+  )(implicit ec: ExecutionContext): Future[Boolean] = {
+    responseFuture.map(response => {
+      val isSuccess = response.status.isSuccess()
+      if (isSuccess) {
+        logger.info(
+          s"Successfully completed command." +
+            s" Response code: ${response.status}"
+        )
+        logger.info(response.toString)
```
danxmoran: Not your code, but since you're changing things here you could move the `logger.info` call out of the if-else (since it's the same in both branches).

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/Commands.scala
```diff
@@ -15,16 +16,19 @@ object Commands extends Enum[CommandType] {
...
+    MoveWgsUbam.toString -> MoveWgsUbam
   )
 
 }
 
 trait Command {
   def execute(webClient: ClioWebClient, config: BaseArgs)(
-    implicit ec: ExecutionContext
+    implicit ec: ExecutionContext,
+    ioUtil: IoUtilTrait
```
danxmoran: I don't think making the `ioUtil` implicit buys us much. It'd be better as an explicit parameter like the web client.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -0,0 +1,135 @@
...
+import org.broadinstitute.clio.util.model.{DocumentStatus, Location}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+object MoveWgsUbamCommand extends Command with LazyLogging {
+
+  implicit val implicitLogger: Logger = logger
+
+  implicit class FutureErrorMessage[A](future: Future[A]) {
```
danxmoran: Naming nitpick: better as `FutureWithErrorMessage`. We're also probably going to want this in multiple spots in the client, so it could be moved out from under the `MoveWgsUbamCommand`.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  override def execute(webClient: ClioWebClient, config: BaseArgs)(
+    implicit ec: ExecutionContext,
+    ioUtil: IoUtilTrait
+  ): Future[HttpResponse] = {
+    for {
+      wgsUbamsResponse <- queryForWgsUbam(webClient, config) withErrorMsg
+        "Could not query the WgsUbam. No files have been moved"
+      firstJson <- getFirstObject(wgsUbamsResponse) withErrorMsg
+        "Could not get only 1 WgsUbam from Clio. No files have been moved"
```
danxmoran: Instead of `Could not get only 1 WgsUbam from Clio`, could we say `Got multiple WgsUbams from Clio`?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -0,0 +1,135 @@
...
+    path: Option[String]
+  )(implicit ioUtil: IoUtilTrait): Future[Unit] = {
+    ioUtil.deleteGoogleObject(path.get) match {
+      case 0 => Future.successful(())
+      case _ =>
+        Future.failed(new Exception("Delete files in google cloud failed"))
+    }
+  }
+
+  private def getFirstObject(ubamsResponseJson: Json): Future[Json] = {
```
danxmoran: I don't think this needs to be its own method; it's simple enough that it could be part of `queryForWgsUbam` above.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -0,0 +1,135 @@
...
+        """The old bam was not able to be deleted. Clio has been updated to point to the new bam.
+          | Please delete the old bam. If this cannot be done, contact Green Team at greenteam@broadinstitute.org
+          | """.stripMargin
+    } yield {
+      upsertUbam
+    }
+  }
+
+  private def queryForWgsUbam(webClient: ClioWebClient,
+                              config: BaseArgs): Future[Json] = {
```
danxmoran: I'd make the return type of this be `Future[TransferWgsUbamV1QueryOutput]`. You'll need to roll `getFirstObject` into this to make it happen.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -6,9 +6,7 @@ import java.util.Comparator
 
 import scala.io.Source
 import scala.sys.process.Process
-
-object IoUtil {
-
+trait IoUtilTrait {
```
danxmoran: I'm not a huge fan of this name... I think it'd be better to leave it as `IoUtil`. You can still call the "real" object `IoUtil` if you want; the compiler will figure out which one you mean based on the usage context. Or you could rename the object to something more specific.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -88,4 +90,21 @@ class ClioWebClient(clioHost: String, clioPort: Int, useHttps: Boolean)(
...
   }
+
+  def getResponseAsJson(httpResponse: Future[HttpResponse]): Future[Json] = {
+    httpResponse.flatMap(Unmarshal(_).to[Json])
+  }
+
+  def ensureOkResponse(
+    httpResponse: Future[HttpResponse]
+  ): Future[HttpResponse] = {
+    httpResponse.flatMap(
```
danxmoran: You could use `map` instead of `flatMap` here, since you're always mapping success to `Future.success` and failure to `Future.failed`. If you `throw` inside a `map` it'll result in a failed `Future`.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -88,4 +90,21 @@ class ClioWebClient(clioHost: String, clioPort: Int, useHttps: Boolean)(
...
+      r =>
+        r.status match {
+          case StatusCodes.OK => Future.successful(r)
```
danxmoran: This might be too restrictive; better to have an if-else checking if `r.status.isSuccess()`.

clio-client/src/test/scala/org/broadinstitute/client/BaseClientSpec.scala
```diff
@@ -22,5 +20,5 @@ abstract class BaseClientSpec
    * See the "Asynchronous execution model" section of
    * http://www.scalatest.org/user_guide/async_testing for more info.
    */
```
danxmoran: This comment block should be deleted now that we're not overriding the EC.

clio-client/src/test/scala/org/broadinstitute/client/BaseClientSpec.scala
```diff
@@ -22,5 +20,5 @@ abstract class BaseClientSpec
...
-  override implicit val executionContext: ExecutionContext = system.dispatcher
+  implicit val ioUtil: IoUtilTrait = IoUtil
```
danxmoran: Shouldn't we be using the mock IoUtil everywhere in tests?

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -88,4 +90,21 @@ class ClioWebClient(clioHost: String, clioPort: Int, useHttps: Boolean)(
       ).addHeader(Authorization(credentials = OAuth2BearerToken(bearerToken)))
     )
   }
+
+  def getResponseAsJson(httpResponse: Future[HttpResponse]): Future[Json] = {
```
danxmoran: Could you make this more general, like
```scala
def unmarshal[T : FromEntityUnmarshaller](httpResponse: Future[HttpResponse]): Future[T]
```

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
tlangs: The error is the same if clio returns 0 ubams or <1 ubams.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
tlangs: good to know!

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
tlangs: I had originally done that, but IntelliJ yelled at me. Did it again, and realized it was a dismissable warning.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
danxmoran: I think it'd be more useful to distinguish the two, and in the case of > 1 to print all the returned metadata.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
tlangs: Its updated in my code and its pushed. IDK why its not showing up here.

danxmoran: You moved the one from the `else` block, but didn't delete this one: https://github.com/broadinstitute/clio/pull/85/commits/24e1bdbf31bdb9d73b9e453e7656d1a753e95f63

danxmoran: Er, more helpful link: https://github.com/broadinstitute/clio/commit/24e1bdbf31bdb9d73b9e453e7656d1a753e95f63#diff-24b82004c0428d3dae5917e987ab972fL56

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/AddWgsUbamCommand.scala
```diff
@@ -15,10 +15,11 @@ import io.circe.parser._
...
 
 object AddWgsUbamCommand extends Command {
-  def execute(webClient: ClioWebClient, config: BaseArgs)(
-    implicit ec: ExecutionContext
-  ): Future[HttpResponse] = {
-    val decodedOrError = parse(IoUtil.readMetadata(config.metadataLocation.get))
+  def execute(
+    webClient: ClioWebClient,
+    config: BaseArgs
+  )(implicit ec: ExecutionContext, ioUtil: IoUtil): Future[HttpResponse] = {
```
danxmoran: I still don't think `ioUtil` should be an implicit, here and elsewhere. The few extra characters it saves us from typing isn't worth the magic in this case.

In general I think implicit args should be avoided unless:
1. An API we're using expects the arg to be implicit (like the Akka stuff that expects an implicit `ActorSystem`, and all the methods on `Future` that expect an implicit `ExecutionContext`), or
2. The API is noticeably cleaner with the parameter as implicit (again looking at the methods on `Future`, they all have signatures like `def xyz(f: X => Y)(implicit ec: EC)`. The fact that the EC is implicit lets us write `fut.xyz { x => x.y }` instead of `fut.xyz(x => x.y, ec)` or `fut.xyz(x => x.y)(ec)`)

`ioUtil` doesn't qualify.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -25,11 +25,17 @@ object ClioClient extends App {
 
 class ClioClient(val webClient: ClioWebClient) {
```
danxmoran: I'd have this take an `IoUtil` as argument, for ultimate mock-ability

clio-client/src/main/scala/org/broadinstitute/clio/client/parser/BaseArgs.scala
```diff
@@ -104,6 +105,36 @@ class BaseParser extends scopt.OptionParser[BaseArgs]("clio-client") {
         .text("The project for this whole genome unmapped bam.")
     )
 
+  cmd(Commands.MoveWgsUbam.toString)
+    .action((_, c) => c.copy(command = Some(Commands.MoveWgsUbam)))
+    .text(
+      "This command is used to copy a unmapped bam from one location to another. Only implemented for cloud unmapped bams"
```
danxmoran, line 111: I can't find any code that asserts the location is only `Location.GCP`. We should be checking it, and raising a nice error if something else is given.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -0,0 +1,123 @@
...
+    config: BaseArgs
+  ): Future[TransferWgsUbamV1QueryOutput] = {
+    val httpResponse = webClient.queryWgsUbam(
+      config.bearerToken.getOrElse(""),
+      TransferWgsUbamV1QueryInput(
+        flowcellBarcode = config.flowcell,
+        lane = config.lane,
+        libraryName = config.libraryName,
+        location = Option(Location.GCP),
+        lcSet = None,
```
pshapiro4broad: Is it possible to set omitted parameters to `None` instead of listing them explicitly?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/FutureWithErrorMessage.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.typesafe.scalalogging.Logger
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+object FutureWithErrorMessage {
+  implicit class FutureWithErrorMessage[A](future: Future[A]) {
+    def withErrorMsg(message: String)(implicit ec: ExecutionContext,
+                                      logger: Logger): Future[A] = {
+      future andThen {
```
danxmoran: `andThen` is only for side-effects, it doesn't actually change the contents of the Future. So if you do:
```scala
Future.successful(123).andThen { case Success(n) => n + 2 }
```
You still get `123`.

Since you're not doing anything extra in the `Success` case here, you can change this to:
`future recover { ex => logger.err(message); throw new Exception(message, ex) }`

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -88,4 +92,23 @@ class ClioWebClient(clioHost: String, clioPort: Int, useHttps: Boolean)(
...
+
+  def ensureOkResponse(
+    httpResponse: Future[HttpResponse]
+  ): Future[HttpResponse] = {
+    httpResponse.map(
+      r =>
+        if (r.status.isSuccess()) {
+          r
+        } else {
+          throw new Exception("Error while upserting the WgsUbam")
```
danxmoran: Can you include the status code in the message?

danxmoran (changes requested):

jacarey:
Gonna get some conflicts with https://github.com/broadinstitute/clio/pull/84

I'm happy doing the resolution but we should try to get 84 merged soon in order to avoid this again.

clio-client/src/main/scala/org/broadinstitute/clio/client/parser/BaseArgs.scala
tlangs: Fixed.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
```diff
@@ -2,57 +2,57 @@ package org.broadinstitute.clio.client.commands
...
-              s" Response code: ${response.status}"
-          )
-          logger.info(response.toString)
-        }
-        isSuccess
-      },
-      ClioClientConfig.responseTimeout
-    )
+  )(implicit ec: ExecutionContext): Future[Boolean] = {
+    responseFuture.map(response => {
```
danxmoran: Nitpick: for multi-line blocks the convention is usually to pass the function in block form like:
```
responseFuture.map { response =>
  ???
}
```

clio-client/src/test/scala/org/broadinstitute/client/commands/MoveWgsUbamSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+      command = Some(MoveWgsUbam),
+      flowcell = testFlowcell,
+      lane = testLane,
+      libraryName = testLibName,
+      location = testLocation,
+      bearerToken = testBearer,
+      ubamPath = testUbamCloudDestinationPath
+    )
+    succeedingDispatcher.dispatch(config).map(_ should be(true))
+  }
```
danxmoran, line 151: Add a test for trying to move a non-GCS ubam?

danxmoran (approved):
Two last things, but overall looks good to merge. Thanks for going through all the refactoring gymnastics!

clio-client/src/test/scala/org/broadinstitute/client/commands/MoveWgsUbamSpec.scala
```diff
@@ -105,6 +105,36 @@ class MoveWgsUbamSpec extends BaseClientSpec {
...
+
+  it should "throw and exception if the destination path is not in GCP" in {
+    a[Exception] should be thrownBy {
+      val config = BaseArgs(
+        command = Some(MoveWgsUbam),
+        flowcell = testFlowcell,
+        lane = testLane,
+        libraryName = testLibName,
+        location = Some("OnPrem"),
```
danxmoran: This location should be `Some("GCP")` to really test it's the `ubamPath` raising the error.

tlangs: Damn copy pasta.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
danxmoran: They should all have default values of `None` now, so omitting them will do the right thing.

pshapiro4broad (approved):
Be sure to give @jacarey a heads up as this will require updating with the new command line args stuff.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
danxmoran: https://github.com/broadinstitute/clio/blob/develop/clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferWgsUbamV1QueryInput.scala

clio-client/src/test/scala/org/broadinstitute/client/BaseClientSpec.scala
```diff
@@ -1,26 +1,24 @@
...
-    with Matchers {
+    with Matchers
+    with BeforeAndAfterEach {
+
+  val succeedingDispatcher =
+    new CommandDispatch(MockClioWebClient.returningOk, MockIoUtil)
+  val failingDispatcher =
+    new CommandDispatch(MockClioWebClient.returningInternalError, MockIoUtil)
+
+  override protected def beforeEach(): Unit = MockIoUtil.resetMockState()
```
danxmoran: Your tests are running inside `Future`s; this might cause the state to be reset in the middle of execution.

PR #83, opened 2017-08-24
DSDEGP-1499 Set up harness for loading service account credentials in Clio

danxmoran:
### Description

Moved the account-loading logic out of the integration-test runner and into its own utility class, and added some infrastructure for loading an environment from config.

----

### Checklist (never delete this)

Never delete this, it is our record that procedure was followed. If you find that for whatever reason one of the checklist points doesn't apply to your PR, you can leave it unchecked but please add an explanation below.

- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: Ensure that you have added or updated tests
- [ ] **Submitter**: If you're adding/switching libraries or otherwise changing the design, update the [design documentation](https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509).
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * Add notes on what you've tested
* Review cycle:
  * Team may comment on PR at will
  * **Reviewer assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to reviewer** for further feedback
- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Squash commits and merge to develop
- [ ] **Submitter**: Delete branch after merge
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [ ] **Submitter**: Move the JIRA issue to QA once this checklist is completed

PR #86, opened 2017-08-24
DSDEGP-1499 Don't use Vault within main app code

danxmoran:
### Description

This partially reverts #83, which worked locally but failed on deploy because we don't keep Vault tokens on our Clio hosts. I talked with @hjfbynara about it and he recommended that we render the service-account JSON to file on deploy and read it from disk within Clio (rather than use the Vault client within our main application code). @jacarey also suggested that we shouldn't plan to persist to gotc-dev storage when running Clio locally, so I set up the config parser to treat the service account JSON path as an optional parameter.

----

* Review cycle:


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -16,6 +22,18 @@ abstract class EnvIntegrationSpec(env: String)
     extends BaseIntegrationSpec(s"Clio in $env")
     with IntegrationSuite {
 
+  /** URL of vault server to use when getting bearer tokens for service accounts. */
+  private val vaultUrl = "https://clotho.broadinstitute.org:8200/"
```
danxmoran, line 26: The integration test code still uses the vault client because it felt simpler than trying to auto-generate the account JSON at test-runtime using SBT, and everywhere we run the ITs has access to vault / a vault token.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
```diff
@@ -1,48 +1,63 @@
...
-      override def read(config: Config, path: String): Env = {
-        val maybeEnv = config.as[String](path)
-        Env.withName(maybeEnv)
+    private val persistence = server.as[Config]("persistence")
+    private val jsonPath = persistence.getAs[Path]("service-account-json")
+    lazy val serviceAccount: Option[ServiceAccount] =
+      jsonPath.map { path =>
+        import io.circe.parser._
+        val jsonBlob =
+          Files.readAllLines(path).asScala.mkString("\n").stripMargin
```
jacarey: Why not Source.fromFile(path.toFile).mkString ?

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
```diff
@@ -1,48 +1,63 @@
...
 /**
   * Configuration file accessors.
   */
 object ClioServerConfig {
-  private val config = ClioConfig.getConfig("server")
+  private val server = ClioConfig.getConfig("server")
+
+  object Persistence extends ModelAutoDerivation {
+    implicit val pathReader: ValueReader[Path] =
+      (config: Config, path: String) => new File(config.getString(path)).toPath
```
jacarey: Paths.get(config.getString(path))

clio-server/src/main/resources/reference.conf
```diff
@@ -8,14 +8,16 @@
...
-  environment: dev
-
   http-server {
     interface: 0.0.0.0
     port: 8080
     shutdown-timeout: 60s
   }
 
+  persistence {
+    service-account-json: null
```
jacarey, line 18: does ficus turn this in to a null Path or a Path to "null"?

jacarey (approved):
Couple of small comments but :+1:

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
pshapiro4broad: Not sure how it's done in scala, but in Java I'd do `new String(Files.readAllBytes(p))` (which is what you did above, maybe it didn't work here for some reason?)

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -37,8 +55,41 @@ abstract class EnvIntegrationSpec(env: String)
...
+
+    val serviceAccount: ServiceAccount =
+      accountJSON
+        .as[ServiceAccount]
+        .fold({ err =>
+          throw new RuntimeException(
+            s"Failed to decode service account JSON from Vault at $vaultPath",
+            err
+          )
+        }, json => json)
```
pshapiro4broad: can you replace `X => X` with `_`?

clio-server/src/main/resources/reference.conf
```diff
@@ -8,14 +8,16 @@
...
 clio.server {
-  environment: dev
-
   http-server {
     interface: 0.0.0.0
     port: 8080
     shutdown-timeout: 60s
   }
 
+  persistence {
```
pshapiro4broad, line 17: Why not `clio.persistence`?

Could this attribute be placed directly under `clio` instead, e.g. `clio.service-account-json`? Or does it need to be defined as a top-level item?

clio-server/src/main/resources/reference.conf
danxmoran: The `getAs[Path]` method turns it into a `None` if null, or `Some[Path]` if given.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
```diff
@@ -1,48 +1,63 @@
...
+
 /**
   * Configuration file accessors.
   */
 object ClioServerConfig {
-  private val config = ClioConfig.getConfig("server")
+  private val server = ClioConfig.getConfig("server")
+
+  object Persistence extends ModelAutoDerivation {
+    implicit val pathReader: ValueReader[Path] =
```
pshapiro4broad: Can this be defined in `ClioConfig` so it can be used by both server and client?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
danxmoran: _ doesn't work, but I found a pre-defined `identity` that I'll try out.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
```diff
@@ -1,48 +1,63 @@
...
 
+import java.io.File
+import java.nio.file.{Files, Path}
+
 /**
   * Configuration file accessors.
   */
 object ClioServerConfig {
-  private val config = ClioConfig.getConfig("server")
+  private val server = ClioConfig.getConfig("server")
```
pshapiro4broad: Why rename this? The object seems more like `config` than `server` to me. Would `serverConfig` be better?

clio-util/src/main/scala/org/broadinstitute/clio/util/model/ServiceAccount.scala
```diff
@@ -0,0 +1,38 @@
...
+import scala.collection.JavaConverters._
+
+import java.net.URI
+
+/**
+  * Representation of service-account JSON produced by GCloud,
+  * used for persisting Clio updates to cloud storage / for
+  * communicating through Clio's OpenIDC proxy.
+  */
+case class ServiceAccount(authProviderX509CertUrl: URI,
```
pshapiro4broad: Not all declared arguments are used. It it required that we declare them to avoid errors when converting from JSON?

pshapiro4broad (approved):
A few questions but looks OK to me.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
danxmoran: Will use `serverConfig`. I renamed it because `ModelAutoDerivation` has a member named `config`, so I was getting name clashes.

clio-util/src/main/scala/org/broadinstitute/clio/util/model/ServiceAccount.scala
danxmoran: Yeah, I was getting errors about leftover fields from Circe. I'll add a note in the scaladoc.

clio-server/src/main/resources/reference.conf
danxmoran: I think `clio.server.persistence` makes more sense than `clio.persistence` since all the GCS-writing is going to happen within the server. I made the `persistence` wrapper around the account-JSON setting in case we end up needing to add more configuration around how we write to / read from the buckets.

pshapiro4broad: OK. I know we will have different bucket locations based on environment, so perhaps that will go here.

Although in this case we aren't using the service account for persistence, but for authentication, correct?

danxmoran: Authentication-for-persistence, so yes and no :)

An alternate setup I looked at was how Cromwell does things, where they have a separate [config block](https://github.com/broadinstitute/cromwell#configuring-authentication) for accounts in general, and then other spots in config can refer to accounts by-name. I decided against it because it seemed like overkill for what we're trying to do.

PR #88, opened 2017-08-26
Tl dsdegp 1376 move ubams round 2

tlangs:
### Description

Major usability improvements. The client no longer hangs when commands throw exceptions. There are also much better error messages for the move command.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
```diff
@@ -27,32 +28,33 @@ class CommandDispatch(val webClient: ClioWebClient, val ioUtil: IoUtil)
...
-      isSuccess
+  )(implicit ec: ExecutionContext): Future[Unit] = {
+    responseFuture.transformWith {
+      case Success(response) =>
+        if (response.status.isSuccess()) {
+          logger.info(
+            s"Successfully completed command." +
+              s" Response code: ${response.status}"
+          )
+          Future.successful(())
```
pshapiro4broad: It seems like this discards `response`. What if the caller needs to know if the response was a `200` or a `201`? Or what the body contains? E.g.  why not return `responseFuture` instead of `Future.successful()`?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
+        "Could not query the WgsUbam. No files have been moved."
+      _ <- copyGoogleObject(wgsUbamPath, config.ubamPath, ioUtil) withErrorMsg
+        "An error occurred while copying the files in google cloud. No files have been moved."
       upsertUbam <- upsertUpdatedWgsUbam(webClient, config) withErrorMsg
         """An error occurred while upserting the WgsUbam. The state of clio is now inconsistent.
           |The ubam exists in both at both the old and the new locations.
           |At this time, Clio only knows about the bam at the old location.
           |Then, try upserting (not moving) the unmapped bam with the correct path then deleting the old bam.
-          |If the state of Clio cannot be fixed, please contact the Green Team at greenteam@broadinstitute.org
+          |If the state of Clio cannot be fixed, please contact the Green Team at greenteam@broadinstitute.org.
```
pshapiro4broad: The email address for green team should be a global constant, at least.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
-      _ <- copyGoogleObject(wgsUbam.ubamPath, config.ubamPath, ioUtil) withErrorMsg
-        "An error occurred while moving the files in google cloud. No files have been moved"
+      _ <- verifyPath(webClient, config) withErrorMsg
+        "Clio client can only handle cloud operations right now."
+      wgsUbamPath <- queryForWgsUbamPath(webClient, config) withErrorMsg
+        "Could not query the WgsUbam. No files have been moved."
+      _ <- copyGoogleObject(wgsUbamPath, config.ubamPath, ioUtil) withErrorMsg
+        "An error occurred while copying the files in google cloud. No files have been moved."
       upsertUbam <- upsertUpdatedWgsUbam(webClient, config) withErrorMsg
         """An error occurred while upserting the WgsUbam. The state of clio is now inconsistent.
```
pshapiro4broad: This doesn't read quite right to me. Clio's state (which I read is its internal state) is still totally consistent. What's inconsistent is the state of the world outside of clio.

If we wanted to try and fix that here, we could add code that deletes the copied object in GCS. That would leave things in the same state that they started out in, before the command was executed.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
           |The ubam exists in both at both the old and the new locations.
           |At this time, Clio only knows about the bam at the old location.
           |Then, try upserting (not moving) the unmapped bam with the correct path then deleting the old bam.
```
pshapiro4broad: Is this a valid recovery scenario? E.g. is there a CLP that can update the metadata in this way?

If this kind of failure can occur, we should describe somewhere (confluence, etc) what the recovery process is. If we can delete the copied resource, then the recovery scenario is to run the same command again, once the underlying problem has been fixed.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
+    }
+  }
+
+  private def queryForWgsUbamPath(webClient: ClioWebClient,
+                                  config: BaseArgs): Future[Option[String]] = {
     implicit val ec: ExecutionContext = webClient.executionContext
+
+    def ensureOnlyOneJson(wgsUbams: Json): Json = {
+      wgsUbams.asArray.getOrElse(throw new Exception()).size match {
+        case 1 =>
```
pshapiro4broad: It looks like all uses of `1` here refer to the same concept. Replacing them all with a named constant would make that clear.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
+          wgsUbams.asArray.get.head
+        case s if s > 1 =>
+          logger.error(
+            s"$s WgsUbams were returned for Key(${prettyKey(config)}), expected 1."
+          )
+          throw new Exception()
+        case s if s == 0 =>
```
pshapiro4broad: Using `case _ =>` would make the intent of this code clearer.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
+        logger.error("Only GCP unmapped bams are supported at this time.")
+        throw new Exception()
+    }).flatMap { _ =>
+      Future(config.ubamPath.foreach {
+        case loc if loc.startsWith("gs://") => ()
+        case _ =>
+          logger.error(
+            s"The destination of the ubam must be a cloud path. ${config.ubamPath.get} is not a cloud path."
+          )
+          throw new Exception()
```
pshapiro4broad: In Java I would construct the exception using the error message and put the logging in the top-level catch block. Does that approach not apply to Scala?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
+          logger.error(s"No WgsUbams were found for Key(${prettyKey(config)})")
+          logger.error("You can add this WgsUbam using the AddWgsUbam command in the Clio client.")
+          throw new Exception()
+      }
+    }
+
+    // Using JSON instead of TransferWgsUbamV1QueryOutput here on purpose.
+    // Currently, the clio server spits json out in snake_case.
+    // Clio client expects it json keys to be in camelCase.
+    // This should be changed to use the transfer model once the issue is fixed
```
pshapiro4broad: Is there a JIRA for this issue? This seems like something we should iron out early on.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -105,23 +137,45 @@ object MoveWgsUbamCommand
...
+          libraryName = config.libraryName.get,
+          location = Location.GCP
+        ),
+        transferWgsUbamV1Metadata =
+          TransferWgsUbamV1Metadata(ubamPath = config.ubamPath)
+      )
+      .map(ensureOkResponse)
+  }
+
+  private def prettyKey(config: BaseArgs): String = {
```
pshapiro4broad, line 172: It seems like this should be a method on `ModelWgsUbamKey`, instead of a utility method here.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -96,7 +126,9 @@ object MoveWgsUbamCommand
     ioUtil.copyGoogleObject(source.get, destination.get) match {
       case 0 => Future.successful(())
       case _ =>
-        Future.failed(new Exception("Copy files in google cloud failed"))
+        logger.error(s"Copy files in google cloud failed from '${source
```
pshapiro4broad: The different error message texts refer to both "google cloud" and "cloud". Maybe just "cloud" would be better everywhere to be consistent? We are probably not working with other cloud storage systems for a while so saying "cloud" is unambiguous.

pshapiro4broad (approved):
A few comments, as far as I can tell the code is 👍 overall

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
danxmoran: +1 to returning the entire `responseFuture`. It'll make using the whole `ClioClient` in the integration tests possible.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
danxmoran: It should apply. I think if the message got put in the thrown Exception, `withErrorMessage` would handle logging it out.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +28,96 @@ object MoveWgsUbamCommand
...
+                                  config: BaseArgs): Future[Option[String]] = {
     implicit val ec: ExecutionContext = webClient.executionContext
+
+    def ensureOnlyOneJson(wgsUbams: Json): Json = {
+      wgsUbams.asArray.getOrElse(throw new Exception()).size match {
+        case 1 =>
+          wgsUbams.asArray.get.head
+        case s if s > 1 =>
+          logger.error(
+            s"$s WgsUbams were returned for Key(${prettyKey(config)}), expected 1."
```
danxmoran: Could this message include the returned JSON, so users can see & debug the duplicates?

danxmoran: Alt: it could log out something about using the `QueryWgsUbam` command to see the duplicates.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
danxmoran: I'd rather we fix it now. The snake-case config is defined in `ModelAutoDerivation`, so it should be applying. @tlangs are you sure that was the cause of whatever error led you to do this?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/FutureWithErrorMessage.scala
```diff
@@ -3,15 +3,18 @@ package org.broadinstitute.clio.client.util
...
   implicit class FutureWithErrorMessage[A](future: Future[A]) {
     def withErrorMsg(message: String)(implicit ec: ExecutionContext,
                                       logger: Logger): Future[A] = {
-      future recover {
-        case ex =>
+      future andThen {
+        case Success(_) => ()
+        case Failure(ex) =>
+          if (Option(ex.getMessage).isDefined)
+            logger.error(ex.getMessage)
```
danxmoran, line 16: This should handle logging out any error messages thrown during processing.

danxmoran (changes requested):
Would much prefer
```scala
throw new Exception("message")
```
Over the current
```scala
logger.error("message")
throw new Exception()
```

The snake-case vs camel case thing also weirds me out. The integration tests are able to unmarshal JSON from the server without any special config apart from `ModelAutoDerivation`, so I don't know what else the client should need.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
tlangs: I assumed that was the problem. It was having errors with fields like 'flowcell_barcode', but when I tried some custom data with 'flowcellBarcode', it worked fine...

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/CommandDispatch.scala
tlangs: That's an oopsie on my part.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
tlangs: I was using "clio's state" as the state of not only the application, but also the external state of ubam locations. I agree its a bit ambiguous. I'll change it.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
tlangs: I'm not sure storing 1 in a named variable buys us much here. Its just the size of the list

clio-client/src/main/scala/org/broadinstitute/clio/client/util/FutureWithErrorMessage.scala
tlangs: I've modified the thrown exceptions to contain messages.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
pshapiro4broad: There are three `1`s here that refer to the same concept; the list size is not the the concept; the limit is. Changing the code to support 1 and 0 cases could also get rid of this issue.
```
match {
case 1 => wgsUbams.asArray.get.head
case 0 => throw new Exception(...)
case _ => throw new Exception(...)
}
```

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
tlangs: I'm not sure that buys much. Here, I'm printing out the config arguments, not a ModelWgsUbamKey. The config might have empty fields, which ModelWgsUbamKey can't handle, since its parameters aren't defined as Option

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
tlangs: Good point. Changed

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
pshapiro4broad: OK. Yeah, you'd have to first make a key here and then ask it to print, and I guess that would require some work.

clio-client/src/main/resources/application.conf
```diff
@@ -1 +1,3 @@
-# Define all config overrides here.
\ No newline at end of file
+# Define all config overrides here.
+
+greenteam.email = "greenteam@broadinstitute.org"
```
danxmoran: This should go in the reference.conf, not application.conf.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +33,97 @@ object MoveWgsUbamCommand
...
-        lane = config.lane,
-        libraryName = config.libraryName,
-        location = Option(Location.GCP),
-        documentStatus = Option(DocumentStatus.Normal)
-      )
-    )
+  private def verifyPath(webClient: ClioWebClient,
+                         config: BaseArgs): Future[Unit] = {
+    implicit val executionContext: ExecutionContext = webClient.executionContext
+    Future(config.location.foreach {
```
danxmoran: More idiomatic style to do:
```scala
Future {
  config.location.foreach {
    ... stuff ...
  }
}
```

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +33,97 @@ object MoveWgsUbamCommand
...
+      case "GCP" => ()
+      case _ =>
+        throw new Exception(
+          "Only GCP unmapped bams are supported at this time."
+        )
+    }).flatMap { _ =>
+      Future(config.ubamPath.foreach {
```
danxmoran: This doesn't need to be a `flatMap`, the `Future` body isn't actually doing anything async. You could either change the `flatMap` to a `map` or just move the `foreach` up into the Future above, under the other `foreach`.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
```diff
@@ -31,63 +33,97 @@ object MoveWgsUbamCommand
...
+      })
+    }
+  }
+
+  private def queryForWgsUbamPath(webClient: ClioWebClient,
+                                  config: BaseArgs): Future[Option[String]] = {
     implicit val ec: ExecutionContext = webClient.executionContext
+
+    def ensureOnlyOneJson(wgsUbams: Json): Json = {
+      wgsUbams.asArray.getOrElse(throw new Exception()).size match {
```
danxmoran: No error message here?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/MoveWgsUbamCommand.scala
danxmoran: Let's huddle about dis.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -83,7 +83,10 @@ object Dependencies {
     "io.circe" %% "circe-generic" % CirceVersion,
     "io.circe" %% "circe-generic-extras" % CirceVersion,
     "io.circe" %% "circe-parser" % CirceVersion,
-    "org.slf4j" % "slf4j-api" % Slf4jVersion
+    "org.slf4j" % "slf4j-api" % Slf4jVersion,
+    "org.typelevel" %% "cats-core" % "1.0.0-MF",
+    "org.typelevel" %% "cats-free" % "1.0.0-MF",
+    "org.typelevel" %% "cats-kernel" % "1.0.0-MF"
```
danxmoran: Was this on purpose?

tlangs: nope

tlangs:
retest this please

danxmoran (approved):

PR #93, opened 2017-08-31
Delete WgsUbam clp

tlangs:
### Description

A CLP to delete WgsUbams in the cloud and in Clio. 

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+import scala.collection.immutable.Seq
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+object DeleteWgsUbamCommand
+    extends Command
+    with LazyLogging
+    with FailFastCirceSupport
+    with ModelAutoDerivation
+    with FutureWithErrorMessage {
```
danxmoran: If these `with`s are common across all our commands, it'd be nice to add them to the base `Command` type.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  implicit val implicitLogger: Logger = logger
```
danxmoran: Wat. Could you add a comment for where this logger is used?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  override def execute(
+    webClient: ClioWebClient,
+    config: BaseArgs,
+    ioUtil: IoUtil
+  )(implicit ec: ExecutionContext): Future[HttpResponse] = {
+    for {
+      _ <- webClient.verifyCloudPaths(config) withErrorMsg ()
```
danxmoran: What's the point of `withErrorMsg` with no message?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+          )
+        )
+        .map(webClient.ensureOkResponse) withErrorMsg "There was a problem querying the Clio server for WgsUbams."
+      wgsUbams <- webClient.unmarshal[Seq[TransferWgsUbamV1QueryOutput]](
+        queryResponses
+      )
+      deleteResponses <- deleteWgsUbams(wgsUbams, webClient, config, ioUtil)
+    } yield {
+      deleteResponses.size match {
+        case s if s != deleteResponses.size =>
```
danxmoran: I'm pretty sure this `if` will never be true, since you're matching on `deleteResponses.size`. I'm also not sure if distinguishing the 0 case is worth much.

Could we change the match to be like:
```scala
deleteResponses.size match {
  case s if s == wgsUbams.size => OK
  case _ => Not OK
}
```

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+          throw new Exception(
+            s"Deleted ${deleteResponses.size} WgsUbams. " +
+              s"Not all of the WgsUbams queried for were able to be deleted! Check the log for details"
```
danxmoran: I'm not sure we should rely on logging to communicate the failed deletes, since users might toggle logging off entirely. Could we zip up the delete responses with their original ubam objects, so when some fail we can include exactly which deletes failed inside this message?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+          logger.info(s"Successfully deleted ${deleteResponses.size} WgsUbams.")
+          deleteResponses.head
+      }
+    }
+  }
+
+  def deleteWgsUbams(wgsUbams: Seq[TransferWgsUbamV1QueryOutput],
+                     webClient: ClioWebClient,
+                     config: BaseArgs,
+                     ioUtil: IoUtil): Future[Seq[HttpResponse]] = {
```
danxmoran: I think a more useful output type from the caller's perspective would be `Future[Seq[(TransferWgsUbamV1QueryOutput, HttpResponse)]]`.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+    implicit val ec: ExecutionContext = webClient.executionContext
+    if (wgsUbams.isEmpty) {
+      return Future.failed(
+        new Exception(
+          s"No WgsUbams were found for $config. Nothing has been deleted."
+        )
+      )
+    }
+    Future.foldLeft(
+      wgsUbams.map(deleteWgsUbam(_, webClient, config, ioUtil).transformWith({
```
danxmoran: Could you pull this call out into its own val, just to de-clutter the `foldLeft` call?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+        case Success(httpResponse) => Future.successful(Right(httpResponse))
+        case Failure(ex)           => Future.successful(Left(ex))
```
danxmoran: Could you add a comment for why you're transforming the futures into futures-of-either? (Is it because `foldLeft` ignores failed futures by default?)

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
```diff
@@ -0,0 +1,145 @@
...
+      ioUtil.deleteGoogleObject(wgsUbam.ubamPath.get) match {
+        case 0 => ()
+        case 1 =>
+          return Future
+            .failed(
+              new Exception(
+                s"Failed to delete ${wgsUbam.ubamPath.get} in the cloud. The WgsUbam still exists in Clio and on cloud storage"
+              )
+            )
+            .withErrorMsg()
```
danxmoran: No error message?

clio-client/src/test/scala/org/broadinstitute/client/commands/DeleteWgsUbamSpec.scala
```diff
@@ -0,0 +1,137 @@
...
+      command = Some(DeleteWgsUbam),
+      flowcell = testFlowcell,
+      lane = testLane,
+      libraryName = testLibName,
+      location = testLocation,
+      bearerToken = testBearer,
+    )
+    succeedingReturningDispatcher(new MockIoUtil)
+      .dispatch(config)
+      .map(_.status should be(StatusCodes.OK))
```
danxmoran, line 103: Do we want to return success if the user tries to delete a nonexistent file?

danxmoran (changes requested):
Mostly questions, one request about the error messages.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
tlangs: `withErrorMessage` prints the Exception message to the logger as well as any given message, in the case where nothing more needs to be said, we can just let `withErrorMessage` print the Exception message

clio-client/src/main/scala/org/broadinstitute/clio/client/util/FutureWithErrorMessage.scala
```diff
@@ -7,14 +7,15 @@ import scala.util.{Failure, Success}
 
 trait FutureWithErrorMessage {
   implicit class FutureWithErrorMessage[A](future: Future[A]) {
-    def withErrorMsg(message: String)(implicit ec: ExecutionContext,
-                                      logger: Logger): Future[A] = {
+    def withErrorMsg(message: String = "")(implicit ec: ExecutionContext,
```
jacarey: withErrorMsg should always have a readable message otherwise we should rename it.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
tlangs: Good idea.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
tlangs: refactored things to make it less wat-ful

clio-client/src/test/scala/org/broadinstitute/client/commands/DeleteWgsUbamSpec.scala
tlangs: The basis for this test is that the file on the cloud has already been deleted, but Clio doesn't know about it. I was unsure what to return in this case. I chose to return OK because the state of Clio will reflect reality after the command is run.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
tlangs: see above

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -99,4 +100,33 @@ class ClioWebClient(clioHost: String, clioPort: Int, useHttps: Boolean)(
...
+  def verifyCloudPaths(config: BaseArgs): Future[Unit] = {
+    Future {
+      config.location.foreach {
+        case "GCP" => ()
+        case _ =>
+          throw new Exception(
+            "Only GCP unmapped bams are supported at this time."
+          )
+      }
+      config.ubamPath.foreach {
```
jacarey: When we run input validations its nice to collect all errors instead of fast failing.. that way you can fix them all instead of one and trying again.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
tlangs: Why would this be more useful?

jacarey (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/DeleteWgsUbamCommand.scala
danxmoran: For implementing [this comment](https://github.com/broadinstitute/clio/pull/93#discussion_r136355053).

tlangs: Ah ok

tlangs: The problem I'm seeing here is that the `Future[Seq[(TransferWgsUbamV1QueryOutput, HttpResponse)]]` will only have successful communications with the server (like, it was able to contact the server, but got a non-OK response). Failures to even contact the server would not be included.

danxmoran (approved):
OK to merge if we assume error logging will at least always be on, would still prefer a different name to `withErrorMsg` if it's going to support not passing in a message.

PR #95, opened 2017-08-31
DSDEGP-1500 DSDEGP-1501 Clio writes metadata updates to storage.

danxmoran:
### Description

Follow-up to #97, not for merging until after it goes in. This actually implements the functionality for 1500 and 1501. I've manually verified writing to GCS works, ~~and I'll set up one last PR on top of this one adding integration tests~~ I went ahead and added the integration-test changes to this PR.

----

* Review cycle:


danxmoran:
retest this please

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -86,6 +101,28 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
+        index.rootDir,
+        now.getYear.toString,
+        now.getMonthValue.toString,
+        now.getDayOfMonth.toString,
+        s"$id.json"
+      )
+    )
+  }
+
+  /** Read an parse JSON out of a path. */
```
tbl3rd: and

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -1,7 +1,10 @@
 package org.broadinstitute.clio.integrationtest.tests
```
tbl3rd, line 1: Is there some convenient way to run WgsUbamTests (TestKit) from IntelliJ?
(I can inline an App object, but ... there must be a way without hacking the code,
or crashing IntelliJ.)

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -117,12 +134,30 @@ trait WgsUbamTests { self: BaseIntegrationSpec =>
     for {
       clioId1 <- clioWebClient
         .addWgsUbam(upsertKey, upsertData)
-        .flatMap(Unmarshal(_).to[String])
+        .flatMap(Unmarshal(_).to[UUID])
```
tbl3rd, line 134: "handle upserts and queries for wgs-ubam location" is slightly easier to read.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
```diff
@@ -20,20 +22,68 @@ object ClioServerConfig extends ConfigReaders {
...
-          Source.fromFile(path.toFile).mkString.stripMargin
-        decode[ServiceAccount](jsonBlob).fold({ error =>
-          throw new RuntimeException(
-            s"Could not decode service account JSON at $path",
-            error
+
+    /**
+      * Types of persistence supported by Clio.
+      */
+    sealed trait Type extends EnumEntry
```
tbl3rd: IWBNI this related to model.Location somehow.
They are the same concept, and then you don't have this loaded Type term.

tbl3rd (approved):
google-cloud-nio is nifty.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -86,6 +101,28 @@ abstract class BaseIntegrationSpec(clioDescription: String)
   /** Get a random identifier to use as a dummy value in testing. */
   def randomId: String = UUID.randomUUID().toString.replaceAll("-", "")
 
+  /** Get the expected storage path for an index / UUID pair. */
+  def expectedStoragePath(index: ElasticsearchIndex[_], id: UUID): Path = {
+    val now = OffsetDateTime.now()
```
pshapiro4broad: The knowledge of how a path in the cloud is constructed shouldn't be duplicated in the test code, as no external program is relying on it. So this code should be calling an API in the deployed code to create the path to compare to.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -86,6 +101,28 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
+    rootPersistenceDir.resolve(
+      Paths.get(
```
pshapiro4broad: Can't you construct a path in Scala using the `/` operator?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -40,6 +45,11 @@ abstract class EnvIntegrationSpec(env: String)
     "https://www.googleapis.com/auth/userinfo.email"
   )
 
+  /** Scopes needed from Google to read Clio's persistence buckets. */
+  private val storageScopes = Seq(
+    "https://www.googleapis.com/auth/devstorage.read"
```
pshapiro4broad: Is there no google library that contains constants for google scope urls?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
pshapiro4broad: I think what we want is an integration spec that launches clio-server locally from source (or connects to an already running clio-server) and spins up a local elasticsearch server. Once we have that, we should be able to debug tests and clio-server in the IDE.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -79,19 +90,36 @@ abstract class EnvIntegrationSpec(env: String)
...
 
+  override lazy implicit val bearerToken: OAuth2BearerToken = {
     val credential = serviceAccount.credentialForScopes(authScopes)
     OAuth2BearerToken(credential.refreshAccessToken().getTokenValue)
   }
+
+  override lazy val rootPersistenceDir: Path = {
+    val storageOptions = StorageOptions
+      .newBuilder()
+      .setProjectId(s"broad-gotc-$env-storage")
```
pshapiro4broad, line 111: Are there unit tests for `LocalConfig` as well?

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServerConfig.scala
```diff
@@ -20,20 +22,68 @@ object ClioServerConfig extends ConfigReaders {
...
+    object Type extends Enum[Type] {
+      override val values: IndexedSeq[Type] = findValues
+      case object Local extends Type
```
pshapiro4broad: These case classes duplicate the ones in `PersistanceConfig`; why not combine them?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/LocalFilePersistenceDAO.scala
```diff
@@ -0,0 +1,16 @@
...
+import java.nio.file.{Files, Path}
+
+/**
+  * Persistence DAO which writes to local disk, for running Clio locally
+  * without thrashing GCS in the dev environment.
+  */
+class LocalFilePersistenceDAO(config: Persistence.LocalConfig)
+    extends PersistenceDAO {
+  override lazy val rootPath: Path = config.rootDir.getOrElse {
+    Files.createTempDirectory("clio-persistence")
```
pshapiro4broad: This will leave directories around on disk after the program exits. If the only use case of `LocalConfig` is for testing, then tests could be required to always pass in the root directory, and always clean it up when the test exits. Is there a use case for `LocalConfig` other than for testing? If not, can tests be responsible for cleaning up the root directory?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
```diff
@@ -60,6 +60,10 @@ class DockerIntegrationSpec
     "dummy-token"
   )
 
+  // SBT sets a local path for persisting metadata updates.
+  override val rootPersistenceDir: Path =
+    Paths.get(sys.env("LOCAL_PERSISTENCE_DIR"))
```
pshapiro4broad: Is there a better way to share this value with tests than using an environment variable? I think it duplicates state that's in `ClioServerConfig.config`.

In any event, the magic string `LOCAL_PERSISTENCE_DIR` should be put in a constant.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -138,12 +173,26 @@ trait WgsUbamTests { self: BaseIntegrationSpec =>
...
     } yield {
-      UUIDUtil.uuid(clioId2).compareTo(UUIDUtil.uuid(clioId1)) should be(1)
+      clioId2.compareTo(clioId1) should be(1)
+
+      val expectedPath1 =
+        expectedStoragePath(ElasticsearchIndex.WgsUbam, clioId1)
+      val expectedPath2 =
+        expectedStoragePath(ElasticsearchIndex.WgsUbam, clioId2)
+      Files.exists(expectedPath1) should be(true)
+      Files.exists(expectedPath2) should be(true)
```
pshapiro4broad: Can the tests be refactored to avoid code duplication?

clio-server/config/clio.conf.ctmpl
```diff
@@ -1,18 +1,23 @@
 include "application.conf"
 
 clio.server {
-    environment: {{env "ENV"}}
+  environment: {{env "ENV"}}
```
pshapiro4broad, line 4: So the indent for these files is one space, not two?

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/MemoryPersistenceDAO.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
+  ClioDocument,
+  ElasticsearchIndex
+}
+
+import com.sksamuel.elastic4s.Indexable
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class MemoryPersistenceDAO extends MockPersistenceDAO {
```
pshapiro4broad: Do we need both `Jimfs` and `Memory` DAOs for tests? Seems like Jimfs could be used for the memory case as well.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      _ <- dao.initialize(index)
+    } yield {
+      Files.exists(wgsPath) should be(true)
+      Files.isDirectory(wgsPath) should be(true)
+    }
+  }
+
+  it should "write metadata updates to storage using the UUID" in {
+    val dao = new JimfsPersistenceDAO()
+    val document = DocumentMock(
```
pshapiro4broad: Is this really a mock? It looks like a valid document that is only used in a test, not a mock.

To make our test code simpler, the test document could also have a constructor that creates it with dummy data, to avoid the need to put that code here.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      mockFieldDouble = Some(0.9876)
+    )
+
+    for {
+      _ <- dao.initialize(index)
+      _ <- dao.writeUpdate(document, index)
+      _ <- dao.writeUpdate(document2, index)
+    } yield {
+      Seq(document, document2).foldLeft(succeed) {
+        case (_, doc) => {
```
pshapiro4broad: scala question - what is the purpose of `foldLeft() when the first argument is ignored? To me that looks like `map`.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+          val expectedPath =
+            dao.currentPathForIndex(index).resolve(s"${doc.clioId}.json")
+
+          Files.exists(expectedPath) should be(true)
+          Files.isRegularFile(expectedPath) should be(true)
+
+          parse(Files.readAllLines(expectedPath).asScala.mkString)
```
pshapiro4broad: I think using `new String(Files.readAllBytes(expectedPath))` is more direct and memory efficient.

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
```diff
@@ -17,16 +17,26 @@ object ClioIntegrationTestSettings {
     target.value / "integration-test" / "logs"
   }
 
+  /** Directory to which Clio will write metadata updates during Dockerized integration tests. */
+  lazy val persistenceTarget: Initialize[File] = Def.setting {
```
pshapiro4broad, line 22: Would be nice to use `Path` here as well, IMO

pshapiro4broad (approved):
Looks 👍 as far as I can tell.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
danxmoran: I don't think there's a `/` method on `Path` in vanilla Scala. The `/` we use elsewhere is from SBT's [RichFile](http://www.scala-sbt.org/0.13/api/index.html#sbt.RichFile) extension, so it's not available here.

clio-server/config/clio.conf.ctmpl
danxmoran: It should be 2 now, but I'll double-check.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
danxmoran: Agreed it's a weird pattern. `map` would leave us with a `Seq[Assertion]`, while `foldLeft` leaves us with just `Assertion` (which is what we need for scalatest's async specs).

But, I think I can change it to use a `foreach` and then just return `succeed` at the end. If any of the tests fail in the `foreach` it'll raise an exception and the `succeed` will never be returned.

clio-server/config/clio.conf.ctmpl
danxmoran: Confirmed, this change makes the files have an indent of 2.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
danxmoran: There is [this plugin](https://github.com/sbt/sbt-buildinfo) for injecting values into app code from SBT via a generated object.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
danxmoran: Not that I could find. [`google-cloud-java`](https://github.com/GoogleCloudPlatform/google-cloud-java/search?utf8=%E2%9C%93&q=https%3A%2F%2Fgoogleapis.com%2Fauth&type=) has them scattered around as strings too.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
danxmoran: Yes, the Docker suite uses it.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/LocalFilePersistenceDAO.scala
danxmoran: `LocalConfig` is also for manual testing / exploration, to avoid polluting dev storage when running a clio-server locally and trying things out. I made the local root directory optional to avoid breaking local setups when we don't care about where the server will write its records.

I'll add a shutdown hook to clean up the storage directory if it's being created here.

PR #97, opened 2017-09-01
DSDEGP-1501 Rip out internal models in prep for adding persistence DAO

danxmoran:
### Description

Resurrection of #94. Description from there was:

Follow-up to #91, not for merging until after it goes through. This deletes Clio's inner models and shifts around the API so that the key+model -> document conversion happens in the search service instead of the search DAO. It should be a no-op chunk of changes that sets up my next PR, which will include the logic for actually writing metadata to GCS.

*Why now?*
To implement the GCS-writing I added a new persistence service / DAO. The "update" function of the search service will move into the persistence service, and will be augmented to first write to storage, then upsert to search. Rather than have both DAOs perform an identical key+value -> document conversion, the service will perform the conversion and pass the resulting document into both DAOs. In prep for making this same flow work across all document types, I pushed the responsibility for defining the converters into the document-specific services, where the transfer -> internal model mappers currently live. With all the mappers in one place, it became pretty clear that the internal models weren't buying us anything, so I scrapped them.

This should also cut down method growth in the search DAO. Instead of defining separate updateWgsUbam, updateCram, etc. methods it just has one update entrypoint (and same for queries).

*How will versioning work now?*
The previous plan for versioning transfer models was to have every transfer model map into the One True internal model. With those internal models gone, we can still implement our versioned transfer models by:

1. Defining a document and query mapper from the latest-versioned transfer models to their corresponding elasticsearch document model.
2. Defining case-class converters from older transfer models to the newest transfer model.

----

* Review cycle:


clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
```diff
@@ -133,9 +158,11 @@ class WgsUbamWebServiceSpec
     // We have to test the MemorySearchDAO because we're not going to implement
     // Elasticsearch logic in our test specs. Here, we're just verifying that
     // the web service passes the appropriate queries onto the search DAO.
-    Post("/query", Map("flowcell_barcode" -> "FC123")) ~> webService.queryall ~> check {
```
danxmoran, line 136: While fixing up the `.head` scolding from Codacy I noticed this test was hitting the wrong route.

tbl3rd: Weird. Why didn't the test realize the URI was wrong?

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
```diff
@@ -133,10 +158,18 @@ class WgsUbamWebServiceSpec
...
+    Post("/queryall", Map("flowcell_barcode" -> "FC123")) ~> webService.queryall ~> check {
+      memorySearchDAO.queryCalls should be(
+        Seq(
+          // From /query call earlier in the test.
+          TransferWgsUbamV1QueryInput(
+            flowcellBarcode = Some("FC123"),
+            documentStatus = Some(DocumentStatus.Normal)
+          ),
+          // No documentStatus restriction from /queryall
+          TransferWgsUbamV1QueryInput(flowcellBarcode = Some("FC123"))
```
tbl3rd, line 170: When I added a status check to the prior version of the the test, the test failed with `Request was rejected`.

tbl3rd (approved):
Like the +351 -584, or at least the - part.
Thanks for breaking this review out.

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
danxmoran: The `check` body was copy-pasted from above, so it was just re-asserting things on the input model from the first `/query` run in the test.

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
danxmoran: That'd be from the route mismatch in `Post("/query") ~> webService.queryall`.

pshapiro4broad (approved):

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebServiceSpec.scala
tbl3rd: Yeah, but for the record, even just reporting the status provoked a rejection.
(Maybe check doesn't call awaitResult unless something in the body calls for it?)

PR #100, opened 2017-09-05
DSDEGP-1549 Split up ClioClient for use in integration tests.

danxmoran:
### Description

This is prep for moving our integration tests from using the clio-web-client to using the whole clio-client (so we can have automated tests for the file operations that happen as part of the move & delete CLPs that don't get covered by the web client).

I tried to comment things pretty heavily, but the main change in this PR is to split up the parsing & sys.exit-ing logic in the client, so we can call the parser piece in the tests. The caseapp `CommandAppWithPreCommand` class we're currently using calls `sys.exit` in most of its helper methods, so rather than override everything I just reimplemented the interesting parts of the class (and tried to improve some of the control flow in the process). This should also include the behavior from #96.

Some incidental changes:

- I angered the shapeless gremlins at some point and couldn't appease them with our existing summoner, but I found [something else](https://github.com/broadinstitute/clio/compare/dm-split-clio-client-DSDEGP-1549?expand=1#diff-737b5d0456ca045772bbc952e105c0a1R64) that seems to work.
- I noticed that the package structure for our tests didn't match the main code's structure, so I moved the tests to match.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,61 +1,259 @@
...
+  *                  both local and in cloud storage
+  * @param ec        the execution context to use when scheduling
+  *                  asynchronous events
+  */
+class ClioClient(webClient: ClioWebClient,
+                 ioUtil: IoUtil)(implicit ec: ExecutionContext)
+    extends LazyLogging {
+  import ClioClient.{EarlyReturn, ParsingError, UsageOrHelpAsked}
+
+  val progName: String = ClioClient.progName
```
jacarey: not sure we need to store vals here if we are just passing them to copy and never using them again.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,61 +1,259 @@
...
+    *
+    * caseapp supports setting these through annotations, but
+    * only if the values are constant strings.
+    */
+  private val beforeCommandMessages: Messages[CommonOptions] =
+    CommonOptions.messages.copy(
+      appName = appName,
+      appVersion = appVersion,
+      progName = progName,
+      optionsDesc = s"[options] [command] [command-options]"
```
jacarey: don't need string interpolation on this string.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,61 +1,259 @@
...
+    )
+
+  /** Names of all valid sub-commands. */
+  private val commands: Seq[String] =
+    ClioCommand.messages.messages.map(_._1)
```
jacarey: ClioCommand.messages.messagesMap.keys

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,61 +1,259 @@
...
+          case ParsingError(error) => {
+            System.err.println(error)
+            Future.failed(new RuntimeException)
+          }
+        },
+        identity
+      )
+      .onComplete {
+        case Success(_) => sys.exit(0)
+        case Failure(_) => sys.exit(1)
```
jacarey: We should be logging errors on failure.

jacarey (commented):
Small changes.. but overall looks great!

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
pshapiro4broad: Yes, and in general we should be logging errors at the top level, not logging at the level where the error is first generated.

jacarey (approved):

danxmoran:
@pshapiro4broad any other changes you want before I merge?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,63 +1,254 @@
...
+          println(message)
+          Future.successful(())
+        }
+        case ParsingError(error) => {
+          Future.failed(new RuntimeException(error))
+        }
+      }, identity)
+      .onComplete {
+        case Success(_) => sys.exit(0)
+        case Failure(ex) => {
```
tlangs, line 74: Tiny gripe: redundant braces are redundantly redundant

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,63 +1,254 @@
...
+        .left
+        .map(ParsingError.apply)
 
-  override def run(command: CommandType, remainingArgs: RemainingArgs): Unit = {
+    maybeParse.flatMap {
+      case (commonParse, commonArgs, maybeCommandParse) => {
+        for {
+          _ <- messageIfAsked(helpMessage, commonParse.help)
+          _ <- messageIfAsked(usageMessage, commonParse.usage)
+          commonOpts <- commonParse.baseOrError.left.map(ParsingError.apply)
```
tlangs: This projection to the left happens often enough that I'm wondering if its worth it to pull it out into its own imported implicit class. Ex:
```scala
  implicit class ProjectParsingErrors(either: Either[String, CommonOptions]) {
    def parsingErrorProjection: Either[ParsingError, CommonOptions] = 
      either.left.map(ParsingError.apply)
  }
```
Not necessary, but it would make changing the way parsing error messages are handled easier.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -1,63 +1,254 @@
...
+  }
 
-    checkRemainingArgs(remainingArgs.args)
+  /**
+    * Utility for wrapping a help / usage message in our type
+    * infrastructure, for cleaner use in for-comprehensions.
+    */
+  private def messageIfAsked(message: String,
+                             asked: Boolean): Either[EarlyReturn, Unit] = {
+    if (asked) {
```
tlangs: if/else can be replaced with `Either.cond(predicate, right, left)`

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
```diff
@@ -0,0 +1,59 @@
...
+        value.isDefined,
+        value.get,
+        s"Unknown enum value $entry for type ${c.runtimeClass.getName} "
+      )
+    }
+  }
+
+  implicit val oauthBearerTokenParser: ArgParser[OAuth2BearerToken] = {
+    ArgParser.instance[OAuth2BearerToken]("token") { token =>
+      //no need for left since this should never fail
```
tlangs, line 37: Why should this never fail?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/CommandDispatch.scala
```diff
@@ -1,18 +1,20 @@
...
-import org.broadinstitute.clio.client.webclient.ClioWebClient
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
 class CommandDispatch(val webClient: ClioWebClient, val ioUtil: IoUtil)
     extends LazyLogging {
 
-  def execute(command: CommandType, webClient: ClioWebClient)(
+  def execute(command: ClioCommand, webClient: ClioWebClient)(
```
tlangs, line 17: 👍 for the rename. Much clearer.

tlangs (commented):
This seams like a lot of kinda obscure boilerplate code for command parsing.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
danxmoran: Because all strings are valid bearer tokens as far as parsing is concerned. (Whether or not Clio accepts the token is a different problem)

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
danxmoran: I tried adding an implicit conversion from `Either[String, X]` to `Either[EarlyReturn, X]`, but type inference wasn't smart enough to figure out all the places where it should be applied. I'll make an explicit method to wrap around the spots that need it.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
tlangs: But what if the constructor to OAuth2BearerToken fails?

danxmoran: It won't, it's a case class that takes a string with no validation: https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCredentials.scala#L46

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
danxmoran: I like braces for anything > 1 line to avoid Python-esque "but where does this block end" adventures. Not really a problem here but I kept them for consistency.

danxmoran:
@tlangs comments addressed!

tlangs (approved):
👍

PR #102, opened 2017-09-05
DSDEGP-1500 Fail fast if configured bucket isn't writeable.

danxmoran:
### Description

This fixes a bug I found from #95. The Google library we're using to write to GCS assumes all directories exist and treats "create directory" calls as no-ops, so we won't detect bad GCS config until the first time Clio tries to write an update into a bucket. To still have the server fail fast on bad config, I've set it up to write the server's current version into the bucket root at startup.

----

* Review cycle:


tbl3rd (approved):
Looks good, but why not just write the version file in the Local case too, and just parameterize the path? (I'd give up the extra diagnostics in the Local case.)

tlangs (approved):
I don't see any issues.

PR #104, opened 2017-09-06
DSDEGP-1549 Port existing integration tests to full clio-client.

danxmoran:
### Description

Another stand alone chunk of DSDEGP-1549, this ports our existing integration tests to use the add and query commands of the clio-client. It also adds commands for getting server health and version, and adds a `--include-deleted` flag to the query command for hitting the /queryall endpoint.

Tests for the move and delete CLPs will follow after this merges.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/GetServerVersionExecutor.scala
```diff
@@ -0,0 +1,24 @@
...
+object GetServerVersionExecutor extends Executor {
+  override def execute(webClient: ClioWebClient, ioUtil: IoUtil)(
+    implicit ec: ExecutionContext,
+    bearerToken: OAuth2BearerToken
+  ): Future[HttpResponse] = {
+    for {
+      response <- webClient.getClioServerVersion
+      version <- webClient.unmarshal[Json](response)
+    } yield {
+      println(version)
```
pshapiro4broad: Any way to refactor common code here? The two methods appear to be identical except for the `webClient.getClioXXX` line.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/QueryWgsUbamExecutor.scala
```diff
@@ -17,7 +17,8 @@ class QueryWgsUbamExecutor(queryWgsUbam: QueryWgsUbam) extends Executor {
   ): Future[HttpResponse] = {
     for {
       response <- webClient.queryWgsUbam(
-        input = queryWgsUbam.transferWgsUbamV1QueryInput
+        input = queryWgsUbam.transferWgsUbamV1QueryInput,
```
pshapiro4broad: Why not pass in `queryWgsUbam` directly? Would it be a problem to make `ClioWebClient.queryWgsUbam()` accept a `QueryWgsUbam`? That would avoid the need to change this code if the parameter list for the API changes again. It could also provide another opportunity for code refactoring.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -91,16 +91,19 @@ class ClioWebClient(
...
-    input: TransferWgsUbamV1QueryInput
+    input: TransferWgsUbamV1QueryInput,
+    includeDeleted: Boolean = false
   )(implicit bearerToken: OAuth2BearerToken): Future[HttpResponse] = {
     val entity =
       HttpEntity(
         ContentTypes.`application/json`,
         input.asJson.pretty(implicitly[Printer])
       )
+    val queryPath = if (includeDeleted) "queryall" else "query"
```
pshapiro4broad, line 102: Was it a mistake to make these two different endpoints? It seems like we're doing the work to convert boolean -> API here and then API -> boolean in the server code.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/BasicTests.scala
```diff
@@ -27,7 +27,7 @@ trait BasicTests { self: BaseIntegrationSpec =>
 
   it should "report the server version of this test at /version" in {
     for {
-      response <- clioWebClient.getClioServerVersion
+      response <- runClient("get-server-version")
```
pshapiro4broad: Is there no constant for the magic string `get-server-version`?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -32,6 +24,24 @@ import java.util.UUID
 /** Tests of Clio's wgs-ubam functionality. */
 trait WgsUbamTests { self: BaseIntegrationSpec =>
 
+  def runUpsert(key: TransferWgsUbamV1Key,
+                metadata: TransferWgsUbamV1Metadata): Future[UUID] = {
+    val tmpMetadata = writeTmpJson(metadata)
+    runClient(
+      WgsUbamTests.upsert,
```
pshapiro4broad: This client command has a constant, seems like the others should too. But putting the constant in test code doesn't seem right, is it possible to put it in the code that actually determines what this value is in production?

danxmoran: The actual value is auto-magically determined by caseapp. It's possible to change the value with an annotation, but from [this issue](https://github.com/alexarchambault/case-app/issues/6) it sounds like the macros behind the annotations can only deal with literal strings.

I'll play with the annotations to double-check the issue (it's a few years old), and also check if we can get the derived class names from the implicits that drive the whole thing. Worst case, I'll move the magic string definitions into the client with a big warning comment.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
danxmoran: I don't think it'd be any less work to use a query parameter instead. We'd get rid of this `if`-`else` but then gain a new one in the server to check whether or not to modify the query input.

pshapiro4broad: This would let us delete an entire endpoint in the server, so the server would now have less code.

If we changed this to send `DocumentStatus` instead of hitting two endpoints then we wouldn't need to do any work on the server side either.

danxmoran: Good point, I didn't think about the fact that we'd get to delete the queryall directive and its associated boilerplate.

I think we still want the query vs queryall logic to happen in the server (vs. the client) so query results are the same regardless of the client used / without needing to duplicate the logic across all clients (which is a real worry since ops would rather use an auto-generated Python client).

In any case, I don't think changing the server API is in scope for this PR.

danxmoran:
@pshapiro4broad comments addressed!

jacarey (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
pshapiro4broad: Clients already need logic to determine which endpoint to hit. How is this different than logic that would provide a different argument to an endpoint?

Regardless of what's done for python, changing this to use `DocumentStatus` seems like a clear win to me. We have to provide logic anyway, may a well specify the data the server needs directly.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -25,28 +25,51 @@ import shapeless.CNil
...
 ) extends ClioCommand
 
 object ClioCommand extends ClioParsers {
 
+  // Names for generic commands.
+  val getServerHealthName = "get-server-health"
+  val getServerVersionName = "get-server-version"
+
+  // Names for WGS uBAM commands.
+  val getWgsUbamSchemaName = "get-wgs-ubam-schema"
```
pshapiro4broad: As in my comments on @ahakone 's checkin, I think we should consider putting the different bit of the name at the start or the end of the string instead of the middle. E.g. `wgs-ubam-get-schema` and `wgs-ubam-add`. Or maybe what we want is `add wgs-ubam` E.g. the command line args are `<operation>` `<type>`. Am just thinking ahead to the point where we have 5-6 indexes instead of just two.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/RetrieveAndPrintExecutor.scala
```diff
@@ -0,0 +1,44 @@
...
+/**
+  * Executor for all commands that retrieve some data from the
+  * clio-server and then print it as JSON.
+  */
+class RetrieveAndPrintExecutor(command: ClioCommand) extends Executor {
+  override def execute(webClient: ClioWebClient, ioUtil: IoUtil)(
+    implicit ec: ExecutionContext,
+    bearerToken: OAuth2BearerToken
+  ): Future[HttpResponse] = {
+    val responseFut = command match {
```
pshapiro4broad, line 21: This seems much better but is there a way to put the `match` inside `ClioCommand`?

pshapiro4broad (approved):
A few more nitpicks but 👍 overall

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
danxmoran: I'll shuffle it around to match her PR. Agreed `<operation>` `<type>` would be ideal and, we should try moving to it between the gvcf stuff and whatever comes next! But I wouldn't be surprised if it takes another epic adventure with shapeless to make it work.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/RetrieveAndPrintExecutor.scala
danxmoran: We could add a method to the `ClioCommand` trait, but it's more idiomatic to separate the data definitions and the operations over them. `ClioCommand` also has to be sealed for caseapp to work, so I'm trying to keep as much code out of that file as possible to minimize its bloat as we add more commands.

pshapiro4broad: ok

pshapiro4broad (approved):

PR #105, opened 2017-09-07
Phase 2 GVCF index

ahakone:
### Description
JIRA tickets: [DSDEGP-1241](https://broadinstitute.atlassian.net/browse/DSDEGP-1241), [DSDEGP-1244](https://broadinstitute.atlassian.net/browse/DSDEGP-1244), [DSDEGP-1245](https://broadinstitute.atlassian.net/browse/DSDEGP-1245), [DSDEGP-1491](https://broadinstitute.atlassian.net/browse/DSDEGP-1491), and [DSDEGP-1492](https://broadinstitute.atlassian.net/browse/DSDEGP-1492)

GVCF index, including unit tests and integration tests. All tests pass when run in local docker and docker-compose. 

The main difference between the ubams & gvcf is are the index fields. Please see [DocumentGvcf.scala](https://github.com/broadinstitute/clio/blob/ah_dsdegp_1491_gvcf_index/clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentGvcf.scala) for the fields & their types. 

----

* Review cycle:


danxmoran:
I think you need to add the new index to [this call](https://github.com/broadinstitute/clio/blob/develop/clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala#L64) too.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -26,24 +21,37 @@ import shapeless.CNil
 sealed trait ClioCommand
 
 case object GetWgsUbamSchema extends ClioCommand
+case object GetGvcfSchema extends ClioCommand
```
pshapiro4broad: I think it would be easier to read if we grouped names defined here by  index, not be operation. E.g. all the WgsUbam names (types/classes/etc) grouped together, followed by all the Gvcf names.

Also I think we want a blank line between each defined name. Or is that not the convention in Scala?

pshapiro4broad: It also occurs to me that, in order to find all the related names easily, it would be better to put the unique part of the name at the end or the beginning. E.g. `GetSchemaGvcf` or GvcfGetSchema`. The same rule would apply to `Input`, `Metadata`, etc. I think having the unique part first might be better since they would all sort together.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/GvcfWebService.scala
```diff
@@ -0,0 +1,71 @@
...
+      post {
+        entity(as[TransferGvcfV1QueryInput]) { input =>
+          complete(gvcfService.queryAllMetadata(input))
+        }
+      }
+    }
+  }
+
+  private[webservice] val gvcfGetSchema: Route = {
+    pathPrefix("schema") {
```
danxmoran: This should be `path` (just in case)

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferGvcfV1QueryInput.scala
```diff
@@ -0,0 +1,10 @@
+package org.broadinstitute.clio.transfer.model
+
+import org.broadinstitute.clio.util.model.{DocumentStatus, Location}
+
+case class TransferGvcfV1QueryInput(documentStatus: Option[DocumentStatus] =
+                                      None,
```
danxmoran, line 6: Strange indentation here

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferGvcfV1QueryOutput.scala
```diff
@@ -0,0 +1,21 @@
...
+  documentStatus: Option[DocumentStatus] = None,
+  gvcfMd5: Option[String] = None,
+  gvcfPath: Option[String] = None,
+  gvcfSize: Option[Long] = None,
+  gvcfMetricsPath: Option[String] = None,
+  location: Location,
+  pipelineVersion: Option[String] = None,
+  project: String,
+  sampleAlias: String,
+  version: Int,
```
danxmoran: Can you move the non-optional parameters to the top? It's the pattern we use in the wgs-ubam models.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentGvcf.scala
```diff
@@ -0,0 +1,22 @@
+package org.broadinstitute.clio.server.dataaccess.elasticsearch
+
+import java.time.OffsetDateTime
+import java.util.UUID
+
+import org.broadinstitute.clio.util.model.{DocumentStatus, Location}
+
+case class DocumentGvcf(analysisDate: Option[OffsetDateTime],
+                        clioId: UUID,
```
danxmoran: Same here, can you move the non-optional fields to the top?

danxmoran (commented):
Noticed a few things, overall looks good as far as I can tell!

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/GvcfWebService.scala
ahakone: YES

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferGvcfV1QueryInput.scala
ahakone: I think that's scalafmt because I keep on fixing it, and it keeps on reformatting whenever I run the tests :( But I'll make sure I fix it before committing

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/GetGvcfSchemaExecutor.scala
```diff
@@ -0,0 +1,26 @@
+package org.broadinstitute.clio.client.commands
```
danxmoran: This should be in package `org.broadinstitute.clio.client.dispatch` (it's a bug in the existing code for wgs-ubams, getting fixed in #104).

clio-client/src/test/resources/testdata/gvcfBadmetadata
```diff
@@ -0,0 +1 @@
+you can't parse this
```
danxmoran: Could we just have a generic "bad metadata" file that all the parsing tests could use, instead of having separate ones for wgs-ubam, gvcf, etc.?

clio-client/src/test/resources/testdata/gvcfMetadata
```diff
@@ -0,0 +1,10 @@
+{
+    "analysisDate" : "2017-08-09T14:21:57.995-04:00",
+    "contamination" : 0.75,
+    "gvcfMd5" : "gvcfMd5",
```
danxmoran: This shouldn't be parseable by the client (we use snake case everywhere instead of camel case).

clio-client/src/test/resources/testdata/gvcfMetadataplusextrafields
```diff
@@ -0,0 +1,11 @@
+{
+    "analysisDate" : "2017-08-09T14:21:57.995-04:00",
+    "contamination" : 0.75,
+    "gvcfMd5" : "gvcfMd5",
+    "gvcfPath" : "gvcfPath",
+    "gvcfSize" : 1,
+    "gvcfMetricsPath": "gvcfMetricsPath",
```
danxmoran: Same here, these should be snake case.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -25,14 +20,14 @@ import shapeless.CNil
   */
 sealed trait ClioCommand
 
-case object GetWgsUbamSchema extends ClioCommand
+case object GetSchemaWgsUbam extends ClioCommand
```
danxmoran: This will change the command name to `'get-schema-wgs-ubam'`. It's hard-coded in the add executor now, so it should be updated.

clio-client/src/test/resources/testdata/gvcfMetadata
ahakone: Oops I missed the dispatch specs during the rebase so that's why it hasn't been yelling about parsing. Changing to snake case & adding the tests back now

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
ahakone: Done (in executor)

clio-client/src/main/resources/reference.conf
```diff
@@ -15,9 +15,10 @@ clio {
   }
   client {
     response-timeout = 30s
-  }
-  greenteam {
-    email = "greenteam@broadinstitute.org"
+    service-account-json = null
+    greenteam {
```
pshapiro4broad: Calling this `greeanteam` doesn't describe what it is, it describes what its value is. E.g. is this used for error email, or support, etc? In general using value instead of purpose when naming leads to cases where a name no longer matches its value. The same is true of the use of `green` in the code below.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -247,6 +247,13 @@ class ClioClient(webClient: ClioWebClient,
       _ <- checkRemainingArgs(args ++ args0)
     } yield {
       implicit val bearerToken: OAuth2BearerToken = commonOpts.bearerToken
```
pshapiro4broad: A lot of the code uses `OAuth2BearerToken` but it looks like a more general type, `HttpCredentials`, would work instead. That would be preferable to avoid many changes if we ever change the way we do authentication.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddGvcfExecutor.scala
```diff
@@ -0,0 +1,49 @@
...
+          err
+        )
+    }
+
+    val decodedOrError = parsedOrError
+      .flatMap(_.as[TransferGvcfV1Metadata])
+      .left
+      .map { err =>
+        new RuntimeException(
+          s"Invalid metadata given at $metadataLoc. Run the 'get-schema-gvcf' command to see the expected JSON format for Gvcfs.",
```
pshapiro4broad: Use a constant here for the command name

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddWgsUbamExecutor.scala
```diff
@@ -30,7 +30,7 @@ class AddWgsUbamExecutor(addWgsUbam: AddWgsUbam) extends Executor {
       .left
       .map { err =>
         new RuntimeException(
-          s"Invalid metadata given at $metadataLoc. Run the 'get-wgs-ubam-schema' command to see the expected JSON format for WgsUbams.",
+          s"Invalid metadata given at $metadataLoc. Run the 'get-schema-wgs-ubam' command to see the expected JSON format for WgsUbams.",
```
pshapiro4broad: Use a constant here too

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/CommandDispatch.scala
```diff
@@ -19,16 +19,26 @@ class CommandDispatch(val webClient: ClioWebClient, val ioUtil: IoUtil)
...
-      case add: AddWgsUbam =>
-        new AddWgsUbamExecutor(add).execute(webClient, ioUtil)
-      case query: QueryWgsUbam =>
-        new QueryWgsUbamExecutor(query).execute(webClient, ioUtil)
-      case move: MoveWgsUbam =>
-        new MoveWgsUbamExecutor(move).execute(webClient, ioUtil)
-      case delete: DeleteWgsUbam =>
-        new DeleteWgsUbamExecutor(delete).execute(webClient, ioUtil)
+      case GetSchemaWgsUbam =>
+        GetSchemaWgsUbamExecutor.execute(webClient, ioUtil)
```
pshapiro4broad: Unfortunately this still puts the differentiator (`WgsUbam`) in the middle of the name `GetSchema...Executor`. Am not sure how far we want to take the rule of putting the name at the start or the end, but it's something to consider here as well.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteGvcfExecutor.scala
```diff
@@ -0,0 +1,157 @@
...
+          )
+        deleteResponses <- deleteGvcfs(gvcf, webClient, ioUtil) logErrorMsg "There was an error while deleting some of all of the Gvcfs."
+      } yield {
+        deleteResponses.size match {
+          case 0 =>
+            throw new Exception(
+              "Deleted 0 Gvcfs. None of the Gvcfs queried were able to be deleted."
+            )
+          case s if s == gvcf.size =>
+            logger.info(s"Successfully deleted ${deleteResponses.size} Gvcfs.")
```
pshapiro4broad: may as well use `s` here since it's bound, e.g. `${s} Gvcfs`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteGvcfExecutor.scala
```diff
@@ -0,0 +1,157 @@
...
+                notes => s"$notes\n${deleteGvcf.metadata.notes.getOrElse("")}"
+              )
+              .orElse(deleteGvcf.metadata.notes)
+          )
+        )
+        .map(webClient.ensureOkResponse)
+        .logErrorMsg(
+          s"Failed to delete the Gvcf ${gvcf.prettyKey()} in Clio. " +
+            s"The file has been deleted in the cloud. " +
+            s"Clio now has a 'dangling pointer' to ${gvcf.gvcfPath.getOrElse("")}. " +
```
pshapiro4broad: The expression `gvcf.gvcfPath.getOrElse("")` appears many times in this method, it should be stored in a `val`.

pshapiro4broad (approved):
Looks 👍 overall, but I think we need a phase2 ticket to refactor; we don't want to cut & paste to do the next index.

clio-client/src/main/resources/reference.conf
ahakone: I think these changes weren't really part of the gvcf PR (probably a result of the rebasing). Maybe @tlangs knows what it's for?

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
ahakone: Also @tlangs 's changes I think

clio-client/src/main/resources/reference.conf
danxmoran: Yeah this is unrelated. But it's already in develop, so I'm not sure why it's showing up here again...

danxmoran (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
tlangs: Why are my changes showing up here? Weird...

tlangs (approved):
Awesome!

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
danxmoran: It's some weirdness with the commit history, we can fix it up when we go through this to resolve the merge conflicts.

danxmoran:
@pshapiro4broad @tlangs I've resolved the merge conflicts, do you want to take another look before I merge?

danxmoran:
I also opened a [ticket](https://broadinstitute.atlassian.net/browse/DSDEGP-1579) to de-duplicate this code before we add the next index.

PR #106, opened 2017-09-11
Tbl dsdegp 1433 backup es

tbl3rd:
DSDEGP-1433: Script Elasticsearch snapshot management for Clio.

Still need a PR for the Jenkins integration.

----

* Review cycle:

### DIFF
--- a/jenkins/snapshot-elasticsearch.sh
+++ b/jenkins/snapshot-elasticsearch.sh
@@ -1,0 +1,120 @@
+#!/bin/bash
+
+set -e
+
+# Show usage and fail if there is a problem with the command line.
+#
+function help {
+    local -r av0="$1" environment="$2" verb="$3"
+    local -r -a environments=(dev prod staging)
+    local ok=no t v
+    local -r -a tools=(curl date jq) verbs=(configure list take)
+    local -r -a  usage=(
+        ''
+        "$av0: Snapshot an Elasticsearch (Es) cluster into GCS."
+        ''
+        "Usage: $av0 <environment> <verb>"
+        ''
+        "Where: <environment> is one of: ${environments[*]}"
+        "       <verb> is one of: ${verbs[@]}."
+        "       configure configures snapshots in the environment"
+        "       list      lists the snapshots"
+        "       take      takes a snapshot now"
+        ''
+        "Example: $av0 dev list"
+        ''
+        "BTW, you ran: $*"
+        ''
+    )
+    for v in ${environments[*]}; do test $v = "$environment" && ok=yes; done
+    if test $ok = no
+    then
+        echo 1>&2 "${av0}: Unrecognized environment: '$environment'"
+    fi
+    ok=no
+    for v in ${verbs[*]}; do test $v = "$verb" && ok=yes; done
+    if test $ok = no
+    then
+        echo 1>&2 "${av0}: Unrecognized verb: '$verb'"
+    fi
+    for t in "${tools[@]}"; do which $t >/dev/null || ok=no; done
+    if test $ok = no
+    then
+        echo 1>&2 "${av0}: You need these tools on PATH: ${tools[*]}"
+    fi
+    if test $ok = no
+    then
+        for line in "${usage[@]}"; do echo "$line" 1>&2; done
+        exit 1
+    fi
+}
+
+# Report that $av0 is running the rest of arguments as a command line.
+#
+function echoRunJq () {
+    local -r av0="$1" ; shift
+    local -a command
+    local arg result
+    for arg in "$@"; do command+=("'$arg'"); done
+    echo $av0: Running "${command[@]}"
+    result=$("$@")
+    echo "$result" | jq .
+}
+
+# Configure snapshots to $bucket in $environment for $es cluster.
+#
+function snapshot_configure () {
+    local -r av0="$1" environment=$2 es=$3 repo=$4
+    local -r bucket=broad-gotc-$environment-clio-elasticsearch
+    local -r url=$es/_snapshot/$repo
+    local -r data='
+    {
+        "type": "gcs",
+        "settings": {
+            "bucket": "'$bucket'",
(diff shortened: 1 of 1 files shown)


jenkins/snapshot-elasticsearch.sh
```diff
@@ -0,0 +1,137 @@
...
+    done
+}
+
+function main () {
+    local -r av0="${0##*/}" environment="$1" verb="$2"
+    help "$av0" "$@" ; shift 2
+    local -r bucket=broad-gotc-$environment-clio-elasticsearch
+    local -r domain=gotc-$environment.broadinstitute.org
+    local -r es=http://elasticsearch1.$domain:9200
+    snapshot_$verb "$av0" $es $bucket "$@"
```
pshapiro4broad: Didn't know you could specify a function using a env var value. Do you find this easier to read than `case`?

jenkins/snapshot-elasticsearch.sh
```diff
@@ -0,0 +1,137 @@
...
+    local -r av0="$1" es=$2 bucket=$3
+    echoRunJq "$av0" curl -s $es/_snapshot/$bucket/_all
+}
+
+# Take a snapshot in $environment for $es cluster.
+#
+function snapshot_take () {
+    local -r av0="$1" es=$2 bucket=$3
+    local -r jq='keys|.[]|select(startswith(".")|not)'
+    local -r snapshot=clio-$(date -u +%Y-%m-%d-%H-%M-%S)
```
pshapiro4broad: Our usage is clio specific but in theory this could be used to backup any ES server, correct?

pshapiro4broad (approved):
Looks OK to me. Is there another PR for the jenkins job?

Do you have a log of running the backup and can point at the GCS location where the backup went to?

jenkins/snapshot-elasticsearch.sh
```diff
@@ -0,0 +1,137 @@
...
+#
+function snapshot_delete () {
+    local -r av0="$1" es=$2 bucket=$3 ; shift 3
+    for snap in "$@"
+    do
+        echoRunJq "$av0" curl -s -X DELETE $es/_snapshot/$bucket/$snap
+    done
+}
+
+function main () {
```
pshapiro4broad: Not sure if the name `snapshot` is best here as this does a few things. When we have a restore operation, would it go here as well? Or another script?

tbl3rd: Suggest a better name.

jenkins/snapshot-elasticsearch.sh
tbl3rd: Probably.
I don't want to generalize too soon, but could name snapshots just 'snap-...'.
(I'd prefix the snapshot names with the bucket name if it were shorter.

jenkins/snapshot-elasticsearch.sh
tbl3rd: Not an env var. It's just a local function.

jenkins/snapshot-elasticsearch.sh
tbl3rd: OK. Now I've generalized too soon. ;-)

jenkins/snapshot-elasticsearch.sh
```diff
@@ -0,0 +1,161 @@
...
+    local -r av0="${0##*/}" environment="$1" verb="$2"
+    help "$av0" "$@" ; shift 2
+    local -r bucket=$(gcsBucketForEnvionment $environment)
+    local -r domain=gotc-$environment.broadinstitute.org
+    local -r es=http://elasticsearch1.$domain:9200
```
hjfbynara: Looks like you are hard-coding a specific node (elasticsearch1)  in the cluster.  While I believe you can submit a snap request to any node - by specifying this specific one - snapshots will fail if that node is down - even though cluster will be functional.

I wonder if the cluster member to run this script against should be passed in to script.  Then outside of script you decide which node to send the request to.

tbl3rd: Good idea. Is there an authoritative list of nodes in the Elasticsearch cluster?

hjfbynara: well by convention the nodes are: elasticsearch1, 2, 3 (cause there are three in the cluster)  If we add more then the number will obviously increase.  So one idea is jenkins cycles through the list until it finds the "master" node and submits it to the master node.   Don't want to just stop at first that responds since in a split brain situation - the first one may be the outlier.  once it finds master it then submits that snapshot request.  

List definitely can go against any.  delete and restore - unclear - might be good to do against master.

hjfbynara (approved):

hjfbynara:
👍

jenkins/snapshot-elasticsearch.sh
```diff
@@ -47,29 +51,35 @@ function help () {
...
     then
         for line in "${usage[@]}"; do echo 1>&2 "$line"; done
         exit 1
     fi
 }
 
+# Echo ELASTICSEARCH_NODE_COUNT names of Elasticsearch nodes.
+#
+function elasticsearchNodes () {
+    local -i n last first=100
```
hjfbynara: the fact that the clusternodes are elasticsearch-<ENV>10X is sort of happenstance.  an artifact of puppet configuration we have in place.  IF I need to build a new node that is not exactly the same as current 10X nodes (like more mem, bigger disk, more CPU) - I must use a different number.  So it would a 20X cluster.  This was why I put in the elasticsearch1,2,3 DNS mappings.  So if today I built bigger servers - they will 201,202,203 and elasticsearch4,5,6.  so the assumption that first node is always 10X is not really valid.

tbl3rd: OK. I'll change first back to 1.

tbl3rd:
The snapshot-elasticsearch.sh script is now in gotc.git/scripts/.

PR #108, opened 2017-09-11
DSDEGP-1580 DSDEGP-1581 Fix inputs for move and delete CLPs.

danxmoran:
Noticed this while writing the integration tests. Currently, our move and delete CLPs accept any metadata field as input, but the move CLPs only care about `--xyz-path` and the delete CLPs only care about `--notes`. The problems with this are:

1. All the extra fields are ignored if given, which is confusing
2. The ignored fields are shown in `--help` output, which is noisy and confusing
3. All metadata fields are optional, so it's possible to delete without adding a note (which doesn't meet our requirements), and the move logic has a bunch of extra logic to deal with the case when no destination is given.

This PR rips out the metadata arguments to both CLPs, and replaces them with the arguments that are actually needed as non-options. Integration-tests for these missing options will follow (I'm still battling cloud storage permissions).

### Description

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorGvcf.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorGvcf.scala
@@ -4,7 +4,7 @@ import akka.http.scaladsl.model.HttpResponse
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import org.broadinstitute.clio.client.commands.DeleteGvcf
 import org.broadinstitute.clio.client.util.IoUtil
-import org.broadinstitute.clio.client.util.GvcfUtil.TransferGvcfV1QueryOutputUtil
+import org.broadinstitute.clio.client.util.GvcfUtil._
 import org.broadinstitute.clio.client.webclient.ClioWebClient
 import org.broadinstitute.clio.transfer.model.{
   TransferGvcfV1Key,
@@ -26,7 +26,7 @@ class DeleteExecutorGvcf(deleteGvcf: DeleteGvcf) extends Executor {
   ): Future[HttpResponse] = {
     if (!deleteGvcf.transferGvcfV1Key.location.equals(Location.GCP)) {
       Future
-        .failed(new Exception("Only GCP Gvcfs are supported at this time"))
+        .failed(new Exception("Only GCP gvcfs are supported at this time"))
     } else {
       for {
         queryResponses <- webClient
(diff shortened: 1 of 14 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorGvcf.scala
```diff
@@ -37,7 +37,8 @@ class DeleteExecutorGvcf(deleteGvcf: DeleteGvcf) extends Executor {
               project = Some(deleteGvcf.transferGvcfV1Key.project),
               sampleAlias = Some(deleteGvcf.transferGvcfV1Key.sampleAlias),
               version = Some(deleteGvcf.transferGvcfV1Key.version)
-            )
+            ),
+            includeDeleted = false
```
pshapiro4broad, line 41: Shouldn't this parameter default to `false`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutorGvcf.scala
```diff
@@ -87,31 +90,35 @@ class MoveExecutorGvcf(moveGvcfCommand: MoveGvcf) extends Executor {
       .map(webClient.ensureOkResponse)
       .flatMap(webClient.unmarshal[Seq[TransferGvcfV1QueryOutput]])
       .map(ensureOnlyOne)
-      .map(_.gvcfPath)
+      .map {
+        _.gvcfPath.getOrElse {
+          throw new Exception(
+            s"The gvcf for Key($prettyKey) has no registered path, and can't be moved."
```
pshapiro4broad: Odd spacing and use of parens: `Key($prettyKey)`. I would write `The gvcf for key $prettyKey ...` or `The gvcf for $prettyKey ...` since key is understood from context.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutorGvcf.scala
```diff
@@ -134,29 +142,13 @@ class MoveExecutorGvcf(moveGvcfCommand: MoveGvcf) extends Executor {
...
-    } yield {
-      if (locationError.isEmpty && pathError.isEmpty) {
-        Future.successful(())
-      } else {
-        Future.failed(new Exception(String.join(" ", locationError, pathError)))
-      }
+  private def verifyCloudPaths(ioUtil: IoUtil): Unit = {
+    if (moveGvcfCommand.transferGvcfV1Key.location != Location.GCP) {
+      throw new Exception("Only GCP gvcfs are supported at this time.")
+    } else if (!ioUtil.isGoogleObject(moveGvcfCommand.destination)) {
```
pshapiro4broad: An `else` after a `throw` has no effect and is redundant.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorGvcf.scala
```diff
@@ -123,35 +124,33 @@ class DeleteExecutorGvcf(deleteGvcf: DeleteGvcf) extends Executor {
           TransferGvcfV1Metadata(
             documentStatus = Option(DocumentStatus.Deleted),
             notes = gvcf.notes
-              .map(
-                notes => s"$notes\n${deleteGvcf.metadata.notes.getOrElse("")}"
-              )
-              .orElse(deleteGvcf.metadata.notes)
+              .map(notes => s"$notes\n${deleteGvcf.note}")
```
pshapiro4broad: Using `map` `orElse` this way seem confusing to me, is there a reason why `match { Some => ... None => ...}` is not used here?

pshapiro4broad (approved):
Looks good. Will be nice to get rid of the duplicate code.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorGvcf.scala
danxmoran: I got rid of the default because I wanted to be explicit here about not including deleted records. Now that the integration-tests have been ported away from using the web-client, these are the only spots that were affected by the change.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorGvcf.scala
danxmoran: Just a style preference; I almost always reach for `.map`, `.orElse`, etc. so that it's clean to insert more chained operations in the middle / end.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutorGvcf.scala
danxmoran: Was mirroring the messages elsewhere in the file, but agreed they're weird. I'll sweep through and fix them all up.

tlangs (approved):
If only there was some way to generalise this code so that we don't have to change so many files when refactoring 🤔

PR #110, opened 2017-09-12
DSDEGP-1549 Add integration tests for move and delete CLPs

danxmoran:
### Description

These run locally but are expected to fail in Jenkins because of gcloud permissions errors. I'll play around with the clio-integration job and open a parallel PR to update its job DSL once I find something that works.

----

* Review cycle:

### DIFF
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
@@ -36,8 +36,7 @@ import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
 
-import java.io.File
-import java.nio.file.{FileSystem, Files, Path}
+import java.nio.file.{FileSystem, Files, Path, Paths}
 import java.util.UUID
 
 /**
@@ -96,9 +95,9 @@ abstract class BaseIntegrationSpec(clioDescription: String)
   private val vaultPath = "secret/dsde/gotc/test/clio/clio-account.json"
 
   /** List of possible token-file locations, in order of preference. */
-  private val vaultTokenFiles = Seq(
-    new File("/etc/vault-token-dsde"),
-    new File(s"${System.getProperty("user.home")}/.vault-token")
+  private val vaultTokenPaths = Seq(
+    Paths.get("/etc/vault-token-dsde"),
+    Paths.get(System.getProperty("user.home"), ".vault-token")
   )
 
   /** Scopes needed from Google to write to our test-storage bucket. */
@@ -110,10 +109,10 @@ abstract class BaseIntegrationSpec(clioDescription: String)
     * Service account credentials for use when accessing cloud resources.
     */
   protected lazy val serviceAccount: ServiceAccount = {
-    val vaultToken: String = vaultTokenFiles
-      .find(_.exists)
-      .map { file =>
-        new String(Files.readAllBytes(file.toPath)).stripLineEnd
+    val vaultToken: String = vaultTokenPaths
+      .find(Files.exists(_))
+      .map { path =>
+        new String(Files.readAllBytes(path)).stripLineEnd
       }
       .getOrElse {
         sys.error("Vault token not found on filesystem!")
@@ -145,24 +144,31 @@ abstract class BaseIntegrationSpec(clioDescription: String)
   }
 
   /**
-    * Path to the cloud directory to which all GCP test data
-    * should be written.
+    * Get a path to the root of a bucket in cloud-storage,
+    * using Google's NIO filesystem adapter.
     */
-  lazy val rootTestStorageDir: Path = {
+  def rootPathForBucketInEnv(env: String, scopes: Seq[String]): Path = {
     val storageOptions = StorageOptions
       .newBuilder()
-      .setProjectId("broad-gotc-test-storage")
-      .setCredentials(serviceAccount.credentialForScopes(testStorageScopes))
+      .setProjectId(s"broad-gotc-$env-storage")
+      .setCredentials(serviceAccount.credentialForScopes(scopes))
       .build()
 
     val gcs: FileSystem = CloudStorageFileSystem.forBucket(
-      "broad-gotc-clio-test",
+      s"broad-gotc-$env-clio",
       CloudStorageConfiguration.DEFAULT,
       storageOptions
     )
 
     gcs.getPath("/")
   }
+
+  /**
+    * Path to the cloud directory to which all GCP test data
+    * should be written.
+    */
+  lazy val rootTestStorageDir: Path =
+    rootPathForBucketInEnv("test", testStorageScopes)
 
   /**
     * Path to the root directory in which metadata updates will
@@ -193,7 +199,9 @@ abstract class BaseIntegrationSpec(clioDescription: String)
         (Seq("--bearer-token", bearerToken.token, command) ++ args).toArray
       )
       .fold(
-        err => Future.failed(new Exception(s"Command exited early with $err")),
+        earlyReturn =>
+          Future
+            .failed(new Exception(s"Command exited early with $earlyReturn")),
         identity
       )
   }
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
@@ -185,15 +185,17 @@ abstract class BaseIntegrationSpec(clioDescription: String)
 
   /**
     * Run a command with arbitrary args through the main clio-client.
-    *
-    * Fails the test immediately if the client exits early.
+    * Returns a failed future if the command exits early.
     */
   def runClient(command: String, args: String*): Future[HttpResponse] = {
     clioClient
       .instanceMain(
         (Seq("--bearer-token", bearerToken.token, command) ++ args).toArray
       )
-      .fold(err => fail(s"Command exited early with $err"), identity)
+      .fold(
+        err => Future.failed(new Exception(s"Command exited early with $err")),
+        identity
+      )
   }
 
   /**
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
@@ -8,12 +8,19 @@ import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
   ElasticsearchIndex
 }
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
+import org.broadinstitute.clio.util.model.ServiceAccount
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import akka.stream.ActorMaterializer
 import akka.testkit.TestKit
+import com.bettercloud.vault.{Vault, VaultConfig}
+import com.google.cloud.storage.StorageOptions
+import com.google.cloud.storage.contrib.nio.{
+  CloudStorageConfiguration,
+  CloudStorageFileSystem
+}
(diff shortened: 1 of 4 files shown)


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -378,4 +378,148 @@ trait WgsUbamTests { self: BaseIntegrationSpec =>
...
+    }
+
+    result.andThen[Unit] {
+      case _ => {
+        val _ = Files.deleteIfExists(cloudPath)
+      }
+    }
+  }
+
+  it should "not delete wgs-ubams without a note" in {
```
tlangs: Are we also going to test that it appends instead of replacing notes?

danxmoran: Yes we should, I'll add it.

danxmoran:
@tlangs I added the note tests

tlangs (approved):
👌

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -1,20 +1,17 @@
 package org.broadinstitute.clio.integrationtest
```
tbl3rd, line 1: This looks like a vault purge (moved to BaseIntegrationSpec) and optimize imports.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -80,6 +89,81 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
+    accountJSON
+      .as[ServiceAccount]
+      .fold({ err =>
+        throw new RuntimeException(
+          s"Failed to decode service account JSON from Vault at $vaultPath",
+          err
+        )
+      }, identity)
+  }
+
```
tbl3rd, line 145: Do we prefer fold to match? (Asking for style guidance. (I prefer fold to map/get, FWIW.))

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -375,4 +375,172 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+        "--sample-alias",
+        randomId,
+        "--version",
+        "123"
+      )
+    }.map {
+      _.getMessage should include("--destination")
+    }
+  }
+
```
tbl3rd, line 442: FWIW: I still see complaints from clio-client about --sampleAlias instead of --sample-alias.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -375,4 +375,172 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+  def addAndDeleteGvcf(
+    deleteNote: String,
+    existingNote: Option[String]
+  ): Future[TransferGvcfV1QueryOutput] = {
+    val project = s"abcdefg$randomId"
```
tbl3rd: project instead of abcdefg? As in sample below, and elsewhere.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -375,4 +375,172 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+      new String(Files.readAllBytes(cloudPath2)) should be(fileContents)
+    }
+
+    result.andThen[Unit] {
+      case _ => {
+        val _ = Seq(cloudPath, cloudPath2).map(Files.deleteIfExists)
+      }
+    }
+  }
+
```
tbl3rd, line 424: Why not just ...
``` scala
result.andThen {
  case _ =>
    Seq(cloudPath, cloudPath2).map(Files.deleteIfExists)
}
```
... and so on.
Or are you just trying to be REALLY explicit about discarding the Booleans?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -20,14 +20,15 @@ import io.circe.Json
```
tbl3rd, line 20: FWIW: I just kinda skimmed this one.

tbl3rd (approved):
👍 Instructive, as always.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
tlangs: I think when there are only two cases, we prefer fold, especially if there's a clear "empty" case.

tbl3rd: Cool. Then why not just ...
```scala
accountJSON
  .as[ServiceAccount]
  .getOrElse(err => throw new RuntimeException(
    s"Failed to decode service account JSON from Vault at $vaultPath",
    err))
```

tbl3rd: Nevermind: That probably just returns a function on error.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -80,6 +89,81 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
 
+  /** URL of vault server to use when getting bearer tokens for service accounts. */
+  private val vaultUrl = "https://clotho.broadinstitute.org:8200/"
+
+  /** Path in vault to the service account JSON to use in testing. */
+  private val vaultPath = "secret/dsde/gotc/test/clio/clio-account.json"
+
+  /** List of possible token-file locations, in order of preference. */
+  private val vaultTokenFiles = Seq(
+    new File("/etc/vault-token-dsde"),
```
pshapiro4broad: I think in general we should prefer `Path` over `File`, since `Path` is the newer API.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -80,6 +89,81 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
+    */
+  lazy val rootTestStorageDir: Path = {
+    val storageOptions = StorageOptions
+      .newBuilder()
+      .setProjectId("broad-gotc-test-storage")
+      .setCredentials(serviceAccount.credentialForScopes(testStorageScopes))
+      .build()
+
+    val gcs: FileSystem = CloudStorageFileSystem.forBucket(
+      "broad-gotc-clio-test",
```
pshapiro4broad: I thought that the bucket format is `gs://broad-gotc-${ENV}-clio`? Is this named differently for some reason?

If it's different it should be documented in confluence, if it's the same it should share code with `EnvIntegrationSpec.rootPersistenceDir`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -101,15 +185,17 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
+    * Returns a failed future if the command exits early.
     */
   def runClient(command: String, args: String*): Future[HttpResponse] = {
     clioClient
       .instanceMain(
         (Seq("--bearer-token", bearerToken.token, command) ++ args).toArray
       )
-      .fold(err => fail(s"Command exited early with $err"), identity)
+      .fold(
+        err => Future.failed(new Exception(s"Command exited early with $err")),
```
pshapiro4broad: If `err` is an exception it should be passed to the `Exception()` constructor so you don't lose the original stack crawl.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -375,4 +375,172 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+        "--destination",
+        cloudPath2.toUri.toString
+      )
+    } yield {
+      Files.exists(cloudPath) should be(false)
+      Files.exists(cloudPath2) should be(true)
+      new String(Files.readAllBytes(cloudPath2)) should be(fileContents)
+    }
+
+    result.andThen[Unit] {
```
pshapiro4broad, line 417: This should be run even if the test fails or throws, in Java I'd use a `finally`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
danxmoran: ^ right, map/getOrElse wouldn't work here because getOrElse throws away the left value. 

For Options I usually reach for map/getOrElse because I built up the muscle memory for it before I learned Option.fold is a thing. I pretty much stopped using match after I started using map because I like having one pattern to think about for Futures, Options, Seqs, etc.

[Odersky weeps.](https://www.reddit.com/r/scala/comments/449td4/fold_or_pattern_matching_with_an_option/czp154s/)

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
danxmoran: Being really explicit because the compiler will error with non-Unit discarded value.

pshapiro4broad (approved):

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
pshapiro4broad: I created https://broadinstitute.atlassian.net/browse/DSDEGP-1589 for this issue

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
danxmoran: It's named differently because that's how Henry made it. I'll ask if it'd be a big pain to rename.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
danxmoran: Yup, `andThen` is the `finally` of the `Future` API.

pshapiro4broad: OK. I thought the `yield` body could be evaluated before `andThen()` is called. But I guess since it throws, it would have to be evaluated in the current thread to work properly.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
danxmoran: `err` is an [`EarlyReturn`](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala#L42) here, which will only happen if we screw up the args. I'll rename it to be more clear that it's not an exception.

PR #112, opened 2017-09-13
Fixed auth util erroring when service account json not found

tlangs:
### Description

AuthUtil was not falling back to its other options when the service account json path pointed to a non-existing file. This fixes that.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
@@ -62,15 +62,21 @@ object ClioClient extends LazyLogging {
 
     client
       .instanceMain(args)
-      .fold({
-        case UsageOrHelpAsked(message) => {
-          println(message)
-          Future.successful(())
-        }
-        case ParsingError(error) => {
-          Future.failed(new RuntimeException(error))
-        }
-      }, identity)
+      .fold(
+        {
+          case UsageOrHelpAsked(message) => {
+            println(message)
+            Future.successful(())
+          }
+          case ParsingError(error) => {
+            Future.failed(new RuntimeException(error))
+          }
+          case AuthorizationError(error) => {
+            Future.failed(new RuntimeException(error))
+          }
+        },
+        identity
+      )
       .onComplete {
         case Success(_) => sys.exit(0)
         case Failure(ex) => {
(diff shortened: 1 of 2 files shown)


clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
```diff
@@ -27,22 +28,33 @@ object AuthUtil extends ModelAutoDerivation {
...
+          .getOrElse(
+            new AccessToken(
+              Process("gcloud auth print-access-token").!!.trim,
+              null
+            )
+          )
+      }.getOrElse(throw new RuntimeException("Could not get authorization"))
   }
 
   def loadServiceAccountJson(serviceAccountPath: Path): ServiceAccount = {
```
danxmoran: I would just have this method return a `Try`

clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
```diff
@@ -24,25 +25,31 @@ object AuthUtil extends ModelAutoDerivation {
...
     */
-  def getAccessToken(serviceAccountPath: Option[Path]): AccessToken = {
-    serviceAccountPath
-      .map { jsonPath =>
-        loadServiceAccountJson(jsonPath)
-      }
-      .fold(
-        new AccessToken(Process("gcloud auth print-access-token").!!.trim, null)
-      )(getCredsFromServiceAccount)
+  def getAccessToken(serviceAccountPath: Option[Path]): Try[AccessToken] = {
```
danxmoran: Oops, lost in translation. I didn't want this method to return a `Try`, I wanted `loadServiceAccountJson` to. You should be able to just put `Try` before the opening curly of its body.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -42,6 +43,7 @@ object ClioClient extends LazyLogging {
   sealed trait EarlyReturn
   final case class ParsingError(error: String) extends EarlyReturn
   final case class UsageOrHelpAsked(message: String) extends EarlyReturn
+  final case class AuthorizationError(error: String) extends EarlyReturn
```
danxmoran: Instead of adding a new error type that doesn't really give us more info, could we just make the existing `ParsingError` more generic? All we need to know is if we should exit with 0 or 1, so distinguishing the types of errors isn't that valuable.

danxmoran (approved):

pshapiro4broad (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -60,15 +60,18 @@ object ClioClient extends LazyLogging {
...
+        {
+          case UsageOrHelpAsked(message) => {
+            println(message)
+            Future.successful(())
+          }
+          case ParsingOrAuthError(error) => {
+            Future.failed(new RuntimeException(error))
+          }
+        },
+        identity
```
danbills, line 73: could use `valueOr` from cats either syntax:
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L63

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -60,15 +60,18 @@ object ClioClient extends LazyLogging {
...
-        }
-      }, identity)
+      .fold(
+        {
+          case UsageOrHelpAsked(message) => {
+            println(message)
+            Future.successful(())
+          }
+          case ParsingOrAuthError(error) => {
+            Future.failed(new RuntimeException(error))
```
pshapiro4broad, line 70: I don't think a command line option error should result in an exception being thrown (which I think is what this will do?). We should print the error and exit with a non-zero status but not print an exception and a stack dump.

PR #113, opened 2017-09-15
DSDEGP-1578 Move document ids out of doc mappers and into doc types.

danxmoran:
### Description

Motivation:

If we have some method `getLatest(index: ESIndex)` on the search DAO that returns the latest `clioId` in Elasticsearch, and another method `getAllSince(id: UUID, index: ESIndex)` on the persistence DAO that returns all documents since that ID, we'd like to be able to do something like:

```scala
elasticsearchIndexes.foreach { index =>
  persistenceDAO.getAllSince(searchDAO.getLatest(index), index)
    .map(searchDAO.updateMetadata(_), index)
}
```

Right now that's not possible, because `updateMetadata` takes a third "id" field which is just the stringification of the fields of the key used to build the document. Currently the logic for building `id` lives in the document mappers and only works on our transfer key types. This PR moves the ids into the document types so we don't have to mess around with type mappers during recovery.

----

* Review cycle:

### DIFF
--- a/clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchDocumentMapperSpec.scala
+++ b/clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchDocumentMapperSpec.scala
@@ -17,19 +17,23 @@ class AutoElasticsearchDocumentMapperSpec extends FlatSpec with Matchers {
     DocumentMock
   ](() => expectedId)
 
+  val keyLong = 12345L
+  val keyString = "key"
+  val expectedEntityId = s"$keyLong.$keyString"
+
   it should "empty" in {
-    mapper.empty(ModelMockKey(12345L, "key")) should be(
+    mapper.empty(ModelMockKey(keyLong, keyString)) should be(
       DocumentMock(
         upsertId = expectedId,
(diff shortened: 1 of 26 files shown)


clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchDocumentMapperSpec.scala
```diff
@@ -17,14 +17,11 @@ class AutoElasticsearchDocumentMapperSpec extends FlatSpec with Matchers {
...
 
-  it should "id" in {
-    mapper.id(ModelMockKey(12345L, "key")) should be("12345.key")
-  }
-
   it should "empty" in {
     mapper.empty(ModelMockKey(12345L, "key")) should be(
       DocumentMock(
         clioId = expectedId,
+        entityId = "12345.key",
```
pshapiro4broad: Since this cross-references the data on lines 31 and 32, the code should use constants to make it clear that where the matched value is coming from. 

The use of "." here should also use a constant which would be defined in `AutoElasticsearchDocumentMapper` (or wherever it's used).

pshapiro4broad (approved):

tlangs (approved):
🛠 <- good work

PR #115, opened 2017-09-18
DSDEGP-1590 Gracefully handle errors when calling out to gcloud.

danxmoran:
### Description

Our error-handling in the client was letting an error slip through when we tried to get a bearer token by calling out to `gcloud` and `gcloud` wasn't on the path. While I was looking into it I found a couple more unhandled spots, so I ended up just wrapping everything that deals with gcloud in an `Either`.

I'm not sure of the best way to write a test for this since it depends on the PATH, but I verified manually that the CLP doesn't hang any more:

image

----

* Review cycle:

### DIFF
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
@@ -34,12 +34,16 @@ abstract class EnvIntegrationSpec(env: String)
     s"http://elasticsearch1.gotc-$env.broadinstitute.org:9200"
   )
 
-  override lazy implicit val bearerToken: OAuth2BearerToken =
-    OAuth2BearerToken(
-      AuthUtil
-        .getCredsFromServiceAccount(serviceAccount)
-        .getTokenValue
+  override lazy implicit val bearerToken: OAuth2BearerToken = {
+    val tokenOrErr = AuthUtil
+      .getCredsFromServiceAccount(serviceAccount)
+      .map(_.getTokenValue)
+
+    tokenOrErr.fold(
+      fail(s"Couldn't get access token for account $serviceAccount", _),
+      OAuth2BearerToken.apply
     )
+  }
 
   override lazy val rootPersistenceDir: Path =
     rootPathForBucketInEnv(env, storageScopes)
--- a/clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
+++ b/clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
@@ -1,13 +1,11 @@
 package org.broadinstitute.clio.util
 
 import java.nio.file.{Files, Path}
-
 import com.google.auth.oauth2.AccessToken
 import io.circe.parser.decode
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
 import org.broadinstitute.clio.util.model.ServiceAccount
 
-import scala.io.Source
 import scala.sys.process.Process
 import scala.util.Try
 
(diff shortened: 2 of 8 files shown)


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -34,12 +34,16 @@ abstract class EnvIntegrationSpec(env: String)
     s"http://elasticsearch1.gotc-$env.broadinstitute.org:9200"
   )
 
-  override lazy implicit val bearerToken: OAuth2BearerToken =
-    OAuth2BearerToken(
-      AuthUtil
-        .getCredsFromServiceAccount(serviceAccount)
-        .getTokenValue
+  override lazy implicit val bearerToken: OAuth2BearerToken = {
```
pshapiro4broad, line 37: It occurs to me that this should really be `val credentials: HttpCredentials` throughout our code, as it's only ever used as type `HttpCredentials` and someday we may want to use a different authentication system.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
```diff
@@ -34,12 +34,16 @@ abstract class EnvIntegrationSpec(env: String)
...
+    val tokenOrErr = AuthUtil
+      .getCredsFromServiceAccount(serviceAccount)
+      .map(_.getTokenValue)
+
+    tokenOrErr.fold(
+      fail(s"Couldn't get access token for account $serviceAccount", _),
+      OAuth2BearerToken.apply
```
pshapiro4broad, line 44: I don't understand the constructor is called here instead of in `AuthUtil`. It looks like every use of `AccessToken` outside of `AuthUtil` is only used to create an `OAuth2BearerToken`. Wouldn't it be simpler to have `AuthUtil` return the bearer token directly?

clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
```diff
@@ -19,44 +17,66 @@ object AuthUtil extends ModelAutoDerivation {
     "https://www.googleapis.com/auth/userinfo.email"
   )
 
+  /** Get an access token by calling out to gcloud. */
+  private def shellOutAuthToken: Either[Throwable, AccessToken] =
```
pshapiro4broad: I am not a scala expert but I thought the convention was for methods with side effects to have an empty param list instead of no param list at all.

clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
```diff
@@ -19,44 +17,66 @@ object AuthUtil extends ModelAutoDerivation {
...
+        error
+      )
+    }
   }
+
+  /**
+    * Request an access token for the given service account with the
+    * scopes needed to talk through Clio's OpenIDC proxy.
+    */
+  def getCredsFromServiceAccount(
```
pshapiro4broad, line 72: Can this be private?

pshapiro4broad (approved):
Just a few questions, looks 👍 overall

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/EnvIntegrationSpec.scala
danxmoran: It would be simpler. @tlangs had it set up that way originally but I asked him to change it because we have to add all of akka-http as a dependency to clio-util to make it happen, and I was hoping to keep that project's dependencies slim (though maybe we don't care, since we're building fat jars everywhere anyways).

clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
danxmoran: That's the convention, I'll change it.

clio-util/src/main/scala/org/broadinstitute/clio/util/AuthUtil.scala
danxmoran: Nope, it's used in `EnvIntegrationSpec` since we load the JSON from vault instead of from disk there.

jacarey (approved):

PR #118, opened 2017-09-19
DSDEGP-1577: Find params file documents upserted on and after upsertId.

tbl3rd:
### Description

----

* Review cycle:

### DIFF
--- a/clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
+++ b/clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
@@ -22,12 +22,17 @@ abstract class ElasticsearchIndex[Document] {
   lazy val rootDir: String = indexName.replaceAll("_", "-")
 
   /**
-    * The source-of-truth directory in which updates to this index should be persisted
-    * at this moment.
+    * Format the directory path for the indexed meta-data files.
+    */
+  lazy val dateTimeFormatter: DateTimeFormatter =
+    DateTimeFormatter.ofPattern("yyyy/MM/dd")
+
+  /**
+    * The source-of-truth directory in which updates to this index
+    * should be persisted now.
     */
   def currentPersistenceDir: String = {
-    val now =
-      OffsetDateTime.now().format(DateTimeFormatter.ofPattern("yyyy/MM/dd"))
+    val now = OffsetDateTime.now().format(dateTimeFormatter)
     s"$rootDir/$now"
   }
 
--- a/clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
+++ b/clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.clio.server.dataaccess.elasticsearch
 
 import java.time.OffsetDateTime
+import java.time.format.DateTimeFormatter
 
 import com.sksamuel.elastic4s.mappings.FieldDefinition
 import org.broadinstitute.clio.util.generic.FieldMapper
@@ -25,8 +26,9 @@ abstract class ElasticsearchIndex[Document] {
     * at this moment.
     */
   def currentPersistenceDir: String = {
-    val now = OffsetDateTime.now()
-    s"$rootDir/${now.getYear}/${now.getMonthValue}/${now.getDayOfMonth}"
+    val now =
+      OffsetDateTime.now().format(DateTimeFormatter.ofPattern("yyyy/MM/dd"))
+    s"$rootDir/$now"
   }
 
   /**
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -85,9 +85,9 @@ trait PersistenceDAO extends LazyLogging {
     }
 
   /**
-    * Return the GCS bucket metadata files for `index` upserted with
-    * or after `upsertId` decoded from JSON as `ClioDocument`s sorted by
-    * `upsertId`.
+    * Return the GCS bucket metadata files for `index` upserted on or
+    * after `upsertId`.  Decode the files from JSON as `ClioDocument`s
+    * sorted by `upsertId`.
     *
     * @param upsertId of the last known upserted document
     * @param index  to query against.
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -101,11 +101,21 @@ trait PersistenceDAO extends LazyLogging {
       logger.debug(s"Wrote document $document to ${written.toUri}")
     }
 
+  /**
+    * Throw or return the `ClioDocument` at `path`.
+    * @param path to a JSON-encoded `ClioDocument``
+    * @tparam D is the type of `ClioDocument` at `path`
+    * @return the `ClioDocument` at `path`
+    */
   def documentFromPath[D <: ClioDocument: Decoder](path: Path): D = {
     decode[D](new String(Files.readAllBytes(path))).fold(throw _, identity)
   }
 
-  def isParamsFile(path: Path) = {
+  /**
+    * @param path in some params `FileSystem`
+    * @return true if `path` is a "params file"
+    */
+  def isParamsFile(path: Path): Boolean = {
     val s = path.toString
     val parts = s.split(File.separator)
     val last = parts.last
@@ -114,6 +124,16 @@ trait PersistenceDAO extends LazyLogging {
     parts.length == 6 && isJson && hasUuid
   }
 
+  /**
+    * Return the `ClioDocument`s under `rootdir` upserted no earlier
+    * than `withId`.
+    *
+    * @param rootDir is path to a bucket of "params files"
+    * @param withId is the path of earliest "params file" to return
+    * @param ec is an `ExecutionContext`
+    * @tparam D is the type of `ClioDocument` expected under `rootDir`
+    * @return the `ClioDocument` at `withId` and all later ones
+    */
   def sinceId[D <: ClioDocument: Decoder](rootDir: Path, withId: Path)(
     implicit ec: ExecutionContext
   ): Future[Seq[D]] = {
@@ -124,13 +144,11 @@ trait PersistenceDAO extends LazyLogging {
       .filter(!before(_))
       .runWith(Sink.seq)
       .map(_.sortBy(before))
-    val result = Source
+    Source
       .fromFuture(after)
       .mapConcat(identity)
-      .map(p => { logger.info(s"p == ${p.toString}"); p })
       .via(Flow[Path].map(p => documentFromPath(p)))
       .runWith(Sink.seq[D])
-    result
   }
 
   /**
@@ -150,27 +168,15 @@ trait PersistenceDAO extends LazyLogging {
   )(implicit ec: ExecutionContext): Future[Seq[D]] = {
     val rootDir = rootPath.resolve(index.rootDir)
     val suffix = s"/${upsertId}.json"
-    val filesWithId = Directory
+    Directory
       .walk(rootDir)
       .filter(p => isParamsFile(p))
       .filter(_.toString.endsWith(suffix))
-      .limit(23)
       .runWith(Sink.seq)
-    val result = filesWithId
       .map(_.map(sinceId(rootDir, _)))
       .map(Future.sequence(_))
       .flatMap(identity)
       .map(_.flatMap(identity))
-    result
-      .map(_.size)
-      .foreach(n => {
-        if (n < 1) {
-          throw new RuntimeException(
-            s"No file in ${rootDir.normalize.toString} ends with '${suffix}'."
-          )
-        }
-      })
-    result
   }
 }
 
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -6,13 +6,18 @@ import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
   ClioDocument,
   ElasticsearchIndex
 }
-
 import com.sksamuel.elastic4s.Indexable
 import com.typesafe.scalalogging.LazyLogging
 
 import scala.concurrent.{ExecutionContext, Future}
-
 import java.nio.file.{Files, Path}
+import java.util.UUID
+
+import io.circe.Decoder
+import io.circe.parser._
+
+import scala.util.Try
+import scala.collection.JavaConverters._
 
 /**
   * Persists metadata updates to a source of truth, allowing
@@ -77,6 +82,43 @@ trait PersistenceDAO extends LazyLogging {
         indexable.json(document).getBytes
       )
       logger.debug(s"Wrote document $document to ${written.toUri}")
+    }
+
+  /**
+    * Return the GCS bucket metadata files for `index` upserted with
+    * or after `upsertId` decoded from JSON as `ClioDocument`s sorted by
+    * `upsertId`.
+    *
+    * @param upsertId of the last known upserted document
+    * @param index  to query against.
+    * @param ec is the implicit ExecutionContext
+    * @tparam D is a ClioDocument type with a JSON Decoder
+    * @return a Future Seq of ClioDocument
+    */
+  def getAllSince[D <: ClioDocument: Decoder](
+    upsertId: UUID,
+    index: ElasticsearchIndex[D]
+  )(implicit ec: ExecutionContext): Future[Seq[D]] =
+    Future {
+      val rootDirectory = rootPath.resolve(index.currentPersistenceDir)
+      val files = Try(Files.newDirectoryStream(rootDirectory))
+        .map(_.asScala)
+        .fold(throw _, identity)
+        .toSeq
+      val filesWithId = files.filter(_.endsWith(s"${upsertId}.json"))
+      val withId = filesWithId match {
+        case Seq(result) => result
+        case _ =>
+          throw new RuntimeException(
+            s"${filesWithId.size} files end with ${upsertId.toString}.json"
+          )
+      }
+      val (_, after) = files.span(path => !path.equals(withId))
+      after
+        .map(Files.readAllBytes)
+        .map(new String(_))
+        .map(decode[D])
+        .map(_.fold(throw _, identity))
     }
 }
 
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -143,7 +143,7 @@ trait PersistenceDAO extends LazyLogging {
           )
         }
       })
-      .flatMap(identity)
+      .flatten
   }
 }
 
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -1,7 +1,5 @@
 package org.broadinstitute.clio.server.dataaccess
 
-import java.io.File
-
 import org.broadinstitute.clio.server.ClioServerConfig
 import org.broadinstitute.clio.server.ClioServerConfig.Persistence
 import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
@@ -15,10 +13,9 @@ import scala.concurrent.{ExecutionContext, Future}
 import java.nio.file.{Files, Path}
 import java.util.UUID
 
-import akka.actor.ActorSystem
-import akka.stream.{ActorMaterializer, ActorMaterializerSettings, Supervision}
+import akka.stream.Materializer
 import akka.stream.alpakka.file.scaladsl.Directory
-import akka.stream.scaladsl.{Flow, Sink, Source}
+import akka.stream.scaladsl.Sink
 import io.circe.Decoder
 import io.circe.parser._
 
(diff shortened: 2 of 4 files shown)


clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -61,4 +60,39 @@ class PersistenceDAOSpec extends AsyncFlatSpec with Matchers {
...
+    documents.map(_.upsertId).toSet.size should be(documents.size)
+    result.flatMap(_.toVector should be(expected))
+  }
+
+  it should "fail unless upsertId is found exactly once" in {
+    val upsertId = UUID.randomUUID()
+    val dao = new MemoryPersistenceDAO()
+    val result = dao
+      .initialize(index)
+      .map(_ => dao.writeUpdate(DocumentMock.default, index))
```
danxmoran: This looks like it's only writing the document once, so it shouldn't throw... am I missing something?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +83,43 @@ trait PersistenceDAO extends LazyLogging {
...
+    * @tparam D is a ClioDocument type with a JSON Decoder
+    * @return a Future Seq of ClioDocument
+    */
+  def getAllSince[D <: ClioDocument: Decoder](
+    upsertId: UUID,
+    index: ElasticsearchIndex[D]
+  )(implicit ec: ExecutionContext): Future[Seq[D]] =
+    Future {
+      val rootDirectory = rootPath.resolve(index.currentPersistenceDir)
+      val files = Try(Files.newDirectoryStream(rootDirectory))
```
danxmoran: Is the ordering assumption of the output seq relying on `Files.newDirectoryStream` to return the names in order? If so, could you add a comment?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +83,43 @@ trait PersistenceDAO extends LazyLogging {
...
+        .toSeq
+      val filesWithId = files.filter(_.endsWith(s"${upsertId}.json"))
+      val withId = filesWithId match {
+        case Seq(result) => result
+        case _ =>
+          throw new RuntimeException(
+            s"${filesWithId.size} files end with ${upsertId.toString}.json"
+          )
+      }
+      val (_, after) = files.span(path => !path.equals(withId))
```
danxmoran: Not sure if it matters, but should we worry about the performance of taking multiple passes over the whole `Seq` of filenames?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +83,43 @@ trait PersistenceDAO extends LazyLogging {
...
+    * @param ec is the implicit ExecutionContext
+    * @tparam D is a ClioDocument type with a JSON Decoder
+    * @return a Future Seq of ClioDocument
+    */
+  def getAllSince[D <: ClioDocument: Decoder](
+    upsertId: UUID,
+    index: ElasticsearchIndex[D]
+  )(implicit ec: ExecutionContext): Future[Seq[D]] =
+    Future {
+      val rootDirectory = rootPath.resolve(index.currentPersistenceDir)
```
danxmoran: This will only get the backups for the day the UUID was added, not all days following the UUID.

danxmoran (changes requested):

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +83,43 @@ trait PersistenceDAO extends LazyLogging {
...
+      val files = Try(Files.newDirectoryStream(rootDirectory))
+        .map(_.asScala)
```
danxmoran: Thinking of performance too, `_.asScala` will pull the entire stream into memory (not sure if `Files.newDirectoryStream` is lazy in the first place though). It might be worth trying to use Alpakka's `Directory.ls` connector instead: http://developer.lightbend.com/docs/alpakka/current/file.html#directory

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
danxmoran: If we keep `Files.newDirectoryStream` instead of the Akka/Alpakka stream, I think we need to manually close the stream after we pull it into memory.

pshapiro4broad: According to the javadoc, `The elements returned by the iterator are in no specific order.`

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
pshapiro4broad: `newDirectoryStream()` is a lazy iterator, I think it reads one element ahead at a time.

I'm not sure how to do this in Scala, in java I'd use a single streaming operation, using `filter` to skip all files until you hit the one you want to recover from. Although this only works if the files are in insertion order.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
pshapiro4broad: Maybe someday scala will have try-with-resources...

Note that each call to `hasNext()` on the `DirectoryStream` result can result in an unchecked exception, if an I/O error occurs.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -61,4 +60,39 @@ class PersistenceDAOSpec extends AsyncFlatSpec with Matchers {
...
+    val result = dao
+      .initialize(index)
+      .map(
+        _ =>
+          documents
+            .map(document => dao.writeUpdate(document, index))
+      )
+      .flatMap(_ => dao.getAllSince(expected.head.upsertId, index))
+    documents.toSet.size should be(documents.size)
+    documents.map(_.upsertId).toSet.size should be(documents.size)
```
pshapiro4broad: It looks like these two asserts could be done on line 74 above, just after `documents` has been created, as they don't involve `expected` or `result`.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
tbl3rd: The write is just to initialize jimfs, which seems to need it.
The search is for some random UUID that won't be in the index.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Yep. I had to walk the tree.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Yep. I sort them now.

tbl3rd: Riding the Alpakka now.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: We gotta sort something in memory, but I reduced it somewhat.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: I'll worry when we discover a problem.

We could eliminate one pass and shorten another if we had the path of the first file to re-upsert.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
tbl3rd: OK. I liked all the asserts at the bottom where I expect them, but YMMV.
They are kind of redundant anyway, but I moved them earlier.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
tbl3rd: I couldn't get this test to work reliably so I removed it.
Feel free to help me resurrect it if you think it worthwhile.

tbl3rd (commented):
I give up.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -1,25 +1,47 @@
...
+import io.circe.Decoder
+import io.circe.parser._
 
 /**
   * Persists metadata updates to a source of truth, allowing
   * for playback during disaster recovery.
   */
 trait PersistenceDAO extends LazyLogging {
 
+  private implicit val system = ActorSystem("clio")
```
danxmoran: We've [already got an ActorSystem](https://github.com/broadinstitute/clio/blob/develop/clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala#L22) running the server, you shouldn't need to create another one here. There's an existing materializer too.

I'd either change `PersistenceDAO` to be an abstract class taking an implicit materializer, or add an implicit materializer parameter to the new method.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +100,84 @@ trait PersistenceDAO extends LazyLogging {
...
+    */
+  def getAllSince[D <: ClioDocument: Decoder](
+    upsertId: UUID,
+    index: ElasticsearchIndex[D]
+  )(implicit ec: ExecutionContext): Future[Seq[D]] = {
+    val rootDir = rootPath.resolve(index.rootDir)
+    val suffix = s"/${upsertId}.json"
+    Directory
+      .walk(rootDir)
+      .filter(p => isParamsFile(p))
```
danxmoran: Do we need this filter?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +100,84 @@ trait PersistenceDAO extends LazyLogging {
...
+      .filter(_.toString.endsWith(suffix))
+      .runWith(Sink.seq)
```
danxmoran: `Sink.head` will complete on the first element that passes the filter, might make things simpler below.

danxmoran: (But then we really give up detection of duplicate IDs)

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +100,84 @@ trait PersistenceDAO extends LazyLogging {
...
+    * @return the `ClioDocument` at `withId` and all later ones
+    */
+  def sinceId[D <: ClioDocument: Decoder](rootDir: Path, withId: Path)(
+    implicit ec: ExecutionContext
+  ): Future[Seq[D]] = {
+    val before = (p: Path) => p.compareTo(withId) < 0
+    val after = Directory
+      .walk(rootDir)
+      .filter(p => isParamsFile(p))
+      .filter(!before(_))
```
danxmoran: Can we move the `.map(p => documentFromPath(p))` up to be right under this line, then sort by upsertId directly? Then I think we should only need one `.runWith`

danxmoran (changes requested):

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -26,8 +26,9 @@ abstract class ElasticsearchIndex[Document] {
     * at this moment.
     */
   def currentPersistenceDir: String = {
-    val now = OffsetDateTime.now()
-    s"$rootDir/${now.getYear}/${now.getMonthValue}/${now.getDayOfMonth}"
+    val now =
+      OffsetDateTime.now().format(DateTimeFormatter.ofPattern("yyyy/MM/dd"))
```
pshapiro4broad: Not sure about Scala, but in Java I would put a `DateTimeFormatter` in a static final since it's immutable and thread safe.

Also it might be worth documenting that the `/` here are GCS path separators.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -26,8 +26,9 @@ abstract class ElasticsearchIndex[Document] {
     * at this moment.
```
pshapiro4broad: Is our source-of-truth storage strategy described somewhere? I looked in confluence but didn't see it.

There is https://github.com/broadinstitute/tbl/blob/master/clio/es.org#what-source-of-truth-should-we-maintain-to-rebuild-clio but it's more of a design doc and not a spec for what is implemented.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +100,84 @@ trait PersistenceDAO extends LazyLogging {
...
+  }
+
+  /**
+    * @param path in some params `FileSystem`
+    * @return true if `path` is a "params file"
+    */
+  def isParamsFile(path: Path): Boolean = {
+    val s = path.toString
+    val parts = s.split(File.separator)
+    val last = parts.last
```
pshapiro4broad: I think this is `path.getFileName.toString`

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +100,84 @@ trait PersistenceDAO extends LazyLogging {
...
+    val isJson = last.endsWith(".json")
+    val hasUuid = last.split('-').length == 5
+    parts.length == 6 && isJson && hasUuid
```
pshapiro4broad: And `parts.length` is `path.getNameCount`

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Tried the abstract class idea yesterday but got an assertion failure I couldn't figure out.
Adding another implicit to the methods works though.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Nope. At least not when running against jimfs.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: I'd really like to catch and test that case, but this code is not doing it anyway.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: isParamsFile() is gone now.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Yep.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Unfortunately, I added a materializer to the Spec class.
Scalatest must have one, or at least an ActorSystem, but I don't know how to get at it.
Maybe you do?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Good to know ... again.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
danxmoran: If you make the spec extend `TestKit(ActorSystem("some-name")) with AsyncFlatSpecLike` instead of `AsyncFlatSpec`, you'll get an actor system.

danxmoran: You'll also want to throw a `with BeforeAndAfterAll` on there, with:
```scala
override def afterAll(): Unit = {
    shutdown()
    super.afterAll()
  }
```
to make sure the actor system dies properly.

tbl3rd:
With your help I was able to re-instate a test of the failure case.
Still don't like the materializer handling in the Spec class.  Ideas?

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -1,20 +1,22 @@
...
 import io.circe.parser._
 import org.scalatest.{AsyncFlatSpec, Matchers}
-
 import java.nio.file.Files
+import java.util.UUID
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
 
 class PersistenceDAOSpec extends AsyncFlatSpec with Matchers {
```
danxmoran: The way we do this elsewhere is to change `AsyncFlatSpec` to `TestKit(ActorSystem("some-name")) with AsyncFlatSpecLike with BeforeAndAfterAll`. The test kit class will handle making the system implicit, and also give you a `shutdown()` method to call in `afterAll()` to make sure everything cleans up correctly. Example [here](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/test/scala/org/broadinstitute/clio/client/BaseClientSpec.scala).

tbl3rd: OK, thanks. I still had to construct a private implicit materializer though.

tbl3rd:
How now?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +83,68 @@ trait PersistenceDAO extends LazyLogging {
...
+        val count = paths.size
+        if (count == 1) {
+          sinceId[D](rootDir, paths.head)
+        } else {
+          throw new RuntimeException(
+            s"${count} files end with ${suffix} in ${rootDir.toString}"
+          )
+        }
+      })
+      .flatMap(identity)
```
danxmoran: Does `.flatten` work instead?

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -1,20 +1,22 @@
...
+import org.scalatest.{AsyncFlatSpecLike, BeforeAndAfterAll, Matchers}
 import java.nio.file.Files
+import java.util.UUID
 
-class PersistenceDAOSpec extends AsyncFlatSpec with Matchers {
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.testkit.TestKit
+
+class PersistenceDAOSpec extends TestKit(ActorSystem("PeristenceDAOSpec")) with AsyncFlatSpecLike with BeforeAndAfterAll with Matchers {
```
danxmoran: In need of a scalafmt-ing

danxmoran (approved):

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
tbl3rd: flatten works here ... but not elsewhere ... hm

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -78,6 +83,68 @@ trait PersistenceDAO extends LazyLogging {
...
+    */
+  def getAllSince[D <: ClioDocument: Decoder](upsertId: UUID,
+                                              index: ElasticsearchIndex[D])(
+    implicit ec: ExecutionContext,
+    materializer: Materializer
+  ): Future[Seq[D]] = {
+    val rootDir = rootPath.resolve(index.rootDir)
+    val suffix = s"/${upsertId}.json"
+    val filesWithId = Directory
+      .walk(rootDir)
```
pshapiro4broad, line 132: Is there a way to write this without walking the same `rootDir` twice?

pshapiro4broad (approved):

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
tbl3rd: Probably.

PR #119, opened 2017-09-19
DSDEGP-1602 First pass at adding publishing.

danxmoran:
### Description

Still working on the Jenkins side of things, but this is the setup needed to point `sbt publish` at our Artifactory.

----

* Review cycle:


clio-util/src/main/scala/org/broadinstitute/clio/util/generic/CaseClassMapper.scala
```diff
@@ -43,7 +43,7 @@ class CaseClassMapper[T: ClassTag] {
   /** Creates an instance of T from a Map of field names to values. */
   def newInstance(vals: Map[String, _]): T = {
 
-    /** When there is no value for an field, default the value as None, or throw an error if the type isn't Option. */
+    /* When there is no value for a field, default the value as None, or throw an error if the type isn't Option. */
```
pshapiro4broad, line 46: Why is this not javadoc/scaladoc?

build.sbt
```diff
@@ -47,6 +49,7 @@ val commonTestDockerSettings: Seq[Setting[_]] = Seq(
 lazy val clio = project
   .in(file("."))
   .settings(commonSettings)
+  .settings(commonNoPublishSettings)
```
pshapiro4broad: Any reason to not put these settings in `commonSettings`? I assume we can override the settings for the projects that need deployment enabled. Either that, or only add the publish plugin to projects that will be publishing.

project/src/main/scala/org/broadinstitute/clio/publish/ClioPublishingPlugin.scala
```diff
@@ -0,0 +1,35 @@
...
+  private def artifactoryResolver(isSnapshot: Boolean): Resolver = {
+    val repoType = if (isSnapshot) "snapshot" else "release"
+    val repoUrl =
+      s"https://$artifactoryHost/broadinstitute/libs-$repoType-local;build.timestamp=$buildTimestamp"
+    val repoName = "artifactory-publish"
+    repoName at repoUrl
+  }
+
+  private val artifactoryCredentials: Credentials = {
+    val username = sys.env.getOrElse("ARTIFACTORY_USERNAME", "")
```
pshapiro4broad: Does this work with an empty password? If not, it should explicitly fail instead of returning the empty string.

pshapiro4broad: Can this get the username/password from vault?

project/src/main/scala/org/broadinstitute/clio/publish/ClioPublishingPlugin.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import sbt._
+import sbt.Keys._
+
+/**
+  * A plugin for adding the settings required for clio subprojects to Artifactory.
+  *
+  * Enable this plugin on every sub-project that should be published.
+  */
+object ClioPublishingPlugin extends AutoPlugin {
```
pshapiro4broad: I don't think adding `Clio` to this class adds to its description, as we don't have any other publishing plugins, and the class is already in a package/project called clio.

A more descriptive name might be `ArtifactoryPublishingPlugin`.

project/src/main/scala/org/broadinstitute/clio/publish/ClioPublishingPlugin.scala
```diff
@@ -0,0 +1,35 @@
...
+    val repoUrl =
+      s"https://$artifactoryHost/broadinstitute/libs-$repoType-local;build.timestamp=$buildTimestamp"
+    val repoName = "artifactory-publish"
+    repoName at repoUrl
+  }
+
+  private val artifactoryCredentials: Credentials = {
+    val username = sys.env.getOrElse("ARTIFACTORY_USERNAME", "")
+    val password = sys.env.getOrElse("ARTIFACTORY_PASSWORD", "")
+    Credentials("Artifactory Realm", artifactoryHost, username, password)
```
pshapiro4broad: Is `"Artifactory Realm"` a magic string here?

pshapiro4broad (approved):

clio-util/src/main/scala/org/broadinstitute/clio/util/generic/CaseClassMapper.scala
danxmoran: Scaladoc (apparently) doesn't bind to nested methods, so this was causing errors to print during `sbt publish`.

pshapiro4broad: whaat

build.sbt
danxmoran: If I put them in common settings, I think the order-of-operations for applying settings will cause them to override the publishing plugin's settings in the projects that need it (since plugins are applied before anything in build.sbt).

These settings aren't really necessary as long as you don't run `sbt publish` from the top-level project. If you do without these settings, SBT will try to publish everything, and complain about missing settings in the projects without the plugin. These settings tell SBT "no really, don't even try to publish this".

project/src/main/scala/org/broadinstitute/clio/publish/ClioPublishingPlugin.scala
danxmoran: That's the plan!

danxmoran: If we explicitly fail here, SBT will crash across the board unless ARTIFACTORY_USERNAME and ARTIFACTORY_PASSWORD are both set (credentials calculation isn't lazy). I'll have it log out a warning if the env vars are missing.

pshapiro4broad: Maybe it should not create and add credentials in that case? E.g. return `Option[Credentials]` and only `+=` it below if it's set?

We'd only want the warning/error printed if the user requested a publish operation. Otherwise it shouldn't matter if the env vars aren't set.

pshapiro4broad (approved):

clio-util/src/main/scala/org/broadinstitute/clio/util/generic/CaseClassMapper.scala
tbl3rd: It's spelled "wat".

PR #121, opened 2017-09-22
DSDEGP-1578 Clio recovers missing elasticsearch documents on startup

danxmoran:
### Description

Main app code + unit test is done, integration test is WIP.

----

* Review cycle:


clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -22,17 +22,11 @@ abstract class ElasticsearchIndex[Document] {
...
-  lazy val dateTimeFormatter: DateTimeFormatter =
-    DateTimeFormatter.ofPattern("yyyy/MM/dd")
-
-  /**
     * The source-of-truth directory in which updates to this index
     * should be persisted now.
     */
   def currentPersistenceDir: String = {
-    val now = OffsetDateTime.now().format(dateTimeFormatter)
+    val now = OffsetDateTime.now().format(ElasticsearchIndex.dateTimeFormatter)
```
pshapiro4broad, line 29: I think in the case of a companion object it's preferable to import all members rather than scope the usage. We probably do both things in our code though.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioMain.scala
```diff
@@ -1,9 +1,10 @@
 package org.broadinstitute.clio.server
 
-object ClioMain {
+import com.typesafe.scalalogging.LazyLogging
+
+object ClioMain extends LazyLogging {
```
pshapiro4broad: What effect does this have, if `ClioMain` doesn't log?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
```diff
@@ -62,31 +62,16 @@ class HttpElasticsearchDAO private[dataaccess] (
...
       searchResponse <- httpClient execute searchDefinition
     } yield {
-      searchResponse.to[D] match {
-        case Seq(document) => document
-        case _ =>
-          throw new RuntimeException(
-            s"Could not get most recent document for index ${index.indexName}, found ${searchResponse.hits.total} documents."
-          )
-      }
+      searchResponse.to[D].headOption
```
pshapiro4broad, line 74: Doesn't this return an incorrect result if multiple documents are found?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -98,15 +99,15 @@ trait PersistenceDAO extends LazyLogging {
     implicit ec: ExecutionContext,
     materializer: Materializer
   ): Future[Seq[D]] = {
-    val before = (p: Path) => p.compareTo(withId) < 0
+    val after = (p: Path) => p.compareTo(withId) > 0
```
pshapiro4broad: Not sure why you changed the sense here, seems like it's the same test either way.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SearchDAO.scala
```diff
@@ -54,4 +54,17 @@ trait SearchDAO {
     queryDefinition: QueryDefinition,
     index: ElasticsearchIndex[D]
   ): Future[Seq[D]]
+
+  /**
+    * Given an elastic search index, return the most recent document for that index.
+    *
+    * Document ordering is determined by the "upsert ID", which must be indexed field.
```
pshapiro4broad: typo: must be _an_ indexed field

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -51,10 +55,59 @@ class PersistenceService private (persistenceDAO: PersistenceDAO,
...
+    * truth fails to be applied to the search index.
+    */
+  def recoverMetadata[D <: ClioDocument: Indexable: HitReader: Decoder](
+    index: ElasticsearchIndex[D]
+  )(implicit ec: ExecutionContext): Future[Long] = {
+    searchDAO.getMostRecentDocument(index).flatMap { maybeLastDocument =>
+      maybeLastDocument.fold(countZero) { lastDocument =>
+        for {
+          documents <- persistenceDAO.getAllSince(lastDocument.upsertId, index)
+          count <- updateAndCount(documents, index)
```
pshapiro4broad: Is `count` different than `documents.size`?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -51,10 +55,59 @@ class PersistenceService private (persistenceDAO: PersistenceDAO,
...
+        } yield {
+          count
+        }
+      }
+    }
+  }
+
+  /**
+    * Apply a sequence of partial documents to an Elasticsearch index,
+    * returning the total number of documents applied.
```
pshapiro4broad: If a document can't be applied, do we want to continue? My thinking is that clio should fail to start up in that case, since we don't want to start up with an incomplete recovery state

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -21,11 +23,13 @@ import java.util.UUID
   * Updates are written to storage to serve as the system of record for Clio,
   * allowing for update playback during disaster recovery.
   */
-class PersistenceService private (persistenceDAO: PersistenceDAO,
-                                  searchDAO: SearchDAO) {
+class PersistenceService private (
+  persistenceDAO: PersistenceDAO,
+  searchDAO: SearchDAO
```
pshapiro4broad: It seems like `SearchDAO` is not the right name for this type, since it supports operations other than search. Not sure what a better name would be, something like `MetadataStorageDAO` seems kind of wordy.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -51,10 +55,59 @@ class PersistenceService private (persistenceDAO: PersistenceDAO,
...
+    *
+    * Updates run serially, and short-circuit if applying any one
+    * update fails.
+    */
+  private def updateAndCount[D <: ClioDocument: Indexable](
+    documents: Seq[D],
+    index: ElasticsearchIndex[D]
+  )(implicit ec: ExecutionContext): Future[Long] = {
+    documents.foldLeft(countZero) {
+      case (countFut, document) => {
```
pshapiro4broad: Where does the convention to abbreviate `future` as `fut` come from? I find that abbreviations in general make the code harder to read, and in this case it's only saving us three characters.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -51,10 +55,59 @@ class PersistenceService private (persistenceDAO: PersistenceDAO,
...
+          count + 1
+        }
+      }
+    }
+  }
 }
 
 object PersistenceService {
-  def apply(app: ClioApp): PersistenceService = {
+  def apply(app: ClioApp)(implicit mat: Materializer): PersistenceService = {
```
pshapiro4broad: Is it obvious to most readers why this implicit is here? It isn't obvious to me.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/SearchService.scala
```diff
@@ -7,6 +7,7 @@ import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
   ElasticsearchIndex,
   ElasticsearchQueryMapper
 }
+
```
pshapiro4broad, line 10: What rules are we following for organizing our imports? It would be nice if we could use the default intellij settings for imports. In this case, I'm not sure why an empty line would be added here.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -18,22 +17,30 @@ class ServerService private (
...
-    }
-    ()
+      _ <- searchDAO.initialize()
+    } yield ()
+  }
+
+  /**
+    * Bind a port, and mark the server as ready to receive messages.
+    */
+  def completeStartup(): Future[Unit] = {
```
pshapiro4broad: Why is it preferable to split the startup into two different methods? In general we'd want to avoid having order dependencies on methods whenever possible.

clio-server/src/test/scala/org/broadinstitute/clio/server/TestKitSuite.scala
```diff
@@ -11,9 +12,12 @@ import org.scalatest.{BeforeAndAfterAll, Suite}
   */
 abstract class TestKitSuite(actorSystemName: String)
     extends TestKit(ActorSystem(actorSystemName))
-    with Suite
+    with AsyncFlatSpecLike
+    with Matchers
     with BeforeAndAfterAll {
 
+  protected implicit val materializer: Materializer = ActorMaterializer()
```
pshapiro4broad: Please add a comment describing why this is necessary.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
```diff
@@ -62,31 +62,16 @@ class HttpElasticsearchDAO private[dataaccess] (
...
     val searchDefinition = search(index.indexName / index.indexType)
       .size(1)
       .sortByFieldDesc(ClioDocument.UpsertIdElasticSearchName)
     for {
       searchResponse <- httpClient execute searchDefinition
     } yield {
-      searchResponse.to[D] match {
-        case Seq(document) => document
-        case _ =>
-          throw new RuntimeException(
```
pshapiro4broad, line 86: Does this still result in an exception being thrown if the `execute` command fails? E.g. can the caller distinguish an error case from the case where no documents are present?

pshapiro4broad (approved):
A few comments but looks good overall.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioMain.scala
danxmoran: This is a leftover from when I thought `ClioMain` might log. I'll remove!

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
danxmoran: If `execute` fails, this whole `yield` block won't run and the caller will get back a failed `Future` containing the exception.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
danxmoran: We'll still be returning the first thing in the results, which should be the latest ID from the sort, but we do lose the ability to notice when multiple results are coming back. But, if we trust elasticsearch to properly sort the results by ID, why don't we trust it to properly only return 1 element?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
danxmoran: The old way would result in the document corresponding to `withId` being returned as the first element of the `Seq`. Not a problem in practice, but it made testing slightly more annoying.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
danxmoran: It shouldn't be

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
danxmoran: It does fail, see the 2nd part of the comment below. I'll reword to be more clear.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
pshapiro4broad: Just paranoia, I think we can trust ES to work properly.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
pshapiro4broad: Seems like using `documents.size` would be less code here.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
danxmoran: It's implicit because the `PersistenceService` constructor needs it to be implicit, because all of the Akka stream-y things need it to be implicit. Usually when I see an implicit on an `apply` method with no apparent use I assume it's dependency injection.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
pshapiro4broad: Yeah, it wasn't clear to me if it stops early and returns count, or stops early and returns a failure.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
pshapiro4broad: Yeah, my problem is that if I don't know what needs it, I won't know when I can remove it. But I guess you get an error if it's needed, so you'll know when you need to add it? I guess it would be nice to get a warning in the case where a declared implicit is never used.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/SearchService.scala
danxmoran: The habitual setup I got from my last job is:
```
org.broadinstitute.*
--blank--
*other*
--blank--
scala.*
--blank--
java.*
```
But I'll revert to defaults to stop the "optimize imports" wars.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
danxmoran: On startup, we want to:

1. Marks server as starting up
2. Initialize all DAOs
3. Recover metadata
4. Bind a port
5. Mark the server as started

1, 2, 4, and 5 all live in the `ServerService`, but 3 (right now) is in the `PersistenceService`. I can un-split the startup logic by either:
1. Moving the "recover" method into the `ServerService`, or
2. Passing a `PersistenceService` as an argument to the `ServerService`

I'd lean towards option 1, WDYT?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
danxmoran: The compiler will yell on an unused implicit parameter just like it does for unused explicit parameters.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
pshapiro4broad: 1 sounds better to me

danxmoran:
TODO: A lot of code cleanup, but this now has a working integration test!

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
```diff
@@ -71,7 +76,31 @@ object ClioServer
   override val gvcfService =
     new GvcfService(persistenceService, searchService)
 
-  def beginStartup(): Unit = serverService.beginStartup()
+  def run(): Unit = {
+    val startupAttempt = for {
+      _ <- serverService.beginStartup()
+      recoveredUbamCount <- persistenceService.recoverMetadata(
```
danxmoran: Note to self: this for-comprehension is causing the gvcf recovery to wait for the ubam recovery, but they could run in parallel.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentGvcf.scala
```diff
@@ -11,13 +11,13 @@ case class DocumentGvcf(upsertId: UUID,
...
-                        analysisDate: Option[OffsetDateTime],
-                        contamination: Option[Float],
-                        documentStatus: Option[DocumentStatus],
-                        gvcfMd5: Option[String],
-                        gvcfPath: Option[String],
-                        gvcfSize: Option[Long],
-                        gvcfMetricsPath: Option[String],
-                        pipelineVersion: Option[Long],
-                        notes: Option[String])
+                        analysisDate: Option[OffsetDateTime] = None,
```
danxmoran, line 14: To make testing easier.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
```diff
@@ -86,7 +85,11 @@ class DockerIntegrationSpec
     )
     val logStream = clioLogLines
       .map { line =>
-        if (!line.contains("DEBUG")) println(line)
```
danxmoran, line 89: Printing not-debug picks up a ton of spam from elastic4s that gets logged out during recovery.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/RecoveryTests.scala
```diff
@@ -0,0 +1,156 @@
...
+import org.broadinstitute.clio.util.model.{DocumentStatus, Location}
+import org.scalatest.{Args, Status}
+
+import scala.util.Random
+
+/** Tests of Clio's recovery mechanisms. */
+trait RecoveryTests extends ForAllTestContainer {
+  self: BaseIntegrationSpec =>
+
+  val documentCount = 10
```
danxmoran: Would be nice to use > 10 documents. Using 50 causes the test startup to time out.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
tbl3rd: It's in the same class. I think it was just moved further down.

tbl3rd: Ah. It's in the companion.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -85,65 +85,69 @@ trait PersistenceDAO extends LazyLogging {
...
-    * @param upsertId of the last known upserted document
+    * @param upsertId of the last known upserted document, `None` if
+    *                 all documents should be restored
     * @param index  to query against.
     * @param ec is the implicit ExecutionContext
     * @tparam D is a ClioDocument type with a JSON Decoder
     * @return a Future Seq of ClioDocument
     */
-  def getAllSince[D <: ClioDocument: Decoder](upsertId: UUID,
+  def getAllSince[D <: ClioDocument: Decoder](upsertId: Option[UUID],
```
tbl3rd, line 125: Nice repurpose.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
tbl3rd: I think the name is meant to convey "some search access" as in "not necessarily Elasticsearch, but maybe something else". That's not to say there isn't a better name ...

clio-server/src/test/scala/org/broadinstitute/clio/server/TestKitSuite.scala
tbl3rd: https://www.superloopy.io/articles/2013/scalatest-with-akka.html

Looks like collecting traits common to our use of TestKit and Scalatest in one place.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -1,29 +1,22 @@
...
-import akka.testkit.TestKit
+import java.nio.file.Files
+import java.time.OffsetDateTime
 
-class PersistenceDAOSpec
-    extends TestKit(ActorSystem("PeristenceDAOSpec"))
-    with AsyncFlatSpecLike
-    with BeforeAndAfterAll
-    with Matchers {
+class PersistenceDAOSpec extends TestKitSuite("PersistenceDAOSpec") {
```
tbl3rd, line 17: Like here, for example.

tbl3rd (approved):

clio-server/src/test/scala/org/broadinstitute/clio/server/TestKitSuite.scala
pshapiro4broad: That linked code doesn't use `ActorMaterializer()`. Maybe if I understood more about how the testing code worked this would be obvious to me.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
pshapiro4broad: Can you use a custom logback.xml file to fix this instead? That would be cleaner than adding a filter to the logging output.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/RecoveryTests.scala
pshapiro4broad: Maybe the integration test could use a larger set.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
pshapiro4broad: If we were using it as a search API, that would make sense. But we're using it as a metadata storage/retrieval API.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/DockerIntegrationSpec.scala
danxmoran: We've already got a custom logback conf causing the clio in Docker to log to file at debug level, which I think is what we want so when tests fail we have the info to potentially track down the problem. This block of code is tailing the log file looking for the "Server started" message (so it knows when to start the tests), printing intermittent lines so it doesn't look like the test has hung. After that, we don't print any server logging to console to try to keep the test output as de-cluttered as possible.

Alt: we could modify the custom logback conf to also have a console appender at info level, and configure testcontainers to auto-tail and print container console logs. I'd rather not do this because:
1. I think we'd then get spammed by all the elasticsearch container logs
2. We write audit logs at info level, so we'd definitely get spammed by all of those logs

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/RecoveryTests.scala
danxmoran: This is the integration test. I think I can get a larger number working when I recover ubams and gvcfs in parallel.

PR #123, opened 2017-09-26
DSDEGP-1579: refactor common code

pshapiro4broad:
### Description

DSDEGP-1579: clio code refactoring

This is an intermediate PR to cover changes made so far, there will be more changes made before this is merged to master.

- replace classes GvcfUtil, WgsUbamUtil with ClassUtil
- add transfer model objects for Index and Key
- change code to use HttpCredentials instead of OAuth2BearerToken
- rename addXXX() client APIs to upsertXXX()

* Review cycle:


clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferIndex.scala
```diff
@@ -0,0 +1,29 @@
...
+    * Encode the index as a url segment for making web service API calls
+    *
+    * @return the url path representation of this index
+    */
+  def urlSegment: String
+
+  def jsonSchema: Json
+}
+
+final case class GvcfIndex() extends TransferIndex {
```
danxmoran: Since these don't take parameters, they can be `case object`s

pshapiro4broad: I knew there was something I was missing here. I was also thinking of using enums, but wasn't sure where this class was going yet.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -59,101 +56,51 @@ class ClioWebClient(
...
-  def getSchemaWgsUbam(
-    implicit bearerToken: OAuth2BearerToken
-  ): Future[HttpResponse] = {
+  def getSchema(
+    transferIndex: TransferIndex
+  )(implicit credentials: HttpCredentials): Future[HttpResponse] = {
     dispatchRequest(
-      HttpRequest(uri = "/api/v1/wgsubam/schema")
-        .addHeader(Authorization(credentials = bearerToken))
+      HttpRequest(uri = s"/api/v1/${transferIndex.urlSegment}/schema")
```
jacarey, line 63: Not in scope for this review but at some point we will want to pass in version.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferWgsUbamV1Key.scala
```diff
@@ -6,3 +6,8 @@ case class TransferWgsUbamV1Key(flowcellBarcode: String,
                                 lane: Int,
                                 libraryName: String,
                                 location: Location)
+    extends TransferKey {
+
+  def getUrlPath: String =
+    s"$flowcellBarcode/$lane/$libraryName/$location"
```
jacarey: I almost feel like the location should be the first path segment similar to gvcfs. That way we are consistent with increasing specificity as you move down the segments. location -> flowcell -> lane -> library

jacarey (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
pshapiro4broad: Yeah I'm not 100% sure how versioning will work, we may need different types for input/query/output if we change the schema in an incompatible way (and want to maintain support for v1 APIs)

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferWgsUbamV1Key.scala
pshapiro4broad: That would be OK with me, doing it now would be better than later. What do you think, @tbl3rd and @danxmoran ?

danxmoran: +1

tbl3rd: Generally: I think we should review this PR for maintainability regressions (and easy improvements) only and approve it otherwise.

Specifically: I see $location as different in kind from $barcode, $lane, and $library. I like moving it first because it is going to be shared across all of our indexes. And another issue just occurred to me: Some indexes will have multiple paths, but only one $location, so ...

tbl3rd (approved):
So much red.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -89,6 +107,12 @@ object ClioCommand extends ClioParsers {
...
   val getServerVersionName = "get-server-version"
 
+  val getSchemaPrefix = "get-schema-"
+  val addPrefix = "add-"
+  val queryPrefix = "query-"
+  val movePrefix = "move-"
+  val deletePrefix = "delete-"
+
   // Names for WGS uBAM commands.
   val getWgsUbamSchemaName = "get-schema-wgs-ubam"
```
danxmoran: This could / should be `s"${getSchemaPrefix}-${WgsUbamIndex.commandName}"`. Similarly for the others.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -21,6 +21,16 @@ import shapeless.CNil
   */
 sealed trait ClioCommand
 
+sealed trait GetSchemaCommand extends ClioCommand {
+  def index: TransferIndex
+}
+
+sealed trait AddCommand extends ClioCommand {
+  def index: TransferIndex
+  def key: TransferKey
```
danxmoran, line 25: If we really want to lock this down, we could add another `type` param to `TransferIndex` for key type, and then this could be `def key: index.KeyType`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -0,0 +1,50 @@
...
+        err
+      )
+    }
+
+    val decodedOrError = parsedOrError
+      .flatMap(_.as[index.metadataType])
+      .left
+      .map { err =>
+        new RuntimeException(
+          s"Invalid metadata given at $location. Run the '${ClioCommand.getSchemaPrefix}${index.commandName}' command to see the expected JSON format for ${index.name}s.",
```
danxmoran: Should there be a `-` between `${ClioCommand.getSchemaPrefix}` and `${index.commandName}`?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/ClassUtil.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.clio.client.util
+
+object ClassUtil {
```
danxmoran: Extreme nitpick: this is generic enough that it might belong in the clio-util project.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -59,101 +56,51 @@ class ClioWebClient(
...
+        .addHeader(Authorization(credentials = credentials))
     )
   }
 
-  def addWgsUbam(
-    input: TransferWgsUbamV1Key,
-    transferWgsUbamV1Metadata: TransferWgsUbamV1Metadata
-  )(implicit bearerToken: OAuth2BearerToken): Future[HttpResponse] = {
+  def upsert[T](transferIndex: TransferIndex, key: TransferKey, metadata: T)(
```
danxmoran, line 68: Instead of `T`, can we use `transferIndex.metadataType`?

danxmoran: Actually, I think this won't work in current Scala unless we break the params into two lists because of how scoping for arguments works:
```scala
def upsert(transferIndex: TransferIndex)(key: transferIndex.KeyType, metadata: transferIndex.MetadataType)
```
Up to you if you want to try locking it down.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferIndex.scala
```diff
@@ -0,0 +1,65 @@
...
+    */
+  def urlSegment: String
+
+  def jsonSchema: Json
+
+  def name: String
+
+  def commandName: String
+
+  type metadataType
```
danxmoran: Nitpick: even for abstract type members, I think the convention is to use upper camel case.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferIndex.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def decoder: Decoder[metadataType]
+  def encoder: Encoder[metadataType]
+}
+
+case object GvcfIndex extends TransferIndex {
+
+  override def urlSegment: String = "gvcf"
```
danxmoran: Nitpick: I think all these `override def`s should work as `override val`s

danxmoran (approved):
Some parting suggestions, but overall looks great! Much better than it was.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
pshapiro4broad: No, the prefixes end in `-`

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferIndex.scala
pshapiro4broad: OK

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferIndex.scala
pshapiro4broad: I'll give it a try

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -31,19 +47,24 @@ case object GetServerVersion extends ClioCommand
 
 // WGS-uBAM commands.
 
-@CommandName(ClioCommand.getWgsUbamSchemaName)
-case object GetSchemaWgsUbam extends ClioCommand
+@CommandName(ClioCommand.getSchemaPrefix + WgsUbamIndex.commandName)
```
danxmoran: Why not still `ClioCommand.getWgsUbamSchemaName`? Similarly for the rest.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -21,6 +21,22 @@ import shapeless.CNil
...
+
+sealed trait AddCommand extends ClioCommand {
+  def index: TransferIndex
+  def key: TransferKey
+  def metadataLocation: String
+}
+
+sealed trait QueryCommand extends ClioCommand {
+  def index: TransferIndex
+  def queryInput: Any
```
danxmoran: Could you add a `FIXME` or `TODO` comment here?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
pshapiro4broad: I'll change this back, I was trying out getting rid of the `ClioCommand` members for each command but in the end it didn't seem worthwhile (yet). One can imagine a future version of the code where commands and index names are passed separately as command line arguments, and the command line parser is initialized programmatically.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
pshapiro4broad: Yes

PR #127, opened 2017-09-26
DSDEGP-1631 Add researchProjectId to wgs-ubam query inputs.

danxmoran:
### Description

For whenever we re-enable Clio usage in Zamboni. I also threw in an unrelated change in the base integration spec that should fix path-related errors causing our nightly integration tests to fail.

----

* Review cycle:

### DIFF
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
@@ -249,7 +249,8 @@ abstract class BaseIntegrationSpec(clioDescription: String)
   ): Document = {
     val expectedPath =
       rootPersistenceDir.resolve(
-        s"${index.currentPersistenceDir}/$upsertId.json"
+        // `currentPersistenceDir` ends with '/' so GCS will recognize it as a directory.
+        s"${index.currentPersistenceDir}$upsertId.json"
       )
 
     Files.exists(expectedPath) should be(true)
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
@@ -248,10 +248,9 @@ abstract class BaseIntegrationSpec(clioDescription: String)
     upsertId: UUID
   ): Document = {
     val expectedPath =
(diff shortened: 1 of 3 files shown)


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -249,7 +249,8 @@ abstract class BaseIntegrationSpec(clioDescription: String)
   ): Document = {
     val expectedPath =
       rootPersistenceDir.resolve(
-        s"${index.currentPersistenceDir}/$upsertId.json"
+        // `currentPersistenceDir` ends with '/' so GCS will recognize it as a directory.
+        s"${index.currentPersistenceDir}$upsertId.json"
```
pshapiro4broad: Since this is a `Path` we should avoid building paths using string operations whenever possible. Instead use `resolve()` and `Path` should remove/add `/` as necessary. e.g.
```
val expectedPath = rootPersistenceDir.resolve(index.currentPersistenceDir, s"$upsertId.json")
```

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
```diff
@@ -230,6 +232,13 @@ trait WgsUbamTests { self: BaseIntegrationSpec =>
       )
       sampleResults <- Unmarshal(sampleResponse)
         .to[Seq[TransferWgsUbamV1QueryOutput]]
+      rpIdResponse <- runClient(
+        ClioCommand.queryWgsUbamName,
+        "--research-project-id",
```
pshapiro4broad, line 237: Is there a way to avoid using a magic string? We can import caseapp and the class that implements this option, can that be used here?

pshapiro4broad (approved):

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsUbamTests.scala
danxmoran: I don't think there's a way to avoid it right now (if we can, it'll be from writing custom shapeless code to say "give me the string for the field in this class with the symbol researchProjectId"). case-app doesn't provide any sort of reverse-parsing capabilities to turn objects into equivalent cmd-line args.

tbl3rd (approved):

PR #138, opened 2017-10-03
DSDEGP-1646 move-gvcf moves the index and metrics too.

danxmoran:
### Description

Also generalize move execution across indexes. Deletes will be coming in a follow-up PR.

----

* Review cycle:


clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveGvcfSpec.scala
```diff
@@ -28,31 +28,18 @@ class MoveGvcfSpec extends BaseClientSpec {
...
         MoveGvcf(
-          transferGvcfV1Key = testGvcfTransferV1Key,
+          key = testGvcfTransferV1Key,
           destination = testGvcfCloudSourcePath
         )
       succeedingDispatcher.dispatch(command)
     }
   }
 
-  it should "throw an exception if the source and destination paths are the same" in {
```
danxmoran: I didn't see much point in this test, but can re-add it if we want it.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -1,10 +1,17 @@
...
 import caseapp.core.help.CommandsHelp
 import caseapp.core.commandparser.CommandParser
+import org.broadinstitute.clio.transfer.model.gvcf.{
+  TransferGvcfV1Key,
+  TransferGvcfV1QueryInput
+}
+import org.broadinstitute.clio.transfer.model.wgsubam.{
+  TransferWgsUbamV1Key,
+  TransferWgsUbamV1QueryInput
+}
```
danxmoran, line 14: Here and elsewhere: the `model` package was getting pretty packed, so I divided it up. The fallout is scattered throughout the PR.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorWgsUbam.scala
```diff
@@ -28,7 +31,7 @@ class DeleteExecutorWgsUbam(deleteWgsUbam: DeleteWgsUbam) extends Executor {
       for {
         queryResponses <- webClient
           .query(
-            index,
+            WgsUbamIndex,
```
danxmoran, line 34: Making these un-generic to mirror the gvcf delete executor, in prep for making a generic replacement for both in the next step.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
```diff
@@ -0,0 +1,212 @@
...
+      upsertResponse <- moveFiles(webClient, ioUtil, existingMetadata)
+    } yield {
+      upsertResponse
+    }
+  }
+
+  private def verifyCloudPaths(ioUtil: IoUtil): Unit = {
+    if (moveCommand.key.location != Location.GCP) {
+      throw new Exception(
+        s"Only GCP unmapped ${name}s are supported at this time."
```
pshapiro4broad: Not sure what `unmapped` means here.

Also this error seem redundant with the error message on line 34 above. Will both errors be printed to the user?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
```diff
@@ -0,0 +1,212 @@
...
+    ec: ExecutionContext
+  ): Future[HttpResponse] = {
+    val sourcesAsString =
+      existingMetadata.pathsToMove.mkString("'", "', '", "'")
+
+    val googleCopies = existingMetadata.pathsToMove.map { p =>
+      Future(copyGoogleObject(p, destination, ioUtil))
+    }
+
+    // VERY IMPORTANT that this is lazy; we shouldn't try to delete
```
pshapiro4broad: replace `shouldn't try to` with `cannot`

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveGvcfSpec.scala
pshapiro4broad: It's worth keeping if it's testing code we wrote to check for this case. If the exception is being generated by some underlying behavior of APIs we didn't write then I would delete it.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveGvcfSpec.scala
```diff
@@ -97,18 +84,45 @@ class MoveGvcfSpec extends BaseClientSpec {
...
-        transferGvcfV1Key = testGvcfTransferV1Key,
+        key = testGvcfTransferV1Key,
         destination = "/this/is/a/local/path"
       )
       succeedingDispatcher.dispatch(command)
     }
   }
 
+  it should "throw an exception if given a non-directory destination for multiple source files" in {
+    recoverToSucceededIf[Exception] {
```
pshapiro4broad, line 96: You could make this a more specific check, to avoid false positives if an exception is thrown for some other reason.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveWgsUbamSpec.scala
```diff
@@ -4,11 +4,10 @@ import org.broadinstitute.clio.client.BaseClientSpec
...
 import org.broadinstitute.clio.client.util.MockIoUtil
 import org.broadinstitute.clio.client.webclient.MockClioWebClient
-import org.broadinstitute.clio.transfer.model.TransferWgsUbamV1Key
 import org.broadinstitute.clio.util.model.Location
-
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
+import org.broadinstitute.clio.transfer.model.wgsubam.TransferWgsUbamV1Key
 
 class MoveWgsUbamSpec extends BaseClientSpec {
```
pshapiro4broad, line 12: Once we have all operations generalized we no longer need to provide a test spec for each different document type, as the code is the same.

Instead we'd want a test for `AddExecutor`, `MoveExecutor`, etc. It would be nice if it could be tested independently of `ClioWebClient` too.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -0,0 +1,39 @@
...
+  /**
+    * Notes used to track changes to metadata (primarily deletions).
+    */
+  val notes: Option[String]
+
+  /**
+    * Paths to all files that should be moved by the clio-client move
+    * command for this metadata's index.
+    */
+  def pathsToMove: Seq[String]
```
pshapiro4broad, line 17: Is there any way to use `Path` for the paths in this class? I think they should be valid `Path` objects. That way you can use `Path` operations instead of string operations.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+    * Rebase all paths in `pathsToMove` onto a destination directory path.
```
pshapiro4broad: Not sure if `rebase` is the best term here. Also it's unclear what this does; I assume it creates a new model but it sounds like it modifies the current object.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -0,0 +1,39 @@
...
+    *
+    * Fails if `destination` doesn't end with '/'.
+    */
+  def rebaseOnto(destination: String): M
+
+  /**
+    * Set the singular path in this metadata to a new destination.
```
pshapiro4broad: This also sounds like it side effects the current object

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -0,0 +1,39 @@
...
+    *
+    * Fails if `pathsToMove` contains more than one element.
+    */
+  def setSinglePath(destination: String): M
+
+  /**
+    * Rebase the path `source` onto the directory path `destination`.
+    */
+  protected def rebaseOnto(source: String, destination: String): String = {
+    destination + source.splitAt(source.lastIndexOf('/'))
```
pshapiro4broad: This is `destination.resolve(source.getFileName())` if you're using `Path`.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/gvcf/TransferGvcfV1Metadata.scala
```diff
@@ -0,0 +1,44 @@
...
+                                  notes: Option[String] = None)
+    extends TransferMetadata[TransferGvcfV1Metadata] {
+
+  override def pathsToMove: Seq[String] =
+    Seq.concat(gvcfPath, gvcfIndexPath, gvcfMetricsPath)
+
+  override def rebaseOnto(destination: String): TransferGvcfV1Metadata = {
+    assert(destination.endsWith("/"))
+
+    this.copy(
```
pshapiro4broad: Why not implement using `pathsToMove`? Then you could do it in the base class.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/gvcf/TransferGvcfV1Metadata.scala
```diff
@@ -0,0 +1,44 @@
...
+                                  gvcfMetricsPath: Option[String] = None,
+                                  pipelineVersion: Option[String] = None,
+                                  notes: Option[String] = None)
+    extends TransferMetadata[TransferGvcfV1Metadata] {
+
+  override def pathsToMove: Seq[String] =
+    Seq.concat(gvcfPath, gvcfIndexPath, gvcfMetricsPath)
+
+  override def rebaseOnto(destination: String): TransferGvcfV1Metadata = {
+    assert(destination.endsWith("/"))
```
pshapiro4broad: IMO I think we should only use `assert` in non-test code. If we want to validate input, it should explicitly throw an exception (or return `Option[]`)

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/gvcf/TransferGvcfV1Metadata.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    this.copy(
+      gvcfPath = gvcfPath.map(rebaseOnto(_, destination)),
+      gvcfIndexPath = gvcfIndexPath.map(rebaseOnto(_, destination)),
+      gvcfMetricsPath = gvcfMetricsPath.map(rebaseOnto(_, destination))
+    )
+  }
+
+  override def setSinglePath(destination: String): TransferGvcfV1Metadata = {
+    assert(pathsToMove.length == 1)
```
pshapiro4broad: I'm unclear how this can ever be true? Is there a subclass that overrides `pathsToMove`?

clio-util/src/main/scala/org/broadinstitute/clio/util/json/ModelAutoDerivation.scala
```diff
@@ -107,3 +107,5 @@ trait ModelAutoDerivation extends AutoDerivation {
     }
   }
 }
+
+object ModelAutoDerivation extends ModelAutoDerivation
```
pshapiro4broad: Why is this necessary? A comment would be helpful.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
```diff
@@ -0,0 +1,212 @@
...
+        )
+    }
+  }
+
+  private def updateAndUpsert(client: ClioWebClient,
+                              existingMetadata: moveCommand.index.MetadataType)(
+    implicit credentials: HttpCredentials,
+    ec: ExecutionContext
+  ): Future[HttpResponse] = {
+    val newMetadata = if (destination.endsWith("/")) {
```
pshapiro4broad: We test `endsWith("/")` in a few places, this should be a function, e.g. `isDirectory()` on some utility class.

pshapiro4broad (approved):
This was a pleasant surprise! A few comments though.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
danxmoran: `unmapped` is a leftover from the pre-generic days. It is duplicated, I'll pick one to remove.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
danxmoran: I think we can, and we might even be able to plug in the google nio filesystem adapter to properly parse the gs:// paths. Will investigate!

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/gvcf/TransferGvcfV1Metadata.scala
danxmoran: `pathsToMove` only gives you the path values, not the names of the parameters those values are coming from. There's no way (that I know of) for the code in the base class to do a `this.copy` and know the field names of the implementer to use in the kwargs.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/gvcf/TransferGvcfV1Metadata.scala
danxmoran: This will be true if we run the delete CLP on a gvcf key (which will delete the gvcf and index, but not metrics), and then run the move CLP on the same key. Not sure if that'll ever happen, but I didn't want to rule it out either since it's not much more code to support.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/gvcf/TransferGvcfV1Metadata.scala
pshapiro4broad: Oh right, I sometimes forget we're not storing our map-like data in an actual map.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
danxmoran: Poked at this a bit, but I think it'll turn out to be a deep rabbit hole. I'll make a tech-debt ticket to move to path-based APIs.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveGvcfSpec.scala
danxmoran: OH I know what this is for, it's to protect us from accidentally deleting things (since we copy then delete the old, if new = old then we'll end up just deleting).

I'll add this back in.

danxmoran: Hrm, or should we just detect when source == destination and do nothing?

tlangs (approved):
Can't wait to see the next step.

PR #139, opened 2017-10-04
DSDEGP-1647 delete-gvcf deletes the index file, but not the metrics.

danxmoran:
### Description

Part 2! This time with more deletions than insertions.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/CommandDispatch.scala
```diff
@@ -19,12 +19,9 @@ class CommandDispatch(val webClient: ClioWebClient, val ioUtil: IoUtil)
...
-      case moveCommand: MoveCommand[_] => new MoveExecutor(moveCommand)
-      case deleteWgsUbam: DeleteWgsUbam =>
-        new DeleteExecutorWgsUbam(deleteWgsUbam)
-      case deleteGvcf: DeleteGvcf =>
-        new DeleteExecutorGvcf(deleteGvcf)
+      case addCommand: AddCommand[_]       => new AddExecutor(addCommand)
+      case moveCommand: MoveCommand[_]     => new MoveExecutor(moveCommand)
+      case deleteCommand: DeleteCommand[_] => new DeleteExecutor(deleteCommand)
       case other =>
         new RetrieveAndPrintExecutor(other)
```
pshapiro4broad, line 26: Could we somehow get the `match .. case` inside `RetrieveAndPrintExecutor` into this one? It seems odd to have two different matches over the same object.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgsubam/TransferWgsUbamV1Metadata.scala
```diff
@@ -60,4 +62,10 @@ case class TransferWgsUbamV1Metadata(
 
     this.copy(ubamPath = Some(destination))
   }
+
+  override def markDeleted(deletionNote: String): TransferWgsUbamV1Metadata =
+    this.copy(
```
pshapiro4broad, line 67: Seems like we should be able to refactor this into a base class, if these fields are part of `TransferMetadata`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/Executor.scala
```diff
@@ -15,8 +15,19 @@ trait Executor
...
   def execute(webClient: ClioWebClient, ioUtil: IoUtil)(
     implicit ec: ExecutionContext,
     credentials: HttpCredentials
   ): Future[HttpResponse]
+
+  protected def deleteGoogleObject(path: String, ioUtil: IoUtil): Future[Unit] = {
+    ioUtil.deleteGoogleObject(path) match {
+      case 0 => Future.successful(())
+      case _ =>
+        Future.failed(
```
pshapiro4broad: It seems like we should bake error handling into `IoUtil` itself instead of making the caller do it. We should also trap the command output and put it in the exception (in this case, since it's run from clio-client, it will all get dumped to stdout, but that seems like a hack to me). Or maybe this can wait until we use the google filesystem library?

pshapiro4broad (approved):
Looks good, just a few comments

danxmoran:
Note to self: should merge after #140 to get the fixes inside.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/CommandDispatch.scala
danxmoran: We could, but I think the end result would be more convoluted than what we've got now.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgsubam/TransferWgsUbamV1Metadata.scala
danxmoran: It has to be in the implementing classes because `this.copy` is only defined there.

pshapiro4broad: Of course, right. I still think we _should_ be able to do it, even if we can't.

gbggrant (approved):
👍  LGTM

PR #140, opened 2017-10-04
DSDEGP-1651 Fix move CLP to properly err out on missing source files.

danxmoran:
### Description

D'oh.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
@@ -195,25 +195,17 @@ class MoveExecutor[TI <: TransferIndex](moveCommand: MoveCommand[TI])
 
   private def copyGoogleObject(source: String,
                                destination: String,
-                               ioUtil: IoUtil): Future[Unit] = {
-    ioUtil.copyGoogleObject(source, destination) match {
-      case 0 => Future.successful(())
-      case _ =>
-        Future.failed(
-          new Exception(
-            s"Copy files in the cloud failed from '$source' to '$destination'"
-          )
-        )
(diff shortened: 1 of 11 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
```diff
@@ -195,25 +195,17 @@ class MoveExecutor[TI <: TransferIndex](moveCommand: MoveCommand[TI])
 
   private def copyGoogleObject(source: String,
                                destination: String,
-                               ioUtil: IoUtil): Future[Unit] = {
```
danxmoran, line 198: This was the culprit (and the same bug was in `deleteGoogleObject` below. These methods were returning `Future`s, but the caller didn't know that and was wrapping the call in another `Future`. The outer `Future` always succeeded because all it was doing was starting the inner `Future`.

These methods are synchronous so I chose to have them just return `Unit` instead of removing the `Future`-wrapping at the call sites.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgsubam/TransferWgsUbamV1QueryInput.scala
```diff
@@ -17,14 +17,3 @@ case class TransferWgsUbamV1QueryInput(
   sampleAlias: Option[String] = None,
   documentStatus: Option[DocumentStatus] = None
 )
-
-object TransferWgsUbamV1QueryInput {
```
danxmoran, line 21: Opportunistically deleting this since it's unused.

pshapiro4broad (approved):

tlangs (approved):

PR #147, opened 2017-10-16
DSDEGP-1672 Add cram field for "analysis_files.txt" path.

danxmoran:
Also simplify boilerplate needed to use Clio move CLPs.

### Description

1. Adds a missing field caught by @tbl3rd.
2. Reduces the boilerplate needed for metadata moves.

----

* Review cycle:


danxmoran:
After talking with @ktibbett and @tlangs, I'm also changing the `cromwellIdPath` field to be `cromwellId`.

danxmoran:
Note to self: `pipelineVersion` needs to become a `String`, not a `Long`.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -32,19 +32,24 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
...
-
-  /**
-    * Returns copy of this object in which a singular file-path has
-    * been sent to a new destination.
-    *
-    * Fails if `pathsToMove` contains more than one element.
-    */
-  def setSinglePath(destination: String): M
+  def moveInto(destination: String): M = {
+    val pathMapper = if (destination.endsWith("/")) {
```
pshapiro4broad: formatting looks odd to me, can `if` go on the next line?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -32,19 +32,24 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
...
+      moveIntoDirectory(_: String, destination)
+    } else {
+      if (pathsToMove.length > 1) {
```
pshapiro4broad: why not `} else if (...) {`?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -32,19 +32,24 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
...
+        sys.error(
+          "Non-directory destination given for metadata with > 1 path to move"
+        )
+      }
+      (_: String) =>
+        destination
+    }
+    mapMove(pathMapper)
```
pshapiro4broad: This is better, but it looks like every use of `pathMapper` is to call `map` on an object. It seems like you should be able to abstract that as well, so callers don't need to call `map` explicitly.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/TransferWgsCramV1Metadata.scala
```diff
@@ -91,105 +92,44 @@ case class TransferWgsCramV1Metadata(
...
+      recalDataPath = recalDataPath.map(pathMapper),
+      baitBiasSummaryMetricsPath = baitBiasSummaryMetricsPath.map(pathMapper),
+      baitBiasDetailMetricsPath = baitBiasDetailMetricsPath.map(pathMapper),
+      wgsMetricsPath = wgsMetricsPath.map(pathMapper),
       readgroupLevelMetricsFiles =
-        readgroupLevelMetricsFiles.map(_.map(moveInto(_, destination)))
-    )
-  }
-
-  override def setSinglePath(destination: String): TransferWgsCramV1Metadata = {
```
pshapiro4broad, line 148: Nice to see this code deleted!

pshapiro4broad (approved):
Looks good, maybe a few more improvements possible.

danxmoran:
I've also made `pipelineVersion` a string.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -32,19 +32,23 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
...
-    * Fails if `pathsToMove` contains more than one element.
-    */
-  def setSinglePath(destination: String): M
+  def moveInto(destination: String): M = {
+    val pathMapper =
+      if (destination.endsWith("/")) { (opt: Option[String]) =>
+        opt.map(path => moveIntoDirectory(path, destination))
+      } else if (pathsToMove.length > 1) {
+        sys.error(
+          "Non-directory destination given for metadata with > 1 path to move"
```
tlangs: Is there a logger that we should be writing to here?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -32,19 +32,23 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
...
-    *
-    * Fails if `pathsToMove` contains more than one element.
-    */
-  def setSinglePath(destination: String): M
+  def moveInto(destination: String): M = {
+    val pathMapper =
+      if (destination.endsWith("/")) { (opt: Option[String]) =>
+        opt.map(path => moveIntoDirectory(path, destination))
+      } else if (pathsToMove.length > 1) {
+        sys.error(
```
tlangs, line 44: Is there a logger that we should be writing to here?

danxmoran: Nope, this will raise a `RuntimeException` which will get caught by the client logic, rolled up and logged at the top level.

pshapiro4broad: I think in general logging should be done at the (root) catch point, not the throw point, because only the caller knows what needs to be done with an exception.

tlangs (approved):

PR #154, opened 2017-10-27
DSDEGP-1640 Set up v2 Elasticsearch index.

danxmoran:
### Description

I'll add comments inline for things that might be unexpected.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -33,7 +35,7 @@ sealed abstract class GetSchemaCommand[TI <: TransferIndex](val index: TI)
 sealed abstract class AddCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
-  def metadataLocation: String
+  def metadataLocation: URI
```
danxmoran, line 38: I chose `URI` instead of `Path` because:
1. Without the google-cloud-nio library, `Paths.get` of a string with a "gs" scheme throws a "no filesystem provider" exception
2. With the google-cloud-nio library, `Paths.get` of a string with a "gs" scheme eagerly tries to connect to the google bucket in the path, throwing an exception. On the client-side we don't want the connection to happen until after we've generated / parsed an access token, and on the server-side we never want the connection to happen.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchIndex.scala
```diff
@@ -14,59 +15,106 @@ import scala.reflect.runtime.universe.Type
...
+    fieldType: Type
+  ): (String => FieldDefinition) = {
+    import scala.reflect.runtime.universe.typeOf
+
+    /*
+     * For Strings, create a top-level text field for searches,
+     * with a nested keyword field for sorting / aggregations.
+     */
+    val textFieldWithKeyword =
+      textField(_: String).fields(keywordField("exact"))
```
danxmoran, line 103: This was mentioned as common practice in the Elasticsearch docs and seemed like a useful idea, but I can remove it if we don't want it.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchIndex.scala
```diff
@@ -14,59 +15,106 @@ import scala.reflect.runtime.universe.Type
...
-      case tpe if tpe =:= typeOf[Option[String]]    => keywordField _
-      case tpe if tpe <:< typeOf[Option[EnumEntry]] => keywordField _
+    fieldType match {
+      case tpe if tpe =:= typeOf[Boolean]           => booleanField
+      case tpe if tpe =:= typeOf[Int]               => intField
+      case tpe if tpe =:= typeOf[Long]              => longField
+      case tpe if tpe =:= typeOf[Float]             => floatField
+      case tpe if tpe =:= typeOf[Double]            => doubleField
+      case tpe if tpe =:= typeOf[String]            => keywordField
+      case tpe if tpe =:= typeOf[Symbol]            => keywordField
```
danxmoran, line 63: I chose `Symbol` to represent arbitrary things that should map to keywords because I didn't see the point in making another class that would do nothing but wrap arbitrary `String`s (as opposed to `UpsertId`, which also contains unique generation and parsing logic).

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentWgsCram.scala
```diff
@@ -1,62 +1,64 @@
...
-  workflowJsonPath: Option[String] = None,
-  optionsJsonPath: Option[String] = None,
-  wdlPath: Option[String] = None,
-  readgroupMd5: Option[String] = None,
+  cramPath: Option[URI] = None,
+  craiPath: Option[URI] = None,
+  cramMd5Path: Option[URI] = None,
+  logPath: Option[URI] = None,
+  fingerprintPath: Option[URI] = None,
+  cromwellId: Option[UUID] = None,
```
danxmoran, line 27: I chose `UUID` for representing cromwell IDs because that's what the cromwell API client uses: https://github.com/broadinstitute/cromwell/blob/29_hotfix/cromwellApiClient/src/main/scala/cromwell/api/model/WorkflowId.scala#L14

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentWgsUbam.scala
```diff
@@ -38,8 +40,8 @@ case class DocumentWgsUbam(upsertId: UpsertId,
                            sampleType: Option[String] = None,
                            targetIntervals: Option[String] = None,
                            notes: Option[String] = None,
-                           ubamMd5: Option[String] = None,
-                           ubamPath: Option[String] = None,
+                           ubamMd5: Option[Symbol] = None,
+                           ubamPath: Option[URI] = None,
```
danxmoran, line 44: Because of the generalization work we've done, it would have been a royal pain to change the gvcf and cram path types to `URI` but not change the type for ubams. This change should be safe because `URI` also maps to a keyword field, so the Elasticsearch mapping definition for this document won't change.

I also changed the md5 field to `Symbol` to match md5 fields in the other documents. It's compatible with the old index, too.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -108,7 +107,6 @@ object Dependencies {
     "com.beachape" %% "enumeratum" % EnumeratumVersion,
     "com.beachape" %% "enumeratum-circe" % EnumeratumCirceVersion,
     "com.chuusai" %% "shapeless" % ShapelessVersion,
-    "com.fasterxml.uuid" % "java-uuid-generator" % UUIDVersion,
```
danxmoran, line 111: We aren't using this library any more.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
pshapiro4broad: [already replied in person; replying here to document it]

Using `Paths.get()` will work fine so long as the input is `String` not `URI`. If it's `String` it will use the default filesystem provider no matter what the scheme is. I think using `Path` will help us in general so we can avoid code that munges paths directly, and possibly make transition to the cloud-nio library easier.

pshapiro4broad (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -1,36 +1,31 @@
...
+import java.net.URI
+import java.nio.file.{FileVisitOption, Files, Path, Paths}
 import java.util.Comparator
 
-import scala.io.Source
 import scala.sys.process.Process
 
 trait IoUtil {
-  val googleCloudStorageScheme = "gs://"
+  val googleCloudStorageScheme = "gs"
```
tlangs, line 11: why did we get rid of the "://"?

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
danxmoran: I tried this out, and want to stick with `URI`. `Paths.get` on a `String` doesn't throw an exception, but it also doesn't parse correctly:
```bash
scala> Paths.get("gs://the-bucket/the-file.txt")
res0: java.nio.file.Path = gs:/the-bucket/the-file.txt

scala> res0.toUri
res1: java.net.URI = file:///Users/danmoran/clio/gs:/the-bucket/the-file.txt
```

The "gs://" scheme gets munged into "gs:/". Without a hack to revert this on every `.toString` call, this will break our `gsutil` calls and cause us to write bad values into our JSON blobs.

pshapiro4broad: Oh well. OK.

tlangs (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
danxmoran: `uri.getScheme` just returns "gs" (or "file", etc)

tlangs: ah ok

PR #169, opened 2017-11-06
DSDEGP-1677 DSDEGP-1620 Search backwards in time during recovery.

danxmoran:
### Description

Also, be more lazy in reading documents to recover.

----

* Review cycle:

### DIFF
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -2,9 +2,10 @@ package org.broadinstitute.clio.server.dataaccess
 
 import java.nio.file.{Files, Path}
 
+import akka.NotUsed
 import akka.stream.Materializer
 import akka.stream.alpakka.file.scaladsl.Directory
-import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.{Sink, Source}
 import com.sksamuel.elastic4s.Indexable
 import com.typesafe.scalalogging.LazyLogging
 import io.circe.Decoder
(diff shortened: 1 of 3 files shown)


pshapiro4broad:
Would it be possible to test that the code only reads documents from the same day? If not, maybe construct a unit test for the directory filtering code to show that it works as expected.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -168,7 +218,15 @@ object PersistenceDAO {
...
-  val versionFileName = "current-clio-version.txt"
+  val VersionFileName = "current-clio-version.txt"
+
+  /**
+    * Directory depth to naively walk during recovery.
+    *
+    * Should be deep enough to cover date-based partitions of documents in storage,
+    * but not the documents themselves.
+    */
+  val StorageWalkDepth = 3
```
pshapiro4broad: We should have a comment somewhere that describes the path elements in our Persistence storage, which will make this value less magic. If we have it already, link to it from here, or add it here if we don't have it already.

pshapiro4broad (approved):

danxmoran:
> Would it be possible to test that the code only reads documents from the same day? If not, maybe construct a unit test for the directory filtering code to show that it works as expected.

[This test](https://github.com/broadinstitute/clio/blob/dm-smarter-recovery-DSDEGP-1677/clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala#L114) is checking that the backwards search doesn't go back any farther than expected, but it's not worded very well. I'll try to improve it.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
danxmoran: Done

tlangs (approved):

PR #171, opened 2017-11-07
DSDEGP-1767 Use refresh-able credentials in clio-client.

danxmoran:
### Description

I targeted this fix at the main clio-client, instead of just the backfill script, so we don't hit the same problem later in any other code that might use a single long-lived clio-client instance.

----

* Review cycle:


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/AuthRefreshTests.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import scala.concurrent.duration._
+import scala.util.Success
+
+trait AuthRefreshTests { self: BaseIntegrationSpec =>
+
+  behavior of "Clio Client"
+
+  it should "refresh the auth token after it expires" in {
+    /* Query the wgs-ubam schema every 3 seconds for 1.5 hours. */
```
tlangs: if the token expires after 1 hour, why are we testing 50% more time than the expiration time? Couldn't we try querying just once more after the token expires?

danxmoran: I wanted to give it some wiggle room, but yeah this is probably overkill.

pshapiro4broad:
As we discussed in person, I think it would be cleaner if we could separate the token generation/refresh code from the webclient code. I can imagine a credentials provider service object that hides the google auth APIs from the rest of our code. This provider would provide `HttpCredentials` as the API used before.

At the same time, the code could be refactored so that the web client object was constructed with the credentials provider service. This would avoid needing to send the credentials object with every web service API call, as is done now via an implicit argument.

This would also let us fix the awkward testing issue, so we could invalidate the token and force a refresh immediately, instead of using clock time to first the token to invalidate.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -98,57 +101,95 @@ class ClioWebClient(
...
+        /*
+         * Refresh the credential if either:
+         *   1. It has no underlying access token, or
+         *   2. The underlying access token has an expiration time, and it's expired
+         *
+         * DON'T refresh if the underlying access token has no expiration time,
+         * because refresh might not be supported for that token.
+         */
+        val maybeToken = Option(creds.getAccessToken)
+        lazy val maybeTokenExpirationInstant = for {
```
pshapiro4broad: Looks weird to me to have `for` not at the start of a line, but maybe this is typical and I just haven't seen it before?

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -98,57 +101,95 @@ class ClioWebClient(
...
+          expiration <- Option(token.getExpirationTime)
+        } yield {
+          expiration.toInstant
+        }
+        val shouldRefresh =
+          maybeToken.isEmpty || maybeTokenExpirationInstant.exists(
+            _.isAfter(now)
+          )
+
+        if (shouldRefresh) {
```
pshapiro4broad: Even if my other comment about refactoring the code is too hard to do now, I think it's worth moving the token refresh logic to another class instead of inlining it here.

This object would replace the use of `OAuth2Credentials` with an object that returns `HttpCredentials` on demand.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -98,57 +101,95 @@ class ClioWebClient(
...
 
   def getSchema(
     transferIndex: TransferIndex
-  )(implicit credentials: HttpCredentials): Future[Json] = {
+  )(implicit credentials: OAuth2Credentials): Future[Json] = {
     dispatchRequest(
-      HttpRequest(uri = s"/api/v1/${transferIndex.urlSegment}/schema")
-        .addHeader(Authorization(credentials = credentials))
+      HttpRequest(uri = s"/api/v1/${transferIndex.urlSegment}/schema"),
+      Some(credentials)
```
pshapiro4broad: I thought the convention was to use `Option()` when constructing from a variable to avoid `Some(null)`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -149,6 +146,23 @@ abstract class BaseIntegrationSpec(clioDescription: String)
       }, identity)
   }
 
+  /** OAuth2 credential for use in direct calls to the ClioWebClient. */
+  final implicit lazy val oauthCredential: OAuth2Credentials = {
+    AuthUtil
+      .getCredsFromServiceAccount(serviceAccount)
```
pshapiro4broad: Does this work when running tests locally?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -149,6 +146,23 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
+        fail(s"Couldn't get access token for account $serviceAccount", _),
+        identity
+      )
+  }
+
+  /**
+    * Temp file for storing service-account JSON, to pass
+    * into calls to the clio-client.
+    */
+  private lazy val serviceAccountJsonPath: Path =
```
pshapiro4broad: Even though this is type `Path` I would call it `serviceAccountJsonFile` since it's a file.

danxmoran:
@tlangs @pshapiro4broad comments addressed! I tightened the timings on the integration test, but left it to run for the "natural" lifetime of a credential. There's a new unit test that injects shorter expirations with mock credentials.

tlangs (approved):

pshapiro4broad (approved):
Looks good, although I still think there's a way to create a useful test that forces the token to "expire" instantly, vs. waiting for the token to actually expire.

PR #172, opened 2017-11-08
DSDEGP-1782 Stop setting a default regulatory_designation.

danxmoran:
### Description

In Phase 2, we set up a default value for regulatoryDesignation to avoid extra backfill work later. The default is applied on every upsert if there's no "regulatory_designation" key in the JSON blob. Setting a default on every cram / gvcf upsert is dangerous, as it'll overwrite any regulatory designation that was set in a previous upsert.

A concrete example of this problem is data delivery. The last step of the deliver-wgs-cram CLP is to upsert:
```
{ "workspace_name": "Some workspace name" }
```
When this upsert comes into the server, the server will look for the regulatory_designation field in the JSON blob. Since it's missing, the server will change the blob to the following:
```
{ "workspace_name": "Some workspace name", "regulatory_designation": "RESEARCH_ONLY" }
```
And then POST that blob into Elasticsearch, overwriting any "regulatory_designation" field that already exists for the cram. It's a happy accident that this isn't also happening for moves and deletes of crams and gvcfs in general.

Now that we're threading regulatory_designation through the Zamboni workflows, we can leave it to the pipeline to handle any default-setting, and remove the logic from Clio.

----

* Review cycle:

### DIFF
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
@@ -80,8 +80,7 @@ trait GvcfTests { self: BaseIntegrationSpec =>
       sampleAlias = s"someAlias $randomId",
       version = 2,
       documentStatus = Some(DocumentStatus.Normal),
-      gvcfPath = Some(URI.create("gs://path/gvcf.gvcf")),
-      regulatoryDesignation = Some(RegulatoryDesignation.ResearchOnly)
+      gvcfPath = Some(URI.create("gs://path/gvcf.gvcf"))
     )
 
     /*
@@ -122,9 +121,6 @@ trait GvcfTests { self: BaseIntegrationSpec =>
           storedDocument.sampleAlias should be(expected.sampleAlias)
           storedDocument.version should be(expected.version)
           storedDocument.gvcfPath should be(expected.gvcfPath)
-          storedDocument.regulatoryDesignation should be(
-            expected.regulatoryDesignation
-          )
         }
       }
     }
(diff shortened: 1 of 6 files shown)


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsCramTests.scala
```diff
@@ -908,57 +906,117 @@ trait WgsCramTests { self: BaseIntegrationSpec =>
...
+            location = Location.GCP,
+            project = project,
+            sampleAlias = sample,
+            version = version,
+            workspaceName = Some(workspaceName),
+            cramPath = Some(cramDestination.toUri),
+            craiPath = Some(craiDestination.toUri),
+            cramMd5 = Some(Symbol(md5Contents)),
+            documentStatus = Some(DocumentStatus.Normal),
+            regulatoryDesignation = cramRegulatoryDesignation
```
danxmoran, line 1003: Before the fix, this was getting flipped to `Some(RegulatoryDesignation.ResearchOnly)`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -770,47 +766,48 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
   }
 
-  it should "respect user-set regulatory designation for gvcfs" in {
-    val project = s"project$randomId"
-    val sample = s"sample$randomId"
-    val version = 3
-
-    val cloudPath = rootTestStorageDir.resolve(
-      s"gvcf/$project/$sample/v$version/$randomId.gvcf"
+  it should "preserve any existing regulatory designation for gvcfs" in {
```
danxmoran, line 774: We don't have a wrapper for gvcfs that does partial upserts (yet), but this covers the same scenario as the cram delivery process.

pshapiro4broad (approved):

PR #173, opened 2017-11-09
DSDEGP-1717: Support adding a prefix to the base name of the cram, crai and md5 during the deliver operation.

tbl3rd:
Add a --new-sample-prefix option to deliver-wgs-cram CLP to add its
value as a prefix to the base name of the .cram, .cram.md5, and
.cram.crai files.

There is now a do-nothing prefixed() method on the TransferMetadata
type class that is overridden for TransferWgsCramV1Metadata to do the
right thing for its files.

It might be nice to hide the --new-sample-prefix option from the move-
CLPs -- at least for now-- but I don't know how to do that.

Also, 13 integration tests started failing when I rebased onto /develop/.
Got me, but /develop/ fails the same way.  Will look into it tomorrow.

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
@@ -50,7 +50,7 @@ sealed abstract class MoveCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
   def destination: URI
-  def newPrefix: Option[String] = None
+  def newSamplePrefix: Option[String] = None
 }
 
 sealed abstract class DeleteCommand[TI <: TransferIndex](val index: TI)
@@ -131,7 +131,7 @@ final case class QueryWgsCram(@Recurse queryInput: TransferWgsCramV1QueryInput,
 @CommandName(ClioCommand.moveWgsCramName)
 final case class MoveWgsCram(@Recurse key: TransferWgsCramV1Key,
                              destination: URI,
-                             override val newPrefix: Option[String])
+                             override val newSamplePrefix: Option[String])
     extends MoveCommand(WgsCramIndex)
 
 @CommandName(ClioCommand.deleteWgsCramName)
@@ -142,7 +142,7 @@ final case class DeleteWgsCram(@Recurse key: TransferWgsCramV1Key, note: String)
 final case class DeliverWgsCram(@Recurse key: TransferWgsCramV1Key,
                                 workspaceName: String,
                                 workspacePath: URI,
-                                newPrefix: Option[String])
+                                newSamplePrefix: Option[String])
     extends ClioCommand
 
 object ClioCommand extends ClioParsers {
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
@@ -50,7 +50,6 @@ sealed abstract class MoveCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
   def destination: URI
-  def newSamplePrefix: Option[String] = None
 }
 
 sealed abstract class DeleteCommand[TI <: TransferIndex](val index: TI)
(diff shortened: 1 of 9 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -128,9 +129,11 @@ final case class QueryWgsCram(@Recurse queryInput: TransferWgsCramV1QueryInput,
     extends QueryCommand(WgsCramIndex)
 
 @CommandName(ClioCommand.moveWgsCramName)
-final case class MoveWgsCram(@Recurse key: TransferWgsCramV1Key,
-                             destination: URI)
-    extends MoveCommand(WgsCramIndex)
+final case class MoveWgsCram(
+  @Recurse key: TransferWgsCramV1Key,
+  destination: URI,
+  override val newSamplePrefix: Option[String] = None
```
danxmoran: You could throw [`@Hidden`](https://github.com/alexarchambault/case-app/blob/master/annotations/shared/src/main/scala/caseapp/Annotations.scala#L62) on this so it won't show up in the help message.

If we really want to wall it off from command-line users, we could set it as a new constructor parameter on the `MoveExecutor` instead of the `MoveCommand`.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -71,4 +71,12 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
...
       case None           => Some(note)
     }
+
+  /**
+    * Return a copy of this in which the `URI` vals are prefixed with
+    * `newPrefix` as necessary
+    *
+    * @return new metadata with some field values prefixed.
+    */
+  def prefixed(newPrefix: String): M = { val _ = newPrefix; self }
```
danxmoran: Instead of a new method that all the subclasses need to implement, could we add a new `Option[String]` parameter to `moveInto` and `moveIntoDirectory` that injects the new prefix along with changing the leading path? That way we can roll the functionality out to all existing subclasses without needing more boilerplate.

danxmoran (commented):
Which integration tests did you see fail on develop? The nightly run passed.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -50,6 +50,7 @@ sealed abstract class MoveCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
   def destination: URI
+  def newSamplePrefix: Option[String] = None
```
pshapiro4broad: The name `new` is used here but it looks like this is just a prefix. Since it's not replacing an `old` prefix, it doesn't need `new` to disambiguate what it is. All sample prefixes in this commit are `new`. So I would remove `new` from the names in this commit.

tbl3rd: I like it. Brain was stuck on new-base-name from the ticket.

pshapiro4broad (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
tbl3rd: What do you think?  Probably a good idea until we want to expose it.

danxmoran: I'd rather it be part of the executor.

tbl3rd: Not sure how to plumb it to here though.  Ideas?

https://github.com/broadinstitute/clio/blob/develop/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/CommandDispatch.scala#L16

tbl3rd: Ah. Never mind.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/TransferWgsCramV1Metadata.scala
```diff
@@ -69,15 +70,20 @@ case class TransferWgsCramV1Metadata(
 
   // As of DSDEGP-1711, we are only delivering the cram, crai, and md5
   override def mapMove(
+    samplePrefix: Option[String] = None,
     pathMapper: Option[URI] => Option[URI]
   ): TransferWgsCramV1Metadata = {
+    val prefixedCram = cramPath.map { cp =>
+      val name = new File(cp.getPath).getName
+      val directory = URI.create(cp.toString.dropRight(name.length))
+      directory.resolve(s"${samplePrefix.getOrElse("")}$name")
```
danxmoran: Isn't `pathMapper` already effectively doing `directory.resolve(<file name>)`?

tbl3rd: Yes, via `moveIntoDirectory()`, but that assumes `pathsToDelete` are also the ones to prefix.

danxmoran: Hm? `moveIntoDirectory` isn't tied to `pathsToDelete`, it's just doing a `destination.resolve(<filename>)` (so it's duplicate work to also do the resolve here).

pshapiro4broad (approved):

danxmoran (approved):

PR #174, opened 2017-11-09
DSDEGP-1783 Fix IoUtil.getMd5HashOfGoogleObject when no md5 exists.

danxmoran:
### Description

gcloud md5 hashes aren't computed for files uploaded through parallel composite uploads, but our gsutil-hash-parsing logic assumes it always exists. This is breaking ubam backfill now.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -8,8 +8,11 @@ import java.util.Comparator
 import scala.sys.process.Process
 
 trait IoUtil {
+
   val googleCloudStorageScheme = "gs"
```
pshapiro4broad: Since this is a constant shouldn't it be in the object `IoUtil`?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -8,8 +8,11 @@ import java.util.Comparator
...
 
+  protected def gsUtil: IoUtil.GsUtil
```
pshapiro4broad: The way `IoUtil` and `GsUtil` work together it's not clear to me why they're different objects. Especially since IoUtil contains google APIs.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -47,34 +50,36 @@ trait IoUtil {
...
+    gsUtil.exists(path.toString) == 0
   }
 
-  def getMd5HashOfGoogleObject(path: URI): Symbol = {
-    Symbol(new GsUtil(None).md5Hash(path.toString))
+  private val md5HashPattern = "Hash \\(md5\\):\\s+([0-9a-f]+)".r
+
+  def getMd5HashOfGoogleObject(path: URI): Option[Symbol] = {
+    /*
+     * Files uploaded through parallel composite uploads won't have an md5 hash.
```
pshapiro4broad, line 77: What is the expected behavior if google doesn't have an MD5 hash for a file? Do we need to manually generate a hash in that case? Will the database/file MD5 be used instead?

How many files are affected by this? Google docs say any file uploaded in parallel will be affected, it seems like that would apply to all CRAM/BAM etc files.

clio-client/src/test/scala/org/broadinstitute/clio/client/util/MockIoUtil.scala
```diff
@@ -3,11 +3,14 @@ package org.broadinstitute.clio.client.util
 import java.net.URI
 
 import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.clio.client.util.IoUtil.GsUtil
 
 import scala.collection.mutable
 
 class MockIoUtil extends IoUtil with LazyLogging {
 
+  override val gsUtil: GsUtil = new GsUtil(None)
```
pshapiro4broad: It looks like all uses of `GsUtil` construct it using `GsUtil(None)`. If so, we should remove the constructor argument.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -118,9 +134,8 @@ object IoUtil extends IoUtil {
...
     }
 
-    def md5Hash(path: String): String = {
-      val output = runGsUtilAndGetStdout(Seq("hash", "-m", "-h", path))
-      output.lines.filter(_.contains("md5")).next.split(":")(1).trim
+    def hash(path: String): String = {
+      runGsUtilAndGetStdout(Seq("hash", "-m", "-h", path))
     }
 
     def runGsUtilAndGetStdout(gsUtilArgs: Seq[String]): String = {
```
pshapiro4broad: I can't comment on line 151 directly; this code should delete the directory immediately after the gsutil command completes instead of waiting for shutdown to do it.

pshapiro4broad (approved):
👍 once comments are responded to

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
danxmoran: I think it's mainly a byproduct of a past design, but right now the separation is useful for testing. We can inject a mock `GsUtil` into an `IoUtil` to check that the `IoUtil` code does the right thing assuming some stdout from the external `gsutil` call. If we moved the `Process(Seq("gsutil", ...)).!` calls into `IoUtil`, mocking them would be more complicated.

pshapiro4broad: Maybe better names would help. Right now it sounds like they're both util classes, one of which does 'IO' and the other goes 'GS' operations. But in fact `IoUtil` includes google APIs, and `GsUtil` is the internal layer that mediates between public google APIs and the gsutil command line program.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
danxmoran: If google doesn't have an md5, and there's no md5 in the database during backfill, then Clio won't know the md5 for the file. As far as I've seen, this is going to affect all ubams (because we use the parallel uploads in push-to-cloud), but not gvcfs or crams (maybe because they're generated in-cloud?).

We were willing to live without ubam md5s when we released Clio phase 1, so the plan is to just treat the md5 as optional in the backfill. Ditto for gvcfs, since their md5s aren't delivered. For crams, the backfill will check that we got an md5 from somewhere, and raise an error if not.

PR #175, opened 2017-11-15
DSDEGP-1779 Tweak Clio integration-test setup so it'll work on Jenkins

danxmoran:
### Description

Also (as is custom), I refactored our usage of `Indexable` and `HitReader` so that they're bound as fields to their corresponding `ElasticsearchIndex`, and not floating in the implicit ether all around the server. I made the change so I could write generic code for pre-seeding documents into storage in the `RecoveryTests`, and it let me delete a bunch of imports and context-bounds in the server classes.

[Here](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-integration/202/) is a Jenkins job that will run the Load and Recovery tests, to show things are working. I'll follow-up with a PR in dsp-jenkins.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
@@ -1,6 +1,5 @@
 package org.broadinstitute.clio.client.util
 
-import java.io.File
 import java.net.URI
 import java.nio.file.{FileVisitOption, Files, Path, Paths}
 import java.util.Comparator
@@ -38,8 +37,7 @@ trait IoUtil {
     Files
       .walk(directory, FileVisitOption.FOLLOW_LINKS)
       .sorted(Comparator.reverseOrder())
-      .map[File](path => path.toFile)
-      .forEach(file => file.deleteOnExit())
+      .forEach(Files.delete(_))
   }
 
   /*
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
@@ -1,8 +1,9 @@
 package org.broadinstitute.clio.client.util
 
+import java.io.IOException
 import java.net.URI
-import java.nio.file.{FileVisitOption, Files, Path, Paths}
-import java.util.Comparator
+import java.nio.file._
+import java.nio.file.attribute.BasicFileAttributes
 
 import scala.sys.process.{Process, ProcessBuilder}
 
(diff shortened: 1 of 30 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -38,8 +37,7 @@ trait IoUtil {
     Files
       .walk(directory, FileVisitOption.FOLLOW_LINKS)
       .sorted(Comparator.reverseOrder())
```
pshapiro4broad: For very deep trees this will be expensive. Using a file visitor instead would avoid having to read all file tree paths at once.

Also I wonder why we're using `FOLLOW_LINKS` here.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/Elastic4sAutoDerivation.scala
```diff
@@ -19,3 +19,5 @@ object Elastic4sAutoDerivation extends ModelAutoDerivation {
   implicit def implicitEncoder(json: Json): String =
     defaultPrinter.pretty(json)
 }
+
+object Elastic4sAutoDerivation extends Elastic4sAutoDerivation
```
pshapiro4broad: With this change do we still need an `object` here?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/Elastic4sAutoDerivation.scala
```diff
@@ -4,7 +4,7 @@ import io.circe.Json
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
 
 /** Automatic JSON encoding and decoding for the Elasticsearch DAO. */
-object Elastic4sAutoDerivation extends ModelAutoDerivation {
+trait Elastic4sAutoDerivation extends ModelAutoDerivation {
```
pshapiro4broad: The name of this class seems bad to me. I guess it was chosen to match `com.sksamuel.elastic4s.` but that package isn't used here. Seems like writing out `ElasticsearchAutoDerivation` would be better.

pshapiro4broad (approved):

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/Elastic4sAutoDerivation.scala
danxmoran: I think it's named this way because it's providing the `implicit printer` argument to [this method](https://github.com/sksamuel/elastic4s/blob/master/elastic4s-circe/src/main/scala/com/sksamuel/elastic4s/circe/package.scala#L41) in elastic4s-circe. I'll update the scaladoc to make it more clear what the trait provides.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -35,11 +35,22 @@ trait IoUtil {
   }
 
   def deleteDirectoryRecursively(directory: Path): Unit = {
-    Files
-      .walk(directory, FileVisitOption.FOLLOW_LINKS)
-      .sorted(Comparator.reverseOrder())
-      .map[File](path => path.toFile)
-      .forEach(file => file.deleteOnExit())
+    val _ =
+      Files.walkFileTree(directory, new SimpleFileVisitor[Path]() {
```
danxmoran, line 39: @pshapiro4broad I haven't used `FileVisitor` before, does this look sane?

danxmoran:
Latest integration-test run on Jenkins is [here](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-integration/204/). The one failure is from a version skew when running the dev-env tests; all the in-docker ones passed.

tbl3rd (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
pshapiro4broad: Looks good to me.

pshapiro4broad (approved):

PR #176, opened 2017-11-16
DSDEGP-1801 Add retries to clio-client.

danxmoran:
### Description

This sets up the web-client to retry whenever an exception is thrown during a request-response cycle, but not when a response is returned successfully with an error status code.

I ended up not using akka-http's built-in retry logic. After doing some digging on the specific exception KT hit last night I found we'd stumbled on a [known bug](https://github.com/akka/akka-http/issues/1459#issuecomment-335487195) in akka's "host-level" API, which also provides the retry logic I had mentioned earlier. The akka team is [completely rewriting](https://github.com/akka/akka-meta/issues/60) the internals of that API, so I moved us off of it and onto the lower-level "connection-level" API instead.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
@@ -36,21 +36,26 @@ trait IoUtil {
 
   def deleteDirectoryRecursively(directory: Path): Unit = {
     val _ =
-      Files.walkFileTree(directory, new SimpleFileVisitor[Path]() {
-        override def visitFile(file: Path,
-                               attrs: BasicFileAttributes): FileVisitResult = {
-          Files.delete(file)
-          FileVisitResult.CONTINUE
-        }
-
-        override def postVisitDirectory(dir: Path,
-                                        exc: IOException): FileVisitResult = {
-          Option(exc).fold({
-            Files.delete(dir)
+      Files.walkFileTree(
+        directory,
+        new SimpleFileVisitor[Path]() {
+          override def visitFile(
+            file: Path,
+            attrs: BasicFileAttributes
+          ): FileVisitResult = {
+            Files.delete(file)
             FileVisitResult.CONTINUE
-          })(throw _)
+          }
+
+          override def postVisitDirectory(dir: Path,
+                                          exc: IOException): FileVisitResult = {
+            Option(exc).fold({
+              Files.delete(dir)
+              FileVisitResult.CONTINUE
+            })(throw _)
+          }
         }
-      })
+      )
   }
 
   /*
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.clio.client.webclient
 
+import akka.Done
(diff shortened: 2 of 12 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -36,21 +36,26 @@ trait IoUtil {
 
   def deleteDirectoryRecursively(directory: Path): Unit = {
     val _ =
-      Files.walkFileTree(directory, new SimpleFileVisitor[Path]() {
```
danxmoran, line 39: Scalafmt

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -91,7 +70,26 @@ class ClioWebClient(
...
+
+        val response = responseStream
+          .recoverWithRetries(maxRequestRetries, {
+            case _ => responseStream
+          })
+          .runWith(Sink.head)
+
+        response.transformWith {
+          case Success(r)  => Future.successful(Success(r) -> promise)
+          case Failure(ex) => Future.successful(Failure(ex) -> promise)
```
pshapiro4broad: can you explain why this is returns `Future.successful()`?

pshapiro4broad (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
danxmoran: Yup! By default, `mapAsyncUnordered` will fail and shut down the web client's entire request-response stream if an element produces a failed `Future`, and we don't want that (otherwise apps like the backfill that use a single long-running client will fall over on any failure that pops up). There's a way to configure the stage to have it resume processing instead of stopping, but in that case it'll still drop the failed element, which we also don't want.

What we do want is to have any `Failure`s preserved and passed to `promise.failure`, so the thread listening on the other end of the `promise` can receive the exception and handle it appropriately. We make this happen by transforming all the `Future[HttpResponse]`s here into successful `Future[Try[HttpResponse]]`s, so the final stage in the stream can handle both successes and failures and we never crash out the stream.

I dug a bit more into the streams API and I think there might be a cleaner way to do it. I'll try it out, and if I can't make it work I'll add a scaladoc explaining all this.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -91,11 +71,64 @@ class ClioWebClient(
...
     )
-    .via(connectionFlow)
-    .toMat(Sink.foreach {
-      case (Success(response), promise) => promise.success(response)
-      case (Failure(ex), promise)       => promise.failure(ex)
-    })(Keep.left)
+    .mapAsyncUnordered(maxConcurrentRequests) {
+      case (request, promise) => {
+
+        /*
```
tlangs, line 77: Thanks for the detailed comments.

tlangs (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
pshapiro4broad: Makes sense, thanks for the explanation.

PR #178, opened 2017-11-20
DSDEGP-1812 Generalize existing move logic to allow setting arbitrary new basename

danxmoran:
### Description

This is needed for reset-ing filenames after testing a cram delivery with a patient-ID prefix, and is also useful in general.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
@@ -27,16 +27,24 @@ import scala.concurrent.{ExecutionContext, Future}
 class DeliverWgsCramExecutor(deliverCommand: DeliverWgsCram)
     extends Executor[UpsertId] {
 
-  val moveCommand =
-    MoveWgsCram(deliverCommand.key, deliverCommand.workspacePath)
-  import moveCommand.index.implicits._
+  import WgsCramIndex.implicits._
 
   override def execute(webClient: ClioWebClient, ioUtil: IoUtil)(
     implicit ec: ExecutionContext
   ): Future[UpsertId] = {
 
-    val moveExecutor =
-      new MoveExecutor(moveCommand, deliverCommand.samplePrefix)
+    val deliveredFileBasename: String =
+      deliverCommand.samplePrefix.fold(deliverCommand.key.sampleAlias) {
+        prefix =>
+          s"$prefix${deliverCommand.key.sampleAlias}"
+      }
+    val moveCommand = MoveWgsCram(
+      deliverCommand.key,
+      deliverCommand.workspacePath,
+      Some(deliveredFileBasename)
+    )
+
+    val moveExecutor = new MoveExecutor(moveCommand)
 
     for {
       _ <- moveExecutor.execute(webClient, ioUtil)
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
@@ -10,7 +10,8 @@ import org.broadinstitute.clio.transfer.model.WgsCramIndex
 import org.broadinstitute.clio.transfer.model.wgscram.{
   TransferWgsCramV1Metadata,
   TransferWgsCramV1QueryInput,
-  TransferWgsCramV1QueryOutput
+  TransferWgsCramV1QueryOutput,
+  WgsCramExtensions
 }
 import org.broadinstitute.clio.util.model.UpsertId
 
@@ -87,8 +88,12 @@ class DeliverWgsCramExecutor(deliverCommand: DeliverWgsCram)
            * we can write directory to the source location instead of writing
            * to temp and then copying.
            */
-          val md5Tmp = Files.createTempFile("clio-cram-deliver", "md5")
-          val cloudMd5Path = s"$cramPath.md5"
+          val md5Tmp = Files.createTempFile(
+            "clio-cram-deliver",
+            WgsCramExtensions.Md5Extension
+          )
+          val cloudMd5Path =
+            s"$cramPath${WgsCramExtensions.Md5ExtensionAddition}"
 
           try {
             Files.write(md5Tmp, cramMd5.name.getBytes)
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
@@ -12,6 +12,7 @@ import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
 }
 import org.broadinstitute.clio.transfer.model.GvcfIndex
 import org.broadinstitute.clio.transfer.model.gvcf.{
+  GvcfExtensions,
   TransferGvcfV1Key,
   TransferGvcfV1Metadata,
   TransferGvcfV1QueryOutput
@@ -81,7 +82,8 @@ trait GvcfTests { self: BaseIntegrationSpec =>
       sampleAlias = s"someAlias $randomId",
       version = 2,
       documentStatus = Some(DocumentStatus.Normal),
-      gvcfPath = Some(URI.create("gs://path/gvcf.g.vcf.gz"))
+      gvcfPath =
+        Some(URI.create(s"gs://path/gvcf${GvcfExtensions.GvcfExtension}"))
     )
 
     /*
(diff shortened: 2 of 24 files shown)


clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveGvcfSpec.scala
```diff
@@ -19,7 +19,8 @@ class MoveGvcfSpec extends BaseClientSpec {
     recoverToSucceededIf[IllegalArgumentException] {
       val command = MoveGvcf(
         key = testGvcfTransferV1Key,
-        destination = MockIoUtil.InvalidPath
+        destination = MockIoUtil.InvalidPath,
+        None
```
pshapiro4broad: could you use an overloaded constructor to avoid having to pass `None`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
```diff
@@ -27,16 +27,24 @@ import scala.concurrent.{ExecutionContext, Future}
...
+  import WgsCramIndex.implicits._
 
   override def execute(webClient: ClioWebClient, ioUtil: IoUtil)(
     implicit ec: ExecutionContext
   ): Future[UpsertId] = {
 
-    val moveExecutor =
-      new MoveExecutor(moveCommand, deliverCommand.samplePrefix)
+    val deliveredFileBasename: String =
+      deliverCommand.samplePrefix.fold(deliverCommand.key.sampleAlias) {
```
pshapiro4broad, line 38: Did you consider accepting a function that could transform the destination filename, instead of the new basename? I think it would make the code here slightly simpler.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -570,10 +572,10 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
       s"$randomId --- I am dummy detail metrics --- $randomId"
 
-    val gvcfName = s"$randomId.gvcf"
-    val indexName = s"$randomId.index"
-    val summaryMetricsName = s"$randomId.summary.metrics"
-    val detailMetricsName = s"$randomId.detail.metrics"
+    val gvcfName = s"$randomId.g.vcf.gz"
+    val indexName = s"$randomId.g.vcf.gz.tbi"
+    val summaryMetricsName = s"$randomId.variant_calling_summary_metrics"
+    val detailMetricsName = s"$randomId.variant_calling_detail_metrics"
```
pshapiro4broad: Even though this is test code, I think we want apply the same rigor to it regarding code duplication. Can the common test setup code be refactored into a common method?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
danxmoran: Not sure I understand... where would the function live? Right now `deliverCommand.samplePrefix` is coming from the command line, so I think it's stuck as an `Option[String]`. We need to perform this computation somewhere, so I'm not sure wrapping it in a function to be called by the methods in the `TransferMetadata` would be any simpler than calling it once here and passing in the result.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveGvcfSpec.scala
danxmoran: I can probably use a default value. Overloads don't play nicely with case classes AFAIK.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
danxmoran: Done! Glorious -642 lines

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/TransferWgsCramV1Metadata.scala
```diff
@@ -70,19 +69,16 @@ case class TransferWgsCramV1Metadata(
...
-    val prefixedCrai = prefixedCram.map(pc => URI.create(s"${pc}.crai"))
+    val movedCram = pathMapper(cramPath, ".cram")
     this.copy(
-      cramPath = pathMapper(prefixedCram),
+      cramPath = movedCram,
       // DSDEGP-1715: We've settled on '.cram.crai' as the extension and
       // want to fixup files with just '.crai' when possible.
-      craiPath = pathMapper(prefixedCrai)
+      craiPath = movedCram.map { cramUri =>
+        URI.create(s"${cramUri.toString}.crai")
```
pshapiro4broad: ~Seems like writing `cramUri + ".crai"` would be simpler here~ actually isn't `toString` unnecessary, e.g. `s"$cramUri.crai"` would work

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
pshapiro4broad: The function is a lambda created at the call site. Does the caller have enough information to replace the basename without needing the extension? If that could be avoided it would be much better. I don't know if we can be sure extensions of these files will never change in the future, and hardcoding them here doesn't seem great to me.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
pshapiro4broad: bitmoji

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
danxmoran: Ah! I think I get what you're suggesting now, but I still don't know how it'd work.

In the special case of cram delivery, the caller knows that it's manipulating files that should end in ".cram", ".crai", etc, but in general that's not the case. The move-with-new-basename functionality is built into the move executor, which knows very little about the files it's operating over, and the new basename is getting parsed straight from the CLI. There's no special-cased intermediate between the parser and the executor that could transform the new basename into a lambda, and no good way to specify the lambda directly on the command line.

I managed to convince myself yesterday that the least-bad way to implement this feature is to hardcode the expected extensions. The concrete situation Sam was trying to recover from was a bug in the delivery WDL that ended up naming things like `..NA12878.cram`. I couldn't think of a sane splitting algorithm that would handle cases like this along with:

- `NA12878.cram`
- `SomePatientId.NA12878.cram`
- `NA12878.cram.crai`
- `HK3T5CCXX.6.Pond-492100.unmapped.bam`
- `NA12878.v10.bb684530-32a6-4461-a9f5-84778f61d935.g.vcf.gz`

Since I had convinced myself that the only way to tell basename apart from extension would be to hardcode the extensions somewhere, I stuck them in the transfer-metadata classes, which is where we've put all the other index-specific logic.

danxmoran: But I agree there are definitely drawbacks to having things hard-coded. I built a jar from my branch and gave it to Sam to unblock him, so we can leave this open over the break and talk about it with the whole group to see if this is something we want to maintain long-term.

danxmoran:
Note from #179: if we do end up keeping the hard-coded extensions, it'd be good to expose them as constants so the integration-tests can use them.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferMetadata.scala
```diff
@@ -34,17 +34,17 @@ trait TransferMetadata[M <: TransferMetadata[M]] { self: M =>
     *
     * Until we move to a Path-based API, `destination` must end with '/'.
     */
-  def moveInto(destination: URI, samplePrefix: Option[String] = None): M = {
+  def moveInto(destination: URI, newBasename: Option[String] = None): M = {
```
tbl3rd, line 37: Probably gotta update the comment here too now.
Maybe for `mapMove()` as well.

tbl3rd (approved):
Wow. Should have seen this coming. (-Kinda did see this coming, but was swayed by the argument that we should allow only prefixing the sample name!-)

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverWgsCramExecutor.scala
tbl3rd: Type these members not as URIs but some ExendedUri type with an extension?

danxmoran: Made [DSDEGP-1827](https://broadinstitute.atlassian.net/browse/DSDEGP-1827) to track it.

danxmoran:
@pshapiro4broad I moved the file extensions into constants.

PR #189, opened 2018-01-17
DSDEGP-1963 Always use exact matching in queries on String fields

danxmoran:
### Description

We set up a nested keyword field on all our text fields when we switched the mapping for `String`s, but we didn't point the query mapper at them.

----

* Review cycle:


clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchIndex.scala
```diff
@@ -17,27 +17,27 @@ import scala.reflect.runtime.universe.Type
   * Builds an ElasticsearchIndex using shapeless and reflection.
   *
   * @param name The base name of the index.
-  * @param version The mapping version to use for the index.
+  * @param mappingsVersion The mapping version to use for the index.
```
pshapiro4broad: `mappingsVersion` not `mappingVersion`? The comment text says 'mapping'.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchQueryMapper.scala
```diff
@@ -92,6 +98,14 @@ class AutoElasticsearchQueryMapper[
           esName.stripSuffix("_end"),
           value.asInstanceOf[OffsetDateTime]
         )
+
+      case tpe
+          if indexVersion > 1 &&
```
pshapiro4broad: The way that `1` is used here seems like a magic number, e.g. it has a special meaning to the code. Replacing it with a name could make the code clearer, and will help a future maintainer if they want to find out everywhere this special version number is used.

pshapiro4broad: Another option would be to replace `indexVersion` with an object that computes `queryVal()` so the object encapsulates this logic, e.g. add it to `ElasticSearchIndex` instead of adding the version number. Or this work could be done in `ModelQueryInput` since that's the object that this method already has in scope.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -65,6 +68,12 @@ object ElasticsearchIndex extends Elastic4sAutoDerivation {
   val WgsCram: ElasticsearchIndex[DocumentWgsCram] =
     indexDocument[DocumentWgsCram](version = 2)
 
+  /**
+    * Name to assign to a nested keyword field under every text field,
+    * to support exact matching.
+    */
+  val TextExactMatchFieldName: String = "exact"
```
pshapiro4broad: Not sure why this is defined here when it's only used in another class. Does this class have another attribute which forces this field to have this name? That would be a reason to keep it in here.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/AutoElasticsearchQueryMapperSpec.scala
```diff
@@ -60,13 +60,20 @@ class AutoElasticsearchQueryMapperSpec extends FlatSpec with Matchers {
       mockKeyLong = None,
       mockKeyString = Option("hello")
     )
-    mapper.buildQuery(input) should be(
+    mapper.buildQuery(input, 1) should be(
```
tlangs: Same issue with "magic" number 1 here.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -0,0 +1,99 @@
...
+}
+
+private[dataaccess] object ElasticsearchFieldMapper
+    extends IntEnum[ElasticsearchFieldMapper] {
+  import com.sksamuel.elastic4s.http.ElasticDsl._
+
+  override val values: immutable.IndexedSeq[ElasticsearchFieldMapper] = findValues
+
+  /** All non-numeric, non-date members mapped to keywords. */
+  case object InitVersion extends ElasticsearchFieldMapper(1) {
```
tlangs: Why not name this `Version1` and the other once `Version2` since its possible that we could have a third version sometime in the future?

danxmoran: Having `Version` in the name at all is an accident from earlier logic... will clean up.

I'll also rename `Init` to be more descriptive about how it actually maps things. I didn't want to go with a `VersionN` pattern because it's not really more informative than just having an int. Not sure what you mean about the third version... if we end up needing it we'll have to make a new `case object` either way, and we can name that object to describe the change in mapping logic it introduces.

pshapiro4broad (approved):

tlangs (approved):

PR #194, opened 2018-01-25
DSDEGP-1990: Make Clio throw an error on double quotes in stringy inputs

als364:
### Description

For [DSDEGP-1990](https://broadinstitute.atlassian.net/browse/DSDEGP-1990), update [ClioParsers.scala](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala) to reject stringy arguments with leading or trailing quotes, and update [ParsersSpec.scala](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/test/scala/org/broadinstitute/clio/client/commands/ParsersSpec.scala) with a test case for that parser.

----

  * (I am not adding/switching libraries or otherwise changing the design.)
* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
@@ -62,15 +62,10 @@ trait ClioParsers {
 
   implicit val stringParser: ArgParser[String] =
     SimpleArgParser.from[String]("string") { str =>
-      if (str contains "\"") {
-        Left(
-          Error.MalformedValue(
-            "string",
-            "Quotes are not allowed in inputs"
-          )
-        )
-      } else {
-        Right(str)
-      }
+      Either.cond(
+        !str.contains("\""),
+        str,
+        Error.MalformedValue("string", "Quotes are not allowed in inputs")
+      )
     }
 }
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
@@ -62,11 +62,11 @@ trait ClioParsers {
 
   implicit val stringParser: ArgParser[String] =
     SimpleArgParser.from[String]("string") { str =>
-      if (str.startsWith("\"") || str.endsWith("\"")) {
+      if (str contains "\"") {
         Left(
           Error.MalformedValue(
             "string",
-            "Quotes are not allowed at the beginning or end of inputs"
+            "Quotes are not allowed in inputs"
           )
         )
       } else {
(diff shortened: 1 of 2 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
```diff
@@ -59,4 +59,18 @@ trait ClioParsers {
   implicit val pathParser: ArgParser[Path] = parseWith("path", Paths.get(_))
 
   implicit val uuidParser: ArgParser[UUID] = parseWith("uuid", UUID.fromString)
+
+  implicit val stringParser: ArgParser[String] =
+    SimpleArgParser.from[String]("string") { str =>
+      if (str.startsWith("\"") || str.endsWith("\"")) {
```
pshapiro4broad: Can't we disallow quotes anywhere in the input? Also, it seems like we'd also want to disallow single quotes.

als364: @danxmoran thought it would be leading/trailing only, and the Jira specified double quotes. Have we seen escaped quotes in the middle of strings in the wild?

danxmoran: Yeah, my worry with disallowing quotes anywhere is that the "notes" field might contain them somewhere inside (i.e. I can see myself writing `Deleting because KT said "..."` as a note for a delete operation.

Agreed single-quotes would also be good to check for, but definitely only on the edges (otherwise we break contractions, also in the notes field).

als364: In the `Someone said "something"` case, would we still want to allow quotes on the trailing edge of the input?

danxmoran: Double-checked with KT, and she said it's OK to just disallow double-quotes entirely as long as single-quotes can work.

danxmoran (approved):

clio-client/src/test/scala/org/broadinstitute/clio/client/commands/ParsersSpec.scala
```diff
@@ -117,4 +117,79 @@ class ParsersSpec extends BaseClientSpec {
     }
     ignoreDeleted should be(true)
   }
+
+  it should "double quotes on the ends of String inputs should cause an error" in {
```
tlangs: not a big deal, but it just reads better if it were `it should "error out when there are double quotes on the ends of String inputs" in {`

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioParsers.scala
```diff
@@ -59,4 +59,18 @@ trait ClioParsers {
...
+      if (str contains "\"") {
+        Left(
+          Error.MalformedValue(
+            "string",
+            "Quotes are not allowed in inputs"
+          )
+        )
+      } else {
+        Right(str)
+      }
```
tlangs: The `cats` library already in use in Clio provides a nice syntax for creating `Either`s based on a condition. This block can be written as:
```scala
Either.cond(
        !str.contains("\""),
        str,
        Error.MalformedValue("string", "Quotes are not allowed in inputs")
      )
```

tlangs (changes requested):

tlangs (approved):

PR #195, opened 2018-01-25
DSDEGP-1989 Support specifying custom hostname in clio deploy script.

danxmoran:
### Description

This adds a new argument `CLIO_HOST_NAME` to Clio's deploy script, to enable deploying to an arbitrary host. I took the chance to clean up the script's structure.

----

* Review cycle:

### DIFF
--- a/clio-server/config/docker-compose.yml.ctmpl
+++ b/clio-server/config/docker-compose.yml.ctmpl
@@ -18,7 +18,7 @@ services:
 
   proxy:
     image: broadinstitute/openidc-proxy:latest
-    hostname: {{env "CLIO_HOST"}}
+    hostname: {{env "CLIO_DOMAIN_NAME"}}
     links:
       - app:app
       - logger
@@ -35,11 +35,11 @@ services:
       - logger
     restart: always
     environment:
-      CALLBACK_URI: https://{{env "CLIO_HOST"}}/oauth2callback
+      CALLBACK_URI: https://{{env "CLIO_DOMAIN_NAME"}}/oauth2callback
       LOG_LEVEL: warn
       PROXY_PATH: /
       PROXY_URL: http://app:{{env "CONTAINER_CLIO_PORT"}}/
-      SERVER_NAME: {{env "CLIO_HOST"}}
+      SERVER_NAME: {{env "CLIO_DOMAIN_NAME"}}
       AUTH_TYPE: AuthType None
       OIDC_CLAIM: Require all granted
       PROXY_PATH2: /api
--- a/clio-server/config/docker-compose.yml.ctmpl
+++ b/clio-server/config/docker-compose.yml.ctmpl
@@ -18,7 +18,7 @@ services:
 
   proxy:
     image: broadinstitute/openidc-proxy:latest
-    hostname: {{env "CLIO_DOMAIN_NAME"}}
+    hostname: {{env "CLIO_FQDN"}}
     links:
       - app:app
       - logger
@@ -35,11 +35,11 @@ services:
       - logger
     restart: always
     environment:
-      CALLBACK_URI: https://{{env "CLIO_DOMAIN_NAME"}}/oauth2callback
+      CALLBACK_URI: https://{{env "CLIO_FQDN"}}/oauth2callback
       LOG_LEVEL: warn
       PROXY_PATH: /
       PROXY_URL: http://app:{{env "CONTAINER_CLIO_PORT"}}/
-      SERVER_NAME: {{env "CLIO_DOMAIN_NAME"}}
+      SERVER_NAME: {{env "CLIO_FQDN"}}
       AUTH_TYPE: AuthType None
       OIDC_CLAIM: Require all granted
       PROXY_PATH2: /api
--- a/jenkins/deploy-clio-server.sh
+++ b/jenkins/deploy-clio-server.sh
@@ -5,6 +5,11 @@ set -ex
 declare -r PROG_NAME=$(basename $0)
 declare -r CLIO_DIR=$(cd $(dirname $(dirname $0)) && pwd)
 
+declare -r -a VAULT_TOKEN_FILE_OPTIONS=(/etc/vault-token-dsde ${HOME}/.vault-token)
+
+# Shamelessly ripped from StackOverflow: https://stackoverflow.com/a/17841619
+join_by() { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
+
 usage() {
   cat <<EOF >&2
 USAGE: ENV=<env> CLIO_HOST_NAME=<hostname> $PROG_NAME
@@ -25,8 +30,7 @@ Note: requires the following ENV vars to be set to non-null values:
 
 Also requires that a file containing a vault token be placed at one of the following paths:
 
-  /etc/vault-token-dsde
-  ${HOME}/.vault-token
+  $(join_by $'\n  ' ${VAULT_TOKEN_FILE_OPTIONS[@]})
 EOF
 }
 
@@ -38,7 +42,7 @@ check_env_args() {
     ret=1
   else
     case "$ENV" in
-      "dev"|"staging"|"prod")
+      dev|staging|prod)
         ;;
       *)
         >&2 echo "Error: Invalid environment '$ENV'!"
@@ -59,7 +63,7 @@ check_env_args() {
 }
 
 render_ctmpls() {
-  local -r docker_tag=$1 clio_domain_name=$2 tmp=$3 app_dir=$4 host_log_dir=$5 pos_log_dir=$6
+  local -r docker_tag=$1 clio_fqdn=$2 tmp=$3 app_dir=$4 host_log_dir=$5 pos_log_dir=$6
 
   local -r ctmpl_dir="${CLIO_DIR}/clio-server/config"
   local vault_token_file
@@ -70,12 +74,14 @@ render_ctmpls() {
   fi
 
   # Try to find the vault token location.
-  if [ -f /etc/vault-token-dsde ]; then
-    vault_token_file=/etc/vault-token-dsde
-  elif [ -f ${HOME}/.vault-token ]; then
-    vault_token_file=${HOME}/.vault-token
-  else
-    >&2 echo "Error: Vault token file not found at /etc/vault-token-dsde or ${HOME}/.vault-token"
+  for token_location in ${VAULT_TOKEN_FILE_OPTIONS[@]}; do
+    if [ -f $token_location ]; then
+      vault_token_file=$token_location
+    fi
+  done
+
+  if [ -z "$vault_token_file" ]; then
+    >&2 echo "Error: Vault token file not found at any of $(join_by ', ' ${VAULT_TOKEN_FILE_OPTIONS[@]})"
     exit 1
   fi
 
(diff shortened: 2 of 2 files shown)


jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  # Name of the service account JSON file used by the Clio instance.
+  local -r clio_service_account_json=clio-service-account.json
+
+  # Port to expose for Clio in the container.
+  local -r container_clio_port=8080
+
+  # Populate an env file with variables for rendering configs.
+  local -r ctmpl_env_file="${tmp}/env-vars.txt"
+  for var in ENV clio_domain_name docker_tag app_dir clio_conf_dir host_log_dir clio_log_dir pos_log_dir clio_app_conf clio_logback_conf clio_service_account_json container_clio_port; do
+    echo "${var^^}=${!var}" >> ${ctmpl_env_file}
```
tlangs: Comment here to explain bash magic

tlangs (approved):
Could you document the less-used bash features?

jenkins/deploy-clio-server.sh
pshapiro4broad: I agree! The original code `echo "${var}=${!var}"` made sense to me, not sure what `^^` is doing here.

clio-server/config/docker-compose.yml.ctmpl
```diff
@@ -18,7 +18,7 @@ services:
 
   proxy:
     image: broadinstitute/openidc-proxy:latest
-    hostname: {{env "CLIO_HOST"}}
+    hostname: {{env "CLIO_DOMAIN_NAME"}}
```
pshapiro4broad: This is a bit confusing to me as I think about host and domain names as the same things. It looks like in this case `CLIO_DOMAIN_NAME` refers to the fully qualified domain name. Not sure it's worth calling this `CLIO_FQDN` or something similar, but the fact that "domain name" and "host name" are synonyms makes it a little harder to understand.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
pshapiro4broad: Is there a case where `bash` isn't at `/bin/bash`, or we want to execute a different `bash` than the default one?

tlangs: the location of bash will probably not change, but `/usr/bin/env bash` is the accepted standard for how to denote a bash script.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+
+check_env_args() {
+  local ret=0
+
+  if [ -z "$ENV" ]; then
+    >&2 echo "Error: ENV not set!"
+    ret=1
+  else
+    case "$ENV" in
+      "dev"|"staging"|"prod")
```
pshapiro4broad: unnecessary double quotes around case labels

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
-
-rm -f ${CTMPL_ENV_FILE}
-
-stopcontainer() {
-  # stop running container on host
-  ${SSHCMD} "test -f ${APP_DIR}/docker-compose.yml && docker-compose -p clio-server -f ${APP_DIR}/docker-compose.yml stop || echo"
-
-  # docker-compose rm -f
-  ${SSHCMD} "test -f ${APP_DIR}/docker-compose.yml && docker-compose -p clio-server -f ${APP_DIR}/docker-compose.yml rm -f || echo"
+declare -r PROG_NAME=$(basename $0)
```
tbl3rd: $(basename $0) is usually just ${0##*/}

danxmoran:
Proof that it works: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-deploy/373

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+declare -r CLIO_DIR=$(cd $(dirname $(dirname $0)) && pwd)
```
tbl3rd: Probably OK but that depends on how the script is invoked.
Maybe declare -r CLIO_DIR=$(cd $(dirname $0)/.. && pwd)
That works for x, x.sh, ./x.sh, ./clio/jenkins/x.sh, and /.../clio/jenkins/x.sh, and so on.
The script is inconsistent in quoting $CLIO_DIR and its derivatives, BTW.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  3. Stop any currently-running clio container network on the host
+  4. Start a new network using the new configs and docker image
+  5. Verify that the new Clio server reports OK health and the expected version
+  6. Roll back the deploy if the new server fails to report its health or version
+
+Note: requires the following ENV vars to be set to non-null values:
+
+  ENV:            Environment to deploy to, one of "dev", "staging", or "prod"
+  CLIO_HOST_NAME: Name of the host within the network specified by 'ENV' to deploy to,
+                  for example: "clio101", "clio201", "clio3", "clio"
```
tbl3rd: Thank you for enumerating values and providing examples.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  /etc/vault-token-dsde
+  ${HOME}/.vault-token
+EOF
+}
+
+check_env_args() {
+  local ret=0
+
+  if [ -z "$ENV" ]; then
+    >&2 echo "Error: ENV not set!"
```
tbl3rd: Do not need the "s here.

jenkins/deploy-clio-server.sh
tbl3rd: Don't need the "s here either. Shell has "bare words" as in Perl or Ruby.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  if [ -z "$CLIO_HOST_NAME" ]; then
+    >&2 echo "Error: CLIO_HOST_NAME not set!"
+    ret=1
+  fi
+
+  if [ $ret -ne 0 ]; then
+    usage
+    exit 1
+  fi
+}
```
tbl3rd: Cool, but a bit too fussy? Maybe just check that "$ENV" is one of the enumerated values and "$CLIO_HOST_NAME" is set, and just show the usage message when not. (This detail would
be more worth implementing if there were a dozen different settings rather than just a couple.)

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+
+render_ctmpls() {
+  local -r docker_tag=$1 clio_domain_name=$2 tmp=$3 app_dir=$4 host_log_dir=$5 pos_log_dir=$6
```
tbl3rd: Nice.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+                  for example: "clio101", "clio201", "clio3", "clio"
+
+Also requires that a file containing a vault token be placed at one of the following paths:
+
+  /etc/vault-token-dsde
+  ${HOME}/.vault-token
+EOF
+}
+
+check_env_args() {
```
tbl3rd: This is really validating the command line or usage conformance.
You could move more checks here, such as the vault token file.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  local -r docker_tag=$1 clio_domain_name=$2 tmp=$3 app_dir=$4 host_log_dir=$5 pos_log_dir=$6
+
+  local -r ctmpl_dir="${CLIO_DIR}/clio-server/config"
+  local vault_token_file
+
+  if [ ! -d "$ctmpl_dir" ]; then
+    >&2 echo "Error: Missing ctmpl source directory ($ctmpl_dir) - Was it accidentally deleted?"
+    exit 1
+  fi
+
```
tbl3rd: Move the following into the prior function?
(The usage message mentions this requirement.)

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  # Try to find the vault token location.
+  if [ -f /etc/vault-token-dsde ]; then
+    vault_token_file=/etc/vault-token-dsde
+  elif [ -f ${HOME}/.vault-token ]; then
+    vault_token_file=${HOME}/.vault-token
+  else
+    >&2 echo "Error: Vault token file not found at /etc/vault-token-dsde or ${HOME}/.vault-token"
+    exit 1
+  fi
+
```
tbl3rd: The following basically implements an "associative array" (map in shell parlance).
Bash has associative arrays.

jenkins/deploy-clio-server.sh
tbl3rd: I'd just uppercase the variables. Local shmokal.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+  rm ${ctmpl_env_file}
 }
 
-startcontainer() {
-  # docker-compose pull
-  ${SSHCMD} "docker-compose -p clio-server -f ${APP_DIR}/docker-compose.yml pull"
+compose_cmd() {
+  echo "docker-compose -p clio-server -f ${1}/docker-compose.yml $2"
+}
```
tbl3rd: compose_cmd() is really just a command prefix (like ssh_cmd) since $1 is always $app_dir ... I think.
Uh ... ok ... maybe not ...
(You could use an array for this to keep the quoting simple.)

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+
+stop_containers() {
+  local -r ssh_cmd=$1 app_dir=$2
+
+  # Stop any running Clio network on the host.
+  ${ssh_cmd} "test -f ${app_dir}/docker-compose.yml && $(compose_cmd ${app_dir} stop) || echo"
+
+  # Remove any stopped containers.
+  ${ssh_cmd} "test -f ${app_dir}/docker-compose.yml && $(compose_cmd ${app_dir} 'rm -f') || echo"
+}
```
tbl3rd: || echo to avoid abort on error?

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
-  then
-    sleep 20
-    echo "Attempt $attempts"
-    STATUS=$(curl -fs https://"${CLIO_HOST}/health" | jq -r .search)
-    REPORTED_VERSION=$(curl -fs https://"${CLIO_HOST}/version" | jq -r .version)
-    attempts=$((attempts + 1))
+  # Start up the new containers.
+  start_containers "${ssh_cmd}" ${app_dir}
+}
+
```
tbl3rd: I think inverting the loop sense in poll_clio_health would be cleaner.

jenkins/deploy-clio-server.sh
```diff
@@ -1,210 +1,256 @@
...
+      reported_version=$(curl -fs "https://${clio_domain_name}/version" | jq -r .version)
+      attempts=$((attempts + 1))
+    else
+      return 1
+    fi
+  done
+
+  return 0
+}
+
```
tbl3rd: This script is often creating strings, modifying them, and then using the results as command lines.
OK, that's what everyone does ... but ...
It is usually (always?) better to manage command lines as argument vectors (bash arrays)
to avoid quote and eval problems. (Let me know if you're interested in learning more.)

tbl3rd (approved):
Good. I think the script would benefit from factoring into more smaller functions. Factor out the vault token management, for example -- maybe even the rollback. Do all the command line and environment validation up front in functions separate from the side-effecting ones.

clio-server/config/docker-compose.yml.ctmpl
tbl3rd: FWIW: I like FQDN too.

PR #196, opened 2018-01-26
DSDEGP-2025 Add more IO utils for use in the backfill.

danxmoran:
### Description

When back-filling, we've been making 3 gsutil calls per main file:
1. `gsutil stat` to check that the ubam / gvcf / cram exists
2. `gsutil du` to get the size of the file
3. `gsutil hash` to get the md5 of the file

A single call to `gsutil stat` can give us all of these things. Since `gsutil` can be really laggy and resource-greedy on-prem, changing to a single `gsutil` call should give the backfill a nice speed bump / make it less likely to munge other running programs.

This PR exposes the `stat` method in our `gsutil` wrapper. There will be a follow-up PR in spitfire to use it.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
@@ -1,11 +1,14 @@
 package org.broadinstitute.clio.client.util
 
 import java.io.IOException
+import java.math.BigInteger
 import java.net.URI
 import java.nio.file._
 import java.nio.file.attribute.BasicFileAttributes
+import java.util.Base64
 
 import scala.sys.process.{Process, ProcessBuilder}
+import scala.util.Try
 
 /**
   * Clio-client component handling all file IO operations.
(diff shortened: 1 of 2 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -111,6 +116,28 @@ trait IoUtil {
   def listGoogleObjects(path: URI): Seq[String] = {
     gsUtil.ls(path.toString)
   }
+
+  /**
+    * Check if a path points to an object in GCS.
+    *
+    * If so, return the object's size and md5 hash (if present).
+    */
+  def getGoogleObjectInfo(path: URI): Option[(Long, Option[Symbol])] = {
```
danxmoran: Hrm, not sure about this returning an `Option[_]` any more. When I use it in the backfill I end up doing a `.getOrElse { throw new RuntimeException(...) }` on the result every time. Thoughts from the reviewer?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -88,15 +91,17 @@ trait IoUtil {
     gsUtil.exists(path.toString) == 0
   }
 
-  private val md5HashPattern = "Hash \\(md5\\):\\s+([0-9a-f]+)".r
+  private val sizePattern = "Content-Length:\\s+([0-9]+)".r
+  private val md5Base64Pattern = "Hash \\(md5\\):\\s+(.+)".r
+  private val md5HexPattern = "Hash \\(md5\\):\\s+([0-9a-f]+)".r
```
jacarey: Should we put these in the object instead of the trait?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
jacarey: would someone use this method to test existence? If so I think the Option makes sense. If not I think its fine to remove it.

jacarey (approved):

pshapiro4broad (approved):

PR #199, opened 2018-01-30
DSDEGP-1709: Switch several keyword fields to text

als364:
### Description

Using [this comment from October](https://github.com/broadinstitute/tbl/blob/master/bugs/DSDEGP-1640.org#final-decision) as a guide, moved most `String` fields to `Symbol` and updated the unit tests accordingly.

I understand that the deployment for this will be Tricky because we'll need to build indexes on fields that have not had indexes before, but I don't have a good idea of what exactly the deployment entails.

----

* Review cycle:


clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentWgsUbam.scala
```diff
@@ -17,36 +17,36 @@ case class DocumentWgsUbam(
...
-  productName: Option[String] = None,
-  productOrderId: Option[String] = None,
-  productPartNumber: Option[String] = None,
-  project: Option[String] = None,
-  readStructure: Option[String] = None,
+  productFamily: Option[Symbol] = None,
+  productName: Option[Symbol] = None,
+  productOrderId: Option[Symbol] = None,
+  productPartNumber: Option[Symbol] = None,
+  project: Option[Symbol] = None,
```
danxmoran: I talked with KT and Charlotte about the mismatch between `Symbol` for project in ubams and `String` for project in gvcfs / crams. Ops has some use-cases where it's helpful to wildcard-match project, so I think this should stay as `String`. `researchProjectId` should also stay as `String` by extension (because a cram `project` can come from either a `project` or `researchProjectId` on the ubam side).

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -71,6 +71,9 @@ private[dataaccess] object ElasticsearchFieldMapper
     private def isDate(tpe: Type): Boolean =
       tpe =:= typeOf[OffsetDateTime] || tpe =:= typeOf[Option[OffsetDateTime]]
 
+    private def isSymbol(tpe: Type): Boolean =
+      tpe =:= typeOf[Symbol] || tpe =:= typeOf[Option[Symbol]]
```
danxmoran: We can take the chance to clean up other existing code by using this in `stringToDefinition` above, too. Same with `isDate`.

danxmoran: Also, something to try: does a generic `is[A]` method work?
```scala
private def is[A](tpe: Type): Boolean =
  ... tpe checks, including check for Seq[A] ...
```
Not sure if it'll get screwed up by type erasure.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/UbamTests.scala
```diff
@@ -323,15 +323,15 @@ trait UbamTests { self: BaseIntegrationSpec =>
...
+        result.project should be(Some(Symbol(project)))
       }
       sampleResults should have length 2
       sampleResults.foldLeft(succeed) { (_, result) =>
         result.sampleAlias should be(Some(samples.head))
       }
       rpIdResults should have length 1
       rpIdResults.headOption.flatMap(_.researchProjectId) should be(
-        researchProjectIds.lastOption
+        Some(Symbol(researchProjectIds.last))
```
danxmoran: `researchProjectIds.lastOption.map(Symbol.apply)` is better practice (even though this is test code and we know the list won't be empty)

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ubam/TransferUbamV1QueryOutput.scala
```diff
@@ -14,36 +14,36 @@ case class TransferUbamV1QueryOutput(
   lane: Int,
   libraryName: String,
   location: Location,
-  analysisType: Option[String] = None,
-  baitIntervals: Option[String] = None,
-  baitSet: Option[String] = None,
-  dataType: Option[String] = None,
+  analysisType: Option[Symbol] = None,
+  baitIntervals: Option[Symbol] = None,
```
danxmoran, line 18: This is going to point to a file, so `URI` might be more appropriate... but also it'll be a reference file that Clio should never touch, so maybe `Symbol` is OK to keep it pretty abstract. What does reviewer 2 think?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ubam/TransferUbamV1QueryOutput.scala
```diff
@@ -14,36 +14,36 @@ case class TransferUbamV1QueryOutput(
...
-  sampleGender: Option[String] = None,
-  sampleId: Option[String] = None,
-  sampleLsid: Option[String] = None,
-  sampleType: Option[String] = None,
-  targetIntervals: Option[String] = None,
+  sampleGender: Option[Symbol] = None,
+  sampleId: Option[Symbol] = None,
+  sampleLsid: Option[Symbol] = None,
+  sampleType: Option[Symbol] = None,
+  targetIntervals: Option[Symbol] = None,
```
danxmoran, line 46: This one will also point to a reference file.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ubam/TransferUbamV1QueryOutput.scala
danxmoran: (We can also decide to swap it later without rebuilding search, since both `Symbol` and `URI` map to a keyword field)

danxmoran (changes requested):

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
als364: It won't if I explicitly call `A` a `TypeTag`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -20,6 +20,16 @@ private[dataaccess] sealed abstract class ElasticsearchFieldMapper(
   def stringToDefinition(fieldType: Type): String => FieldDefinition
 
   def valueToQuery(fieldName: String, fieldType: Type): Any => QueryDefinition
+
+  private def is[A: TypeTag](tpe: Type): Boolean = {
+    /*
+     * Elasticsearch has no notion of an array field; instead, any
+     * field can hold 1+ values of the same type.
+     */
+    tpe =:= typeOf[A] || tpe =:= typeOf[Option[A]] || tpe =:= typeOf[Seq[A]] || tpe =:= typeOf[
```
danxmoran: For the `Seq` types, I think we want `<:<` instead of `=:=` so we catch subtypes too.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -32,34 +42,18 @@ private[dataaccess] object ElasticsearchFieldMapper
...
+        case tpe if super.is[Long](tpe)           => longField
+        case tpe if super.is[Float](tpe)          => floatField
+        case tpe if super.is[Double](tpe)         => doubleField
+        case tpe if super.is[String](tpe)         => keywordField
+        case tpe if super.is[Symbol](tpe)         => keywordField
+        case tpe if super.is[EnumEntry](tpe)      => keywordField
+        case tpe if super.is[OffsetDateTime](tpe) => dateField
+        case tpe if super.is[UpsertId](tpe)       => keywordField
+        case tpe if super.is[UUID](tpe)           => keywordField
+        case tpe if super.is[URI](tpe)            => keywordField
```
danxmoran: I don't think you need `super.` for these.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -20,6 +20,16 @@ private[dataaccess] sealed abstract class ElasticsearchFieldMapper(
   def stringToDefinition(fieldType: Type): String => FieldDefinition
 
   def valueToQuery(fieldName: String, fieldType: Type): Any => QueryDefinition
+
+  private def is[A: TypeTag](tpe: Type): Boolean = {
```
danxmoran: Should be `protected`, since subclasses use it.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -68,22 +62,23 @@ private[dataaccess] object ElasticsearchFieldMapper
...
-    private def isDate(tpe: Type): Boolean =
-      tpe =:= typeOf[OffsetDateTime] || tpe =:= typeOf[Option[OffsetDateTime]]
-
     override def valueToQuery(
       fieldName: String,
       fieldType: Type
     ): Any => QueryDefinition = {
       fieldType match {
-        case tpe if isDate(tpe) && fieldName.endsWith("_start") =>
+        case tpe if super.is[OffsetDateTime](tpe) && fieldName.endsWith("_start") =>
```
danxmoran: Same `super.` comment throughout here.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
als364: I do if it's `private`, apparently. I definitely don't if it's `protected`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/UbamTests.scala
```diff
@@ -331,7 +331,7 @@ trait UbamTests { self: BaseIntegrationSpec =>
       }
       rpIdResults should have length 1
       rpIdResults.headOption.flatMap(_.researchProjectId) should be(
-        researchProjectIds.lastOption
+        Some(researchProjectIds.lastOption)
```
danxmoran: I don't think you want the `Some` wrapper here... you'll end up with `Option[Option[_]]`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -32,34 +42,18 @@ private[dataaccess] object ElasticsearchFieldMapper
...
-        case tpe if tpe <:< typeOf[Option[Seq[URI]]]    => keywordField
+        case tpe if is[Boolean](tpe)        => booleanField
+        case tpe if is[Int](tpe)            => intField
+        case tpe if is[Long](tpe)           => longField
+        case tpe if is[Float](tpe)          => floatField
+        case tpe if is[Double](tpe)         => doubleField
+        case tpe if is[String](tpe)         => keywordField
+        case tpe if is[Symbol](tpe)         => keywordField
+        case tpe if is[EnumEntry](tpe)      => keywordField
+        case tpe if is[OffsetDateTime](tpe) => dateField
```
danxmoran: (Since I made another comment) Nitpick: can you swap this line and the `EnumEntry` line, so all the `keywordField`s are together?

danxmoran (approved):
Some last things, LGTM once addressed. Really good cleanup in the ElasticsearchFieldMapper!

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ubam/TransferUbamV1QueryOutput.scala
pshapiro4broad: I would keep `Symbol` since it's opaque data to clio.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -20,6 +20,16 @@ private[dataaccess] sealed abstract class ElasticsearchFieldMapper(
   def stringToDefinition(fieldType: Type): String => FieldDefinition
 
   def valueToQuery(fieldName: String, fieldType: Type): Any => QueryDefinition
+
+  protected def is[A: TypeTag](tpe: Type): Boolean = {
+    /*
+     * Elasticsearch has no notion of an array field; instead, any
+     * field can hold 1+ values of the same type.
+     */
+    tpe =:= typeOf[A] || tpe =:= typeOf[Option[A]] || tpe <:< typeOf[Seq[A]] || tpe <:< typeOf[
```
pshapiro4broad, line 29: Nice refactor!

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -100,7 +100,7 @@ object ElasticsearchIndex extends Elastic4sAutoDerivation {
     implicitly[ElasticsearchIndex[Document]]
 
   implicit val WgsUbam: ElasticsearchIndex[DocumentWgsUbam] =
-    ElasticsearchIndex(ElasticsearchFieldMapper.NumericBooleanDateAndKeywordFields)
+    ElasticsearchIndex(ElasticsearchFieldMapper.StringsToTextFieldsWithSubKeywords)
```
pshapiro4broad: With this change, do we still need two different mappers? I thought we added a 2nd because we needed different behaviors for different indexes. If all indexes use the same mapper I think we can fold them into a single class.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/UbamTests.scala
```diff
@@ -360,8 +360,8 @@ trait UbamTests { self: BaseIntegrationSpec =>
     }
 
     val upsertData = TransferUbamV1Metadata(
-      sampleAlias = metadata.sampleAlias,
-      project = metadata.project
+      project = metadata.project,
```
pshapiro4broad, line 363: Not sure what happened here and below on line 655 (and in `TransferUbamV1Metadata.markDeleted` below). Are these intentional changes?

pshapiro4broad (approved):
👍  once (minor) comments are addressed

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/UbamTests.scala
als364: No, I probably did a keyboard shortcut in IntelliJ accidentally.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
```diff
@@ -20,6 +20,16 @@ private[dataaccess] sealed abstract class ElasticsearchFieldMapper(
   def stringToDefinition(fieldType: Type): String => FieldDefinition
 
   def valueToQuery(fieldName: String, fieldType: Type): Any => QueryDefinition
+
+  protected def is[A: TypeTag](tpe: Type): Boolean = {
```
tlangs, line 24: I love the refactor, but this can actually be even cleaner syntactically. By defining: 
```scala
implicit class IsTypeTag(tpe: Type) {
    def is[A: TypeTag]: Boolean = {
      /*
     * Elasticsearch has no notion of an array field; instead, any
     * field can hold 1+ values of the same type.
     */
      tpe =:= typeOf[A] || tpe =:= typeOf[Option[A]] || tpe <:< typeOf[Seq[A]] || tpe <:< typeOf[
        Option[Seq[A]]
        ]
    }
  }
```
checking the type of a `tpe` becomes `tpe.is[Boolean]`, instead of `is[Boolean](tpe)`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
als364: It looks like there's probably some sneaky dependencies, but I'll take a timeboxed look into it

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchFieldMapper.scala
danxmoran: If we were doing this sort of check all over the place it might be worth the DSL-y-ness, but I'm not sure it's worth doing here...

pshapiro4broad: Seem like more or less the same to me, both seem equally readable IMO.

PR #208, opened 2018-02-08
DSDEGP-2068 Expose parameter for controlling parallelism in recovery.

danxmoran:
### Description

Went with a config param so we can experiment without needing to build a new docker for each new thread count.

----

* Review cycle:

### DIFF
--- a/clio-integration-test/src/it/resources/org/broadinstitute/clio/integrationtest/clio-compose.yml
+++ b/clio-integration-test/src/it/resources/org/broadinstitute/clio/integrationtest/clio-compose.yml
@@ -5,7 +5,7 @@ services:
   clio-server:
     image: broadinstitute/clio-server:${CLIO_DOCKER_TAG}
     environment:
-      JAVA_OPTS: -Xmx128m -Xms128m -Dclio.server.elasticsearch.http-hosts.0.hostname=${ELASTICSEARCH_HOST} -Dlogback.configurationFile=/app/clio-logback.xml -Dclio.server.persistence.root-dir=/persistence
+      JAVA_OPTS: -Xmx128m -Xms128m -Dclio.server.persistence.recovery-parallelism=4 -Dclio.server.elasticsearch.http-hosts.0.hostname=${ELASTICSEARCH_HOST} -Dlogback.configurationFile=/app/clio-logback.xml -Dclio.server.persistence.root-dir=/persistence
     volumes:
       - ${CONF_DIR}/clio-logback.xml:/app/clio-logback.xml
       - ${CLIO_LOG_FILE}:/logs/clio.log
--- a/clio-server/config/clio.conf.ctmpl
+++ b/clio-server/config/clio.conf.ctmpl
@@ -10,6 +10,7 @@ clio.server {
     project-id: broad-gotc-{{env "ENV"}}-storage
     bucket: broad-gotc-{{env "ENV"}}-clio
     service-account-json: {{env "CLIO_CONF_DIR"}}/{{env "CLIO_SERVICE_ACCOUNT_JSON"}}
+    recovery-parallelism: 8
   }
 
   elasticsearch.http-hosts: [
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.clio.server.dataaccess
 
+import java.nio.ByteBuffer
 import java.nio.file.{Files, Path}
 
 import akka.NotUsed
@@ -8,7 +9,7 @@ import akka.stream.alpakka.file.scaladsl.Directory
 import akka.stream.scaladsl.{Sink, Source}
 import com.typesafe.scalalogging.LazyLogging
 import io.circe.Decoder
-import io.circe.parser._
+import io.circe.jawn.JawnParser
 import org.broadinstitute.clio.server.ClioServerConfig.Persistence
 import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
   ClioDocument,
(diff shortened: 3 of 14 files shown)


clio-integration-test/src/it/resources/org/broadinstitute/clio/integrationtest/clio-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
   clio-server:
     image: broadinstitute/clio-server:${CLIO_DOCKER_TAG}
     environment:
-      JAVA_OPTS: -Xmx128m -Xms128m -Dclio.server.elasticsearch.http-hosts.0.hostname=${ELASTICSEARCH_HOST} -Dlogback.configurationFile=/app/clio-logback.xml -Dclio.server.persistence.root-dir=/persistence
+      JAVA_OPTS: -Xmx128m -Xms128m -Dclio.server.persistence.recovery-parallelism=4 -Dclio.server.elasticsearch.http-hosts.0.hostname=${ELASTICSEARCH_HOST} -Dlogback.configurationFile=/app/clio-logback.xml -Dclio.server.persistence.root-dir=/persistence
```
danxmoran, line 8: Completely arbitrary choice of 4 here.

clio-server/config/clio.conf.ctmpl
```diff
@@ -10,6 +10,7 @@ clio.server {
     project-id: broad-gotc-{{env "ENV"}}-storage
     bucket: broad-gotc-{{env "ENV"}}-clio
     service-account-json: {{env "CLIO_CONF_DIR"}}/{{env "CLIO_SERVICE_ACCOUNT_JSON"}}
+    recovery-parallelism: 8
```
danxmoran, line 13: Equally-arbitrary choice of 8 here.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -38,13 +37,13 @@ trait PersistenceDAO extends LazyLogging {
...
-    indexes: Seq[ElasticsearchIndex[_]]
+    indexes: Seq[ElasticsearchIndex[_]],
+    version: String
   )(implicit ec: ExecutionContext): Future[Unit] = Future {
     // Make sure the rootPath can actually be used.
     val versionPath = rootPath.resolve(VersionFileName)
     try {
-      val versionFile =
-        Files.write(versionPath, ClioServerConfig.Version.value.getBytes)
+      val versionFile = Files.write(versionPath, version.getBytes)
```
danxmoran, line 49: Also continuing my personal quest to isolate references to `ClioServerConfig` to the server initializer / `apply` methods in companion objects, to improve mockability during testing.

tbl3rd (approved):
:sigh:

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -124,7 +123,7 @@ trait PersistenceDAO extends LazyLogging {
        * We use `mapAsync` then `mapConcat`, instead of just `flatMapConcat`,
        * so we can sort the intermediate Seq result of the ls.
        */
-      .mapAsync(1)(Directory.ls(_).runWith(Sink.seq))
+      .mapAsync(recoveryParallelism)(Directory.ls(_).runWith(Sink.seq))
```
pshapiro4broad, line 129: I didn't think that this operation was implicated in the profiling results. Or were we not able to distinguish it from the `Files.readAllBytes()` below?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
pshapiro4broad: This doesn't seem that much better to me than making the test mock the config to provide the setting needed for a test.

In the case of `version` it makes sense that it would be provided by the caller, since it's not a setting specific to this class. In the case of `recoveryParallelism`, I think it would make more sense to have the class read this setting itself. It only has one option now, but if it had many more you wouldn't want to have to keep updating the caller to handle each different setting.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/FailingPersistenceDAO.scala
```diff
@@ -9,7 +9,7 @@ import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class FailingPersistenceDAO extends PersistenceDAO {
+class FailingPersistenceDAO extends PersistenceDAO(recoveryParallelism = 1) {
```
pshapiro4broad: any way to make this an optional and avoid having to provide it to get the no-op behavior?

pshapiro4broad (approved):
A few minor comments

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
danxmoran: Yeah, they both hit the network. `Files.readAllBytes` is the one I'd expect the most benefit from parallelizing, but we should be able to squeeze out some wait time by doing the same thing here.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
danxmoran: > This doesn't seem that much better to me than making the test mock the config to provide the setting needed for a test.

I agree, but in the current setup it's not possible to do this. Since `ClioServerConfig` is an `object` it can't be subclassed. The scalamock library used to have a way of monkey-patching `object`s, but they dropped the functionality in their latest versions.

My ideal would be for all configs to be `case class`es, with each component that needs some config value defining their own `TypesafeConfig -> CaseClassConfig` method. I think that's too big of a change for this PR, but I can open a ticket for it.

PR #210, opened 2018-02-12
Fixes / cleanups / more tests for circe utils

danxmoran:
### Description

While poking around with using `Json` instead of `Document` types in Clio's core, I ran across some bugs / inconsistencies.

----

* Review cycle:


clio-util/src/main/scala/org/broadinstitute/clio/util/json/ModelAutoDerivation.scala
```diff
@@ -41,7 +41,7 @@ trait ModelAutoDerivation extends AutoDerivation {
     * By default, compact JSON as much as possible by removing spaces and keys with null / None values.
     */
   implicit val defaultPrinter: Printer =
-    Printer.noSpaces.copy(dropNullValues = true)
+    Printer.noSpaces.copy(dropNullValues = true, reuseWriters = true)
```
danxmoran: Stumbled across this optimization option. According to [this PR](https://github.com/circe/circe/pull/657) it has a significant affect on throughput / allocations, but it's not enabled by default because of some potential problem with Java's `ThreadLocal`... not sure how to tell if we're at risk.

clio-util/src/main/scala/org/broadinstitute/clio/util/json/StateDecoder.scala
```diff
@@ -78,7 +78,7 @@ object StateDecoder extends LowPriorityStateDecoder {
 
     field.as[Option[A]] match {
       // Found the field. Delete it, and return the modified cursor.
-      case Right(a @ Some(_)) => Right((field.delete, a))
+      case Right(a) if field.focus.isDefined => Right((field.delete, a))
```
danxmoran, line 81: When decoding objects like `{"field": null}`, doing `field.as[Option[_]]` will return `Right(None)`, but we still want to call `field.delete`.

clio-util/src/test/scala/org/broadinstitute/clio/util/json/StateDecoderSpec.scala
```diff
@@ -40,8 +40,16 @@ class StateDecoderSpec
     }
   }
 
+  it should "decode a field with an option explicitly set to null" in {
+    implicit val decoder: Decoder[TestClass] =
+      Decoder.fromState(StateDecoder[TestClass].state)
+
+    decode[TestClass]("""{"fieldA": null, "fieldB": 123}""").right.value should be(
```
danxmoran, line 47: Without the fix in the `StateDecoder`, this would raise an error saying `fieldA` was an extra key.

pshapiro4broad (approved):
Looks 👍 as far as I can tell

clio-util/src/main/scala/org/broadinstitute/clio/util/json/ModelAutoDerivation.scala
tbl3rd: This makes me nervous unless we know we'll notice any corruption of the output.
(Es will probably notice bad JSON, but maybe not missing fields or objects.)

tbl3rd (approved):
... but reuseWriters makes me nervous. I think we should avoid risk until the new Clio deployment is in place. The PR claims it doubles throughput and reduces allocation by 90%. If the throughput depends on allocation this change probably won't buy us much.

jacarey (approved):

PR #211, opened 2018-02-12
DSDEGP-2069 Don't do the JSON -> Document -> JSON dance during recovery

danxmoran:
### Description

This includes changes from #210, so there will be some duplication until that one merges.

----

* Review cycle:


clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
```diff
@@ -62,18 +63,18 @@ class HttpElasticsearchDAO private[dataaccess] (
       .query(queryDefinition)
 
     val responsePublisher = httpClient.publisher(searchDefinition)
-    Source.fromPublisher(responsePublisher).map(_.to[D])
+    Source.fromPublisher(responsePublisher).map(_.to[Json].as[D].fold(throw _, identity))
```
danxmoran: `.to[D]` -> `.to[Json].as[D]` needed because we no longer store an implicit `HitReader[D]` in the `ElasticsearchIndex`. The goal would be to just to `.to[Json]`, but that starts bleeding into other APIs so I didn't go through with it.

jacarey:
How much more work will it be to remove all json -> doc -> json ?

danxmoran:
Unsure... the SPIKE-y work will be in changing the `ElasticsearchIndex`'s mapping-generator code to work over key / metadata type-pairs instead of just one document type, and in making the document / query mappers work with `Json`. I think it'd probably be enough to block the ID fix release for a few more days.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -226,13 +234,13 @@ abstract class PersistenceDAO(recoveryParallelism: Int) extends LazyLogging {
         .fold[Option[Path]](None)((_, p) => Some(p))
         .flatMapConcat { maybePathToLast =>
           maybePathToLast.fold(
-            Source.failed[D](
+            Source.failed[Json](
               new NoSuchElementException(
-                s"No document found in storage for ID ${document.upsertId}"
+                s"No document found in storage for ID ${id.id}"
```
jacarey: id.id is a little awkward.. maybe upsertId.id?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SearchDAO.scala
```diff
@@ -39,8 +41,12 @@ trait SearchDAO {
     * @param index    The index in which to update the document.
     * @tparam D The type of the document.
     */
-  def updateMetadata[D <: ClioDocument](document: D)(
+  final def updateMetadata[D <: ClioDocument](document: D)(
     implicit index: ElasticsearchIndex[D]
+  ): Future[Unit] = updateMetadata(document.asJson(index.encoder))
```
jacarey, line 46: sometimes we import index.decoder or encoder and sometimes it is being passed in as implicit param. I'm curious why we do it differently at times?

jacarey (approved):
look good :+1:

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SearchDAO.scala
danxmoran: My idea was to import if it's used in > 1 spot, and otherwise just pass it explicitly, but it looks like all usages are just for 1 spot. I'll make it consistent.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -73,19 +76,26 @@ abstract class PersistenceDAO(recoveryParallelism: Int) extends LazyLogging {
...
+    val jsonString = ModelAutoDerivation.defaultPrinter.pretty(document.asJson)
+
+    val writePath =
+      Files.createDirectories(rootPath.resolve(index.persistenceDirForDatetime(dt)))
+
+    val written = Files.write(
+      writePath.resolve(ClioDocument.persistenceFilename(document.upsertId)),
+      jsonString.getBytes
+    )
+    logger.debug(s"Wrote document $document to ${written.toUri}")
```
pshapiro4broad, line 96: Not sure what `toUri()` buys you here. Is it a shorter way to ensure you print the absolute path than writing `toAbsolutePath()`?

danxmoran: `toUri()` includes the `gs://bucket-name` prefix in its stringified form, and (I'm pretty sure) `toAbsolutePath` doesn't.

pshapiro4broad (approved):

pshapiro4broad:
That’s true, but we don’t have the GCS streaming library imported, so we can’t make gs:// Paths in the first place.

E.g.

Path p = Paths.get(URI.create("gs://some/gcs/path.bam"));

Will fail with an exception.

danxmoran:
We do have the gcloud NIO adapter loaded into the clio-server... that's how the abstracted recovery logic works at all.

PR #212, opened 2018-02-12
DSDEGP-2068 Collect ES updates into batches to relieve recovery backpressure.

danxmoran:
### Description

@jacarey noticed from repeated profiling that the `recovery-parallelism` setting had steeply diminishing returns, because the whole stream was getting backpressured waiting on the sequential
updates to Elasticsearch. This PR tweaks the end stage of the stream to collect updates into batches, which are sent as bulk updates to ES.

The change halves the runtime of the `RecoveryIntegrationSpec` when I run it locally 🎉

----

* Review cycle:

### DIFF
--- a/clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentMock.scala
+++ b/clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentMock.scala
@@ -6,6 +6,8 @@ import java.time.OffsetDateTime
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
 import org.broadinstitute.clio.util.model.{DocumentStatus, UpsertId}
 
+import scala.util.Random
+
 case class DocumentMock(
   upsertId: UpsertId,
   entityId: Symbol,
@@ -24,17 +26,21 @@ case class DocumentMock(
 
 object DocumentMock extends ModelAutoDerivation {
 
-  def default: DocumentMock = DocumentMock(
-    upsertId = UpsertId.nextId(),
-    entityId = Symbol("1234.the-string"),
-    mockKeyLong = 1234L,
-    mockKeyString = "the-string",
-    mockFileMd5 = Some(Symbol("12345abcdefg")),
-    mockFilePath = Some(URI.create("gs://the-bucket/the-path.file")),
-    mockStringArray = Some(Seq.empty),
-    mockPathArray = Some(Seq.empty),
-    mockDocumentStatus = Some(DocumentStatus.Normal)
-  )
+  def default: DocumentMock = {
+    val long = Random.nextLong()
+    val string = Random.nextString(10)
+    DocumentMock(
+      upsertId = UpsertId.nextId(),
+      entityId = Symbol(s"$long.$string"),
+      mockKeyLong = long,
+      mockKeyString = string,
+      mockFileMd5 = Some(Symbol("12345abcdefg")),
+      mockFilePath = Some(URI.create("gs://the-bucket/the-path.file")),
+      mockStringArray = Some(Seq.empty),
+      mockPathArray = Some(Seq.empty),
+      mockDocumentStatus = Some(DocumentStatus.Normal)
+    )
+  }
 
   implicit val index: ElasticsearchIndex[DocumentMock] =
     new ElasticsearchIndex[DocumentMock](
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
@@ -3,38 +3,38 @@ package org.broadinstitute.clio.integrationtest
 import java.io.File
 import java.net.URI
 
-import akka.Done
 import akka.http.scaladsl.model.StatusCodes
 import akka.stream.scaladsl.Sink
+import io.circe.Json
+import io.circe.syntax._
 import org.broadinstitute.clio.client.commands.ClioCommand
 import org.broadinstitute.clio.client.webclient.ClioWebClient.FailedResponse
-import org.broadinstitute.clio.server.dataaccess.elasticsearch.{
-  DocumentGvcf,
-  DocumentWgsCram,
-  DocumentWgsUbam,
-  ElasticsearchIndex
-}
+import org.broadinstitute.clio.server.dataaccess.elasticsearch._
 import org.broadinstitute.clio.status.model.{ClioStatus, StatusInfo, VersionInfo}
-import org.broadinstitute.clio.transfer.model.gvcf.TransferGvcfV1QueryOutput
-import org.broadinstitute.clio.transfer.model.ubam.{
-  TransferUbamV1Metadata,
-  TransferUbamV1QueryOutput
-}
-import org.broadinstitute.clio.transfer.model.wgscram.TransferWgsCramV1QueryOutput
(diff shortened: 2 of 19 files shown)


clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -64,9 +64,9 @@ class ServerService private (
 
       persistenceDAO
         .getAllSince(latestUpsert)
-        .runWith(Sink.foldAsync(0) { (count, json) =>
-          logger.debug(s"Recovering document with ID ${getUpsertId(json)}")
```
danxmoran, line 68: I deleted this log because elastic4s already logs out the full JSON body of the POST when set to debug logging, so this doesn't buy us much.

pshapiro4broad (approved):

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -64,9 +64,9 @@ class ServerService private (
...
-          searchDAO.updateMetadata(json).map(_ => count + 1)
+        .conflateWithSeed(Vector(_))(_ :+ _)
```
jacarey: Will conflate cause variable size batches to be sent to update?  Do we want to set up something that will send fixed size batches ? Or at least a max batch size?

jacarey (commented):
Just one question about batch size, but otherwise LGTM.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
danxmoran: Right now it'll be variable (it'll pull whatever `Vector` has managed to build up since the last pull). I'll look into setting a max size, also playing around with merging JSON in-memory if multiple records build up for the same ID before the ES pull.

danxmoran:
I tried to beef up the recovery integration test (right now it only loads & recovers 50 of each document...), and hit weirdness where some records fail to be updated. Looking into it now.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
jacarey: I think you can use sliding or batch to limit the total size. I think sliding would keep the batch size constant (aside from potentially the last batch after the stream is done).

jacarey (approved):

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -156,10 +169,10 @@ class RecoveryIntegrationSpec
...
       )
     } yield {
-      ubams should have length documentCount.toLong
-      ubams.map(_.ubamPath) should contain theSameElementsAs storedUbams.map(
-        _.ubamPath
-      )
+      val ubamUris = ubams.flatMap(_.ubamPath).sorted
+      ubamUris should contain theSameElementsInOrderAs updatedUbams
+        .flatMap(_.ubamPath)
+        .sorted
```
danxmoran: Sorting here to make it easier to diff the actual vs. expected

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -137,6 +156,12 @@ class ServerService private (
 
 object ServerService {
 
+  /**
+    * Maximum number of updates that should be sent in a single bulk request
+    * to Elasticsearch during document recovery.
+    */
+  val RecoveryMaxBulkSize: Long = 100L
```
danxmoran: We should probably up this number once we figure out why the integration tests aren't working

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -24,63 +24,76 @@ import org.broadinstitute.clio.transfer.model.wgscram.TransferWgsCramV1QueryOutp
...
 
 import scala.concurrent.Future
-import scala.util.Random
 
 /** Tests for recovering documents on startup. Can only run reproducibly in Docker. */
 class RecoveryIntegrationSpec
     extends DockerIntegrationSpec("Clio in recovery", "Recovering metadata") {
 
-  val documentCount = 50
+  val documentCount = 500
```
danxmoran: Once we fix the tests as they are now, would love to push this number up to 5K+

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -55,26 +54,46 @@ class ServerService private (
...
+            val newJson = idMap.get(id) match {
+              case Some(oldJson) => {
+                logger.debug(
+                  s"Merging upserts ${getUpsertId(oldJson)} and ${getUpsertId(json)} for id $id"
+                )
+                oldJson.deepMerge(json)
+              }
+              case None => json
+            }
+            idMap.updated(id, newJson)
```
danxmoran, line 80: New logic here to:
1. Enforce a max batch size
2. Merge upserts in-memory when possible

The default max-batch-size enforcer isn't smart enough to notice when two records merge, but there's a "cost function" parameter we could tweak so that it doesn't increment the current-total-batch-size when we merge entries in the map instead of adding a new entry.

clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/DocumentMock.scala
```diff
@@ -24,17 +26,21 @@ case class DocumentMock(
...
-    mockStringArray = Some(Seq.empty),
-    mockPathArray = Some(Seq.empty),
-    mockDocumentStatus = Some(DocumentStatus.Normal)
-  )
+  def default: DocumentMock = {
+    val long = Random.nextLong()
+    val string = Random.nextString(10)
+    DocumentMock(
+      upsertId = UpsertId.nextId(),
+      entityId = Symbol(s"$long.$string"),
```
danxmoran, line 34: Updated with randomness so we wouldn't get unintentional merging in our existing recovery unit tests.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
danxmoran: No luck with the cost function, it's only allowed to take the next element (vs. the next element and the existing batch aggregation). I don't think there's any harm in being conservative with the batch size.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -55,26 +54,46 @@ class ServerService private (
...
         }
+        .runWith(Sink.foldAsync(()) { (_, jsonMap) =>
+          val jsons = jsonMap.valuesIterator.toVector
+          searchDAO.updateMetadata(jsons: _*).map { _ =>
+            logger.debug(s"Sent ${jsons.size} upserts to Elasticsearch for index $name")
```
jacarey, line 85: I'll be curious to see how consistent the batch size is when we run in on dev. Can we be sure to output debug logs when we test it?

jacarey (approved):

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
danxmoran: Yup!

PR #213, opened 2018-02-14
DSDEGP-2058 Throw error when attempting to add a document that exists…

jacarey:
… and we are changing a metadata value

### Description

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+  ): Map[String, Any] = {
+    val storedFields =
+      new CaseClassMapper[addCommand.index.MetadataType].vals(existingMetadata)
+    val upsertFields =
+      new CaseClassMapper[addCommand.index.MetadataType].vals(newMetadata)
+
+    (upsertFields.toSet diff storedFields.toSet).toMap
+  }
+
+  private def queryForKey(
```
danxmoran: We've got variations of this logic in the add, move, and delete CLPs... could we take the chance to move it into the web-client? Or better to split off as a tech-debt ticket?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
       Future
         .failed(_) logErrorMsg s"Metadata at $location cannot be added to Clio", {
         decoded =>
-          webClient.upsert(addCommand.index)(addCommand.key, decoded)
+          {
+            for {
+              existingMetadata <- queryForKey(webClient).logErrorMsg(
+                s"Could not query the $name. No files have been deleted."
```
danxmoran: Copy-paste from the delete CLP, needs updated.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
       Future
         .failed(_) logErrorMsg s"Metadata at $location cannot be added to Clio", {
         decoded =>
-          webClient.upsert(addCommand.index)(addCommand.key, decoded)
+          {
+            for {
+              existingMetadata <- queryForKey(webClient).logErrorMsg(
```
danxmoran: Do we even want to do this query if `forceUpdate` is `true`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+                s"Could not query the $name. No files have been deleted."
+              )
+              upsertResponse <- addFiles(
+                webClient,
+                decoded,
+                existingMetadata,
+                addCommand.forceUpdate
+              ).fold(
+                left => Future.failed(left) logErrorMsg left.getMessage,
+                right => right
```
danxmoran: `right => right` could be `identity`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+              existingMetadata <- queryForKey(webClient).logErrorMsg(
+                s"Could not query the $name. No files have been deleted."
+              )
+              upsertResponse <- addFiles(
+                webClient,
+                decoded,
+                existingMetadata,
+                addCommand.forceUpdate
+              ).fold(
+                left => Future.failed(left) logErrorMsg left.getMessage,
```
danxmoran: Nitpick: we don't use infix very much

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
     )
   }
+
+  private def addFiles(
+    client: ClioWebClient,
+    newMetadata: addCommand.index.MetadataType,
+    existingMetadata: Option[addCommand.index.MetadataType],
+    forceUpdate: Boolean
+  )(implicit ec: ExecutionContext): Either[Throwable, Future[UpsertId]] = {
+    val map = existingMetadata.map(diffMetadata(_, newMetadata)).getOrElse(Map.empty)
```
danxmoran: `.map(...).getOrElse(Map.empty)` would be `.fold(Map.empty)(...)`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+    for {
+      upsertResponse <- Either.cond(
+        map.isEmpty || forceUpdate,
+        client
+          .upsert(addCommand.index)(addCommand.key, newMetadata)
+          .logErrorMsg(
+            s"""An error occurred while adding the $name record in Clio.""".stripMargin
```
danxmoran: Leftover `"""` and `stripMargin` from something else?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+          ),
+        new RuntimeException(
+          s"Adding this document that will overwrite the following existing metadata: " +
```
danxmoran: `s` not needed if we're `+`-ing the pieces together.

Remove `that`, for better wording?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+            map.keys.mkString(", ") +
```
danxmoran: Hrm, would it be valuable to include the existing values on top of the keys? Also this seems like it could build up a huge line, would it make sense to join with `"\n"` instead of `", "`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,106 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+            s" Use the '--force-update true' to overwrite this data."
```
danxmoran: `s` also not needed here.

Case-app shouldn't require the explicit `true` for a boolean option w/ a default value.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AddGvcfSpec.scala
```diff
@@ -47,4 +48,21 @@ class AddGvcfSpec extends BaseClientSpec {
       .map(_ shouldBe an[UpsertId])
   }
 
+  it should "fail to add a gvcf that would overwrite an existing document" in {
+    recoverToSucceededIf[Exception] {
+      val mockIoUtil = new MockIoUtil
+
+      succeedingDispatcher(mockIoUtil, testGvcfLocation)
+        .dispatch(goodGvcfAddCommand)
+        .map(_ shouldBe an[UpsertId])
```
danxmoran: I'm missing something here... have we already set up the `succeedingDispatcher` to return data for the key in `goodGvcfAddCommand`?

clio-client/src/test/scala/org/broadinstitute/clio/client/util/TestData.scala
```diff
@@ -174,6 +178,13 @@ trait TestData {
   val goodAddCommand =
     AddWgsUbam(metadataLocation = metadataFileLocation, key = testTransferV1Key)
 
+  val goodAddCommandForceUpdate =
+    AddWgsUbam(
+      metadataLocation = metadataFileLocation,
+      key = testTransferV1Key,
+      forceUpdate = true
+    )
```
danxmoran: (I think) this could be `goodAddCommand.copy(forceUpdate = true)`

clio-client/src/test/scala/org/broadinstitute/clio/client/util/TestData.scala
```diff
@@ -189,6 +200,12 @@ trait TestData {
     key = testGvcfTransferV1Key
   )
 
+  val goodGvcfAddCommandForceUpdate = AddGvcf(
+    metadataLocation = gvcfMetadataFileLocation,
+    key = testGvcfTransferV1Key,
+    forceUpdate = true
+  )
```
danxmoran: Same here.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AddGvcfSpec.scala
danxmoran: I'd also remove the `.map(_ shouldBe ...)`, since it's not an assertion that we expect to run / that we care about for this test.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AddWgsUbamSpec.scala
```diff
@@ -47,4 +48,21 @@ class AddWgsUbamSpec extends BaseClientSpec {
       .map(_ shouldBe an[UpsertId])
   }
 
+  it should "fail to add a ubam that would overwrite an existing document" in {
+    recoverToSucceededIf[Exception] {
+      val mockIoUtil = new MockIoUtil
+      mockIoUtil.putFileInCloud(testUbamCloudSourcePath)
+      succeedingDispatcher(mockIoUtil, testWgsUbamLocation)
+        .dispatch(goodAddCommand)
+        .map(_ shouldBe an[UpsertId])
```
danxmoran: Same questions from the gvcf spec.

danxmoran (commented):
Do the integration tests need updated? I feel like we must have at least one that overwrites existing metadata.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
jacarey: Yeah I started doing this but it ended up being a little more work and seemed outside the scope of the ticket. I'll take another stab at it.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AddGvcfSpec.scala
jacarey: I think the succeeding dispatcher just returns the json that you set.. so it isn't a real key lookup. We could have succeedingDispatcher add the same data and then change it and run it again.. but I think it ends up testing the same thing essentially.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
jacarey: Yeah I debated this. Unfortunately the diff returns the fieldNames and values of the newMetadata. I can run the diff in both directions though and tie them together with the field name. I'll give it a shot.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,118 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+            s"An error occurred while adding the $name record in Clio."
+          ),
+        new RuntimeException(
+          "Adding this document will overwrite the following existing metadata: " +
+            differences
+              .map(
+                diff => s"Field: ${diff._1} Old value: ${diff._3} New value: ${diff._2}"
+              )
+              .mkString("\n") +
+            " Use the '--force-update' to overwrite this data."
```
als364: Wording is a bit awkward here, maybe "Use the '--force-update' flag" or "Use '--force-update'"?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -47,7 +47,8 @@ trait GvcfTests { self: BaseIntegrationSpec =>
       "--version",
       key.version.toString,
       "--metadata-location",
-      tmpMetadata.toString
+      tmpMetadata.toString,
+      "--force-update"
```
danxmoran: Can we parameterize the `runUpsertXXX` methods, so we can write int tests for upsert-ing without the force flag?

als364 (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,118 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+    existingMetadata: addCommand.index.MetadataType,
+    newMetadata: addCommand.index.MetadataType
+  ): Iterable[(String, Any, Any)] = {
+    val storedFields =
+      new CaseClassMapper[addCommand.index.MetadataType].vals(existingMetadata)
+    val upsertFields =
+      new CaseClassMapper[addCommand.index.MetadataType].vals(newMetadata)
+
+    val newToOldDiff = upsertFields.toSet.diff(storedFields.toSet).toMap
+    val oldToNewDiff = storedFields.toSet.diff(upsertFields.toSet).toMap
```
danxmoran: I'm not sure this'll do what we want... won't it flag fields-only-in-the-server and fields-only-in-new-metadata as errors, instead of just fields-in-both?

Could we do a single set-intersection on the keys, and then diff the values?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,118 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+
+    for (key <- newToOldDiff.keys ++ oldToNewDiff.keys)
+      yield (key, newToOldDiff.getOrElse(key, None), oldToNewDiff.getOrElse(key, None))
+  }
+
+  private def queryForKey(
```
danxmoran: Not sure of what we decided on this, are you trying to refactor into the web client, or pushing it to later?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -47,7 +48,8 @@ trait GvcfTests { self: BaseIntegrationSpec =>
       "--version",
       key.version.toString,
       "--metadata-location",
-      tmpMetadata.toString
+      tmpMetadata.toString,
+      if(forceUpdate)"--force-update" else ""
```
danxmoran: Can you add a new integration test that tests upserting without the force? Covering:
1. Fresh upsert
2. Filling in missing keys in an existing record
3. Failing to overwrite existing keys with a new value

I don't trust the client unit tests.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
jacarey: the nested types make this more of a pain than it seems. Probably not going to tackle it now.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
jacarey: This is getting the vals and comparing them.. the variables are poorly named.

The only reason we run the new -> old then old -> new diff is so we get values for both old and new to output in the message. But the results are the same.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
jacarey: Agreed. Client tests seem sorely lacking.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,142 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+                )
+              } else {
+                Future.successful(None)
+              }
+              upsertResponse <- addFiles(
+                webClient,
+                decoded,
+                existingMetadata
+              ).fold(
+                left => Future.failed(left).logErrorMsg(left.getMessage),
```
danxmoran: Is the `logErrorMsg` redundant? It'll wrap `left` with a new `RuntimeException`, with the same message, so we'll see it twice in the ending stack trace.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -42,8 +46,142 @@ class AddExecutor[TI <: TransferIndex](addCommand: AddCommand[TI])
...
+  }
+
+  private def diffMetadata(
+    existingMetadata: addCommand.index.MetadataType,
+    newMetadata: addCommand.index.MetadataType
+  ): Iterable[(String, Any, Any)] = {
+    val mapper = new CaseClassMapper[addCommand.index.MetadataType]
+
+    def filterNones(value: (String, Any)) = {
+      value._2 match {
```
danxmoran: Does `!value._2.equals(None)` work? If so we can probably just inline it into the `filter` calls.

danxmoran (approved):
Some last ideas for cleanup, LGTM either way!

PR #216, opened 2018-02-16
Jc dsdegp 2051 error on delete non existent

jacarey:
DSDEGP-2051 we want to error instead of warn when trying to delete a file that doesn't exist on the cloud. This can be overriden with the --force-delete flag.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
@@ -57,6 +57,7 @@ sealed abstract class DeleteCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
   def note: String
+  def forceDelete: Boolean
 }
 
 // Generic commands.
@@ -93,8 +94,11 @@ final case class MoveWgsUbam(
 ) extends MoveCommand(WgsUbamIndex)
 
 @CommandName(ClioCommand.deleteWgsUbamName)
-final case class DeleteWgsUbam(@Recurse key: TransferUbamV1Key, note: String)
-    extends DeleteCommand(WgsUbamIndex)
+final case class DeleteWgsUbam(
+  @Recurse key: TransferUbamV1Key,
+  note: String,
+  forceDelete: Boolean = false
+) extends DeleteCommand(WgsUbamIndex)
 
 // GVCF commands.
 
@@ -122,8 +126,11 @@ final case class MoveGvcf(
 ) extends MoveCommand(GvcfIndex)
 
 @CommandName(ClioCommand.deleteGvcfName)
-final case class DeleteGvcf(@Recurse key: TransferGvcfV1Key, note: String)
-    extends DeleteCommand(GvcfIndex)
+final case class DeleteGvcf(
+  @Recurse key: TransferGvcfV1Key,
+  note: String,
+  forceDelete: Boolean = false
+) extends DeleteCommand(GvcfIndex)
 
 // WGS-cram commands.
 
@@ -151,8 +158,11 @@ final case class MoveWgsCram(
 ) extends MoveCommand(WgsCramIndex)
 
 @CommandName(ClioCommand.deleteWgsCramName)
-final case class DeleteWgsCram(@Recurse key: TransferWgsCramV1Key, note: String)
-    extends DeleteCommand(WgsCramIndex)
+final case class DeleteWgsCram(
+  @Recurse key: TransferWgsCramV1Key,
+  note: String,
+  forceDelete: Boolean = false
+) extends DeleteCommand(WgsCramIndex)
 
 @CommandName(ClioCommand.deliverWgsCramName)
 final case class DeliverWgsCram(
@@ -188,8 +198,11 @@ final case class MoveHybselUbam(
 ) extends MoveCommand(HybselUbamIndex)
 
 @CommandName(ClioCommand.deleteHybselUbamName)
-final case class DeleteHybselUbam(@Recurse key: TransferUbamV1Key, note: String)
-    extends DeleteCommand(HybselUbamIndex)
+final case class DeleteHybselUbam(
+  @Recurse key: TransferUbamV1Key,
+  note: String,
+  forceDelete: Boolean = false
+) extends DeleteCommand(HybselUbamIndex)
 
 object ClioCommand extends ClioParsers {
 
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
@@ -37,7 +37,7 @@ sealed abstract class GetSchemaCommand[TI <: TransferIndex](val index: TI)
 sealed abstract class AddCommand[TI <: TransferIndex](val index: TI) extends ClioCommand {
   def key: index.KeyType
   def metadataLocation: URI
-  def forceUpdate: Boolean
+  def force: Boolean
 }
 
 sealed abstract class QueryCommand[TI <: TransferIndex](val index: TI)
@@ -57,7 +57,7 @@ sealed abstract class DeleteCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
   def note: String
-  def forceDelete: Boolean
+  def force: Boolean
 }
 
 // Generic commands.
@@ -77,7 +77,7 @@ case object GetSchemaWgsUbam extends GetSchemaCommand(WgsUbamIndex)
 final case class AddWgsUbam(
   @Recurse key: TransferUbamV1Key,
   metadataLocation: URI,
-  forceUpdate: Boolean = false
+  force: Boolean = false
 ) extends AddCommand(WgsUbamIndex)
 
 @CommandName(ClioCommand.queryWgsUbamName)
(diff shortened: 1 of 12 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -57,6 +57,7 @@ sealed abstract class DeleteCommand[TI <: TransferIndex](val index: TI)
     extends ClioCommand {
   def key: index.KeyType
   def note: String
+  def forceDelete: Boolean
```
danxmoran: Bikeshedding: since the command will already be named something like `delete-xyz`, do we need `--force-delete` instead of just `--force`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -103,10 +103,14 @@ class DeleteExecutor[TI <: TransferIndex](deleteCommand: DeleteCommand[TI])
   )(implicit ec: ExecutionContext): Future[UpsertId] = {
     val pathsToDelete = existingMetadata.pathsToDelete.filter { path =>
       val pathExists = ioUtil.googleObjectExists(path)
-      if (!pathExists) {
-        logger.warn(
-          s"'$path' associated with $prettyKey does not exist in the cloud."
-        )
+      val err = s"'$path' associated with $prettyKey does not exist in the cloud."
```
danxmoran: (As seen in the nightly tests) `ioUtil.googleObjectExists` could return false if `gsutil stat` fails for other reasons, so this message won't always be accurate... I'm not sure if we can/should reliably distinguish `gsutil` return codes to fix `googleObjectExists` or if we should just make this message more generic.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -103,10 +103,14 @@ class DeleteExecutor[TI <: TransferIndex](deleteCommand: DeleteCommand[TI])
...
+      (pathExists, deleteCommand.forceDelete) match {
+        case (false, false) =>
+          logger.error(err + " Use --force-delete to delete this file.")
```
danxmoran: This is redundant since the rolled-up exception will get re-logged at the top level. I'd just throw the `err + " Use ..."`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -103,10 +103,14 @@ class DeleteExecutor[TI <: TransferIndex](deleteCommand: DeleteCommand[TI])
...
+          throw new Exception(err)
```
danxmoran: I've been trying to use more specific exceptions lately... would `IllegalStateException` make sense?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -103,10 +103,14 @@ class DeleteExecutor[TI <: TransferIndex](deleteCommand: DeleteCommand[TI])
...
+        case (false, true) =>
+          logger.warn(err)
+        case _ =>
```
danxmoran: I'd do nested `if`s instead of the `match`, to avoid the awkward no-op catch-all.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/TransferWgsCramV1Metadata.scala
```diff
@@ -60,14 +60,15 @@ case class TransferWgsCramV1Metadata(
...
+  // Delete the cramPath.md5 file only if a workspaceName is defined otherwise there will be no md5
+  // (foo.cram.md5 where foo.cram is cramPath).
   override def pathsToDelete: Seq[URI] =
     Seq.concat(
       cramPath,
       craiPath,
-      cramPath.map(
-        cp => URI.create(s"$cp${WgsCramExtensions.Md5ExtensionAddition}")
-      )
+      if (workspaceName.isDefined) cramPath.map { cp =>
```
danxmoran: Could be `workspaceName.flatMap(_ => cramPath.map(...))`

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
jacarey: Should we change the --force-update on add as well for consistency?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
jacarey: Does gsutil return something other than just 1? The docs seem to indicate that it is 1 or 0

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
danxmoran: I'd say 👍

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -103,10 +103,14 @@ class DeleteExecutor[TI <: TransferIndex](deleteCommand: DeleteCommand[TI])
...
     val pathsToDelete = existingMetadata.pathsToDelete.filter { path =>
       val pathExists = ioUtil.googleObjectExists(path)
+      val err = s"'$path' associated with $prettyKey does not exist in the cloud."
       if (!pathExists) {
-        logger.warn(
-          s"'$path' associated with $prettyKey does not exist in the cloud."
-        )
+        if (!deleteCommand.forceDelete) {
+          throw new IllegalStateException(
+            err + " Use --force-delete to delete this file."
```
danxmoran: Oops, missed it on first pass. `to delete this file` is confusing because the 1st half of the message says the file doesn't exist. `to mark the record as deleted in Clio` is more accurate.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/TransferWgsCramV1Metadata.scala
```diff
@@ -60,13 +60,17 @@ case class TransferWgsCramV1Metadata(
...
+  // Delete the cramPath.md5 file only if a workspaceName is defined otherwise there will be no md5
+  // (foo.cram.md5 where foo.cram is cramPath).
   override def pathsToDelete: Seq[URI] =
     Seq.concat(
       cramPath,
       craiPath,
-      cramPath.map(
-        cp => URI.create(s"$cp${WgsCramExtensions.Md5ExtensionAddition}")
+      workspaceName.flatMap(
+        _ =>
```
danxmoran: oh scalafmt

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: Probably just 1... I'll try to force an auth error and see if it spits out something else.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -644,20 +649,26 @@ trait GvcfTests {
     ).map {
       case (path, contents) => Files.write(path, contents.getBytes)
     }
+
+    if (testNonExistingFile) IoUtil.deleteGoogleObject(gvcfPath.toUri)
```
danxmoran, line 653: Seems more straightforward to not `Files.write` above at all if `testNonExistingFile` is true, rather than writing then deleting.

danxmoran (approved):
Last idea for cleanup, LGTM

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
jacarey: Yeah I think the only reason I didn't do this is because it write metrics files as well which it then checks later to make sure the are around. It might make more sense to split this off from testDeleteGvcf entirely

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/TransferWgsCramV1Metadata.scala
tbl3rd: I don't understand why this is here. (Oh, never mind. I get it.)

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
tbl3rd: On the other hand, this tests the actual scenario that led to the problem.

tbl3rd (approved):

PR #220, opened 2018-03-02
DSDEGP-2069 - Stop expensive json -> case class -> json conversions

als364:
### Description

This isn't ready for merging yet, but I'm opening this for comments in the hopes that people who are not @danxmoran can look at it.

I basically deleted ClioDocument and followed the compiler complaints. I also collapsed `AutoElasticsearchQueryMapper` and `AutoElasticsearchDocumentMapper` into their parent classes, since they were the only subclasses.

`ElasticsearchIndex` is now parameterized on `TransferIndex` and uses `TransferIndex`'s `KeyType` and `MetadataType` rather than using `TransferKey`s and `MetadataKey`s directly.

I haven't started looking at the tests, and there's still a `not enough arguments for method implicitly` compile error that I don't understand how to fix.

Ideally I can get rid of the encoders/decoders in `Elasticsearchindex` eventually and have it just deal with json.

----

* Review cycle:


clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -58,14 +65,17 @@ class ElasticsearchIndex[Document: FieldMapper](
 
   /** The fields for the index. */
   def fields: Seq[FieldDefinition] =
-    FieldMapper[Document].fields.toSeq.sortBy {
+    (FieldMapper[index.KeyType].fields.toSeq ++ FieldMapper[index.MetadataType].fields.toSeq).sortBy {
```
pshapiro4broad: Is there a way to combine the two `fields` before calling `toSeq()`?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -75,28 +85,31 @@ object ElasticsearchIndex extends ModelAutoDerivation {
...
-  def apply[Document: ElasticsearchIndex]: ElasticsearchIndex[Document] =
-    implicitly[ElasticsearchIndex[Document]]
+  def apply[Index]: ElasticsearchIndex[Index] =
+    implicitly[ElasticsearchIndex[Index]]
 
-  implicit val WgsUbam: ElasticsearchIndex[DocumentWgsUbam] =
-    new ElasticsearchIndex[DocumentWgsUbam](
+  implicit val WgsUbam: ElasticsearchIndex[WgsUbamIndex.type] =
+    new ElasticsearchIndex(
       "wgs-ubam",
```
pshapiro4broad: This is the same as `WgsUbamIndex.commandName`. Can this value be moved to a property of index? We could make `commandName` default to the index name, which seems like what you'd typically want.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -75,28 +85,31 @@ object ElasticsearchIndex extends ModelAutoDerivation {
...
+      WgsUbamIndex,
       ElasticsearchFieldMapper.StringsToTextFieldsWithSubKeywords
```
pshapiro4broad: If the mapper is the same for all, does it need to be provided as an argument?

build.sbt
```diff
@@ -92,6 +92,7 @@ lazy val `clio-transfer-model` = project
   */
 lazy val `clio-dataaccess-model` = project
   .dependsOn(`clio-util`)
+  .dependsOn(`clio-transfer-model`)
```
danxmoran: Nitpick: `.dependsOn` is varargs, so you can change the existing line to
```
.dependsOn(`clio-util`, `clio-transfer-model`)
```

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
```diff
@@ -63,15 +64,15 @@ class HttpElasticsearchDAO private[dataaccess] (
...
+      .map(_.to[Json])
   }
 
   override def getMostRecentDocument(
     implicit index: ElasticsearchIndex[_]
   ): Future[Option[Json]] = {
     val searchDefinition = searchWithType(index.indexName / index.indexType)
       .size(1)
-      .sortByFieldDesc(ClioDocument.UpsertIdElasticSearchName)
+      .sortByFieldDesc(ElasticsearchUtil.toElasticsearchName(UpsertId.UpsertIdFieldName))
```
pshapiro4broad: Both `toElasticSearchName()` and `UpsertIdFieldName` are static, so the value of this function call won't change, so it seems like it should be done once as a property of `UpsertId`.

clio-util/src/main/scala/org/broadinstitute/clio/util/model/EntityId.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.clio.util.model
+
+object EntityId {
+
+  /**
+    * Clio name for an entity ID.
+    */
+  val EntityIdFieldName: String = "entityId"
+}
```
danxmoran: Having a separate object just for this seems like overkill.

Looking ahead, I think it (and the method to get the entity ID out of a JSON blob) belongs in `ElasticsearchIndex`, because after [DSDEGP-2121](https://broadinstitute.atlassian.net/browse/DSDEGP-2121) the hard-coded `entity_id` field won't exist any more and we'll need to be able to vary our ID-calculation logic on an index-by-index basis.

danxmoran: We can also go straight to snake case for the field name now that it's not part of any case-classes.

clio-util/src/main/scala/org/broadinstitute/clio/util/model/UpsertId.scala
```diff
@@ -72,6 +79,11 @@ object UpsertId {
     */
   private val next6bitIndex = (_: Int) => source.nextInt(Encoding.length)
 
+  /**
+    * Clio name for an upsert ID.
+    */
+  val UpsertIdFieldName: String = "upsertId"
```
danxmoran: I think this'd make more sense to live next to the corresponding field for `entity_id` in `ElasticsearchIndex`, since it's really an artifact of how we've decided to keep notes in our indexes and not an inherent property of the `UpsertId` type.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,50 @@
...
   * Maps metadata to an Elasticsearch document.
   *
-  * @tparam ModelKey      The type of key.
-  * @tparam ModelMetadata The type of the metadata.
-  * @tparam Document      The type of the document.
+  * @tparam Key      The type of the TransferKey.
+  * @tparam Metadata The type of the TransferMetadata.
   */
-abstract class ElasticsearchDocumentMapper[ModelKey, ModelMetadata, Document] {
+class ElasticsearchDocumentMapper[Key <: Product: ClassTag, Metadata: ClassTag] (genId: () => UpsertId)(
```
danxmoran: By the end, we shouldn't need the `: ClassTag` bounds.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,50 @@
...
+  implicit
+  private[elasticsearch] val keyEncoder: Encoder[Key],
+  private[elasticsearch] val metadataEncoder: Encoder[Metadata]
```
danxmoran: Since these aren't used explicitly, then can be moved up into the type signature as context bounds. i.e.:
`Key <: Product: Encoder`

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,50 @@
...
+) {
+  private val keyMapper = new CaseClassMapper[Key]
+  private val metadataMapper = new CaseClassMapper[Metadata]
```
danxmoran: By the end, we shouldn't need these mappers.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,50 @@
...
-  def empty(key: ModelKey): Document
+  def document(key: Key, metadata: Metadata): Json = {
+    val keyVals = keyMapper.vals(key)
+    val metadataVals = metadataMapper.vals(metadata)
+    val bookkeeping = Map(
+      UpsertId.UpsertIdFieldName -> genId(),
+      EntityId.EntityIdFieldName -> Symbol(
+        key.productIterator.mkString(".")
+      )
+    )
```
danxmoran: These steps aren't needed, we can go straight to JSON: `key.asJson`.

`Json.fromFields` should help with building up the bookkeeping object.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
danxmoran: I'm not sure it'd work in general, because of the -v2 tacked on the end of the gvcf & cram indexes.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -75,28 +85,31 @@ object ElasticsearchIndex extends ModelAutoDerivation {
...
     // Since we compute GCS paths from the ES index name, inconsistency would break GCS paths.
-    new ElasticsearchIndex[DocumentGvcf](
+    new ElasticsearchIndex(
       "gvcf-v2",
+      GvcfIndex,
       ElasticsearchFieldMapper.StringsToTextFieldsWithSubKeywords
     )
 
-  implicit val WgsCram: ElasticsearchIndex[DocumentWgsCram] =
+  implicit val WgsCram: ElasticsearchIndex[WgsCramIndex.type] =
```
pshapiro4broad: Since `WgsCramIndex` is an argument to the constructor, is there a way to avoid having to pass `.type` as a type parameter too?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAO.scala
```diff
@@ -73,23 +69,24 @@ abstract class PersistenceDAO(recoveryParallelism: Int) extends LazyLogging {
...
-    index: ElasticsearchIndex[D]
+    index: ElasticsearchIndex[_]
   ): Future[Unit] = Future {
-    val jsonString =
-      ModelAutoDerivation.defaultPrinter.pretty(document.asJson(index.encoder))
-
     val writePath = (rootPath / index.persistenceDirForDatetime(dt)).createDirectories()
 
-    val written = (writePath / ClioDocument.persistenceFilename(document.upsertId))
+    val upsertId = new UpsertId(document.findAllByKey(UpsertId.UpsertIdFieldName).head.findAllByKey("id"))
```
pshapiro4broad: I am not sure what this constructor call is doing. Also, I see two copies of it. Maybe a named method somewhere (or a constructor override) with a comment that describes what it's doing would be helpful.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SearchDAO.scala
```diff
@@ -33,18 +29,6 @@ trait SearchDAO {
     */
   def close(): Future[Unit]
 
-  /**
```
pshapiro4broad: Did you mean to delete this javadoc? It looks like it could apply to the method that still remains.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SystemElasticsearchDAO.scala
```diff
@@ -27,7 +27,7 @@ trait SystemElasticsearchDAO extends SearchDAO { this: HttpElasticsearchDAO =>
   }
 
   override def initialize(
-    indexes: immutable.Seq[ElasticsearchIndex[_]]
+    indexes: immutable.Seq[ElasticsearchIndex[_, _]]
```
pshapiro4broad: am not sure why there are two type arguments here, I only see one above

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
```diff
@@ -1,39 +1,81 @@
 package org.broadinstitute.clio.server.dataaccess.elasticsearch
 
+import com.sksamuel.elastic4s.http.ElasticDsl.boolQuery
 import com.sksamuel.elastic4s.searches.queries.QueryDefinition
+import org.broadinstitute.clio.util.generic.{CaseClassMapper, CaseClassMapperWithTypes}
 
 /**
-  * Maps query input and outputs.
+  * Builds an ElasticsearchQueryMapper using shapeless and reflection.
```
pshapiro4broad: Are we using shapeless or reflection anymore?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/GvcfService.scala
```diff
@@ -62,20 +62,17 @@ class GvcfService(
 
 object GvcfService {
```
pshapiro4broad, line 66: It looks like we still have duplication across the different service objects, it would be nice if we could eliminate this duplication too, at some point.

pshapiro4broad (commented):
Just a few comments, looking forward to seeing the final version

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: Even if the names aren't exactly the same, it should be moved to the index class, since it is an index property.

danxmoran: I'm not sure which index you mean... `TransferIndex` shouldn't care what the name of the backing ES index is (for ex., whenever we switch all the client-side APIs from "wgs-ubam" to just "ubam" we'll probably keep the ES index named as "wgs-ubam", until we have a good way to migrate cloud storage). This _is_ a property of the `ElasticsearchIndex` class as-is, but it's one that's filled in through a constructor parameter.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/GvcfService.scala
danxmoran: That's @ApChagi's ticket

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: From a separation of concerns point of view, I agree that they aren't in the same domain. But I think you could imagine `WgsUbamIndex` as a configuration object that represents the superset of all configurable attributes of an index for clio.

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/GvcfWebServiceSpec.scala
```diff
@@ -112,14 +113,12 @@ class GvcfWebServiceSpec extends BaseWebserviceSpec {
...
-        .fold(throw _, identity)
-
-      firstUpdate.upsertId should be(responseAs[UpsertId])
-      firstUpdate.gvcfMd5 should be(Some('abcgithashdef))
-      firstUpdate.notes should be(Some("some note"))
-      firstUpdate.gvcfPath should be(
-        Some(URI.create(s"gs://path/gvcf${GvcfExtensions.GvcfExtension}"))
+
+      ElasticsearchIndex.getUpsertId(firstUpdate) should be(responseAs[UpsertId])
+      getStringByName(firstUpdate, "gvcf_md5") should be("abcgithashdef")
```
als364: Theoretically this is a `Symbol`, but I couldn't get `ScalaTest` to believe me that it was a `Symbol`:
```
[info] - should upsert a record, delete it and then fail to find it with query, but find it with queryall *** FAILED ***
[info]   'abcgithashdef has neither an abcgithashdef nor an isAbcgithashdef method (GvcfWebServiceSpec.scala:118)
```

danxmoran: Yeah, scalatest treats `Symbol` args to `be` weirdly. If you want it to be `Symbol` you could use `equal` instead of `be`, but I'm OK leaving it as a string here.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
als364: @pshapiro4broad, are you saying that `"wgs-ubam"`, `"gvcf-v2"`, etc. should be made into members of the `WgsUbamIndex`, `GvcfIndex`, etc. objects, and accessed and passed into the `ElasticsearchIndex` constructor from there?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
als364: Probably not, but I'd prefer to get rid of that as part of a task for deprecating the other mapper.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
als364: In order to do that, I'd either have to put another magic string in `UpsertId` (not great but probably not problematic) or include `clio-dataaccess-model` in `clio-util` (probably not ok). Are you ok with the magic string?

clio-util/src/main/scala/org/broadinstitute/clio/util/model/UpsertId.scala
als364: Makes sense, I can move this and `EntityId` into ElasticsearchIndex and get rid of the camel case names.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
als364: Nope, compiler complains that `ElasticsearchIndex` needs type parameters.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SearchDAO.scala
als364: I did not.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: Yes to moving them to members. Once they're members you don't need to pass them anymore because the index is being passed, so the `ElasticsearchIndex` code can use the members of the index instead of requiring them as parameters.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
pshapiro4broad: Not sure what you mean by magic string, my thinking was that we could add a `val` to `UpsertId` that contains the value of the expression `ElasticsearchUtil.toElasticsearchName(UpsertId.UpsertIdFieldName)`.  If that's not possible due to module dependencies you could make it a `val` in `HttpElasticsearchDAO` (which might be a better place for it anyway).

als364: I meant something like getting rid of `UpsertId.UpsertIdFieldName` entirely and creating a `val UpsertIdElasticsearchName = "upsert_id"` in `ElasticsearchIndex`.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: Did you try removing the type parameter from `ElasticsearchIndex` and using `index.type` in the code? It seems like there should be a way to avoid duplicating the arguments here. In other cases we were able to migrate type parameters to fields on `Index` objects.

als364: Removing the type parameter from `ElasticsearchIndex` entirely?

danxmoran: We could probably make that work... though it doesn't pair as nicely with Charley's in-flight changes. The idea was that we'd have a (simplified) signature like this for the service classes:
```scala
class IndexService[TI <: TransferIndex](index: TI, esIndex: ElasticsearchIndex[TI])
```
So you can bind the two together a bit.

build.sbt
```diff
@@ -145,7 +146,11 @@ lazy val `clio-client` = project
   * Integration tests for the clio-server and the clio-client.
   */
 lazy val `clio-integration-test` = project
-  .dependsOn(`clio-client`, `clio-dataaccess-model`)
+  .dependsOn(
+    `clio-client`,
+    `clio-dataaccess-model`,
+    `clio-util` % "compile->compile;it->test"
```
danxmoran: I think this can be consolidated down to `"it->compile,test"`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -58,45 +59,57 @@ class ElasticsearchIndex[Document: FieldMapper](
...
 }
 
 object ElasticsearchIndex extends ModelAutoDerivation {
+  val EntityIdElasticsearchName = "entity_id"
+
+  val UpsertIdElasticsearchName = "upsert_id"
+
+  def getEntityId(json: Json): String =
+    json.hcursor
+      .get[String](toElasticsearchName(EntityIdElasticsearchName))
```
danxmoran: Don't need `toElasticsearchName` here anymore.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -58,45 +59,57 @@ class ElasticsearchIndex[Document: FieldMapper](
...
+      .fold(throw _, identity)
+
+  def getUpsertId(json: Json): UpsertId =
+    json.hcursor
+      .get[UpsertId](toElasticsearchName(UpsertIdElasticsearchName))
```
danxmoran: Same here.

clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndexSpec.scala
```diff
@@ -32,7 +33,10 @@ class ElasticsearchIndexSpec extends FlatSpec with Matchers with ModelAutoDeriva
     it should "fields for v1 document" in {
       // Snake-case-ify the bookkeeping fields.
       val bookkeeping =
-        Seq(ClioDocument.UpsertIdFieldName, ClioDocument.EntityIdFieldName).map { name =>
+        Seq(
+          ElasticsearchIndex.UpsertIdElasticsearchName,
+          ElasticsearchIndex.EntityIdElasticsearchName
+        ).map { name =>
           keywordField(name.replaceAll("([A-Z])", "_$1").toLowerCase)
```
danxmoran: I don't think you need the `replaceAll().toLowerCase` here any more.

clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndexSpec.scala
```diff
@@ -64,7 +68,10 @@ class ElasticsearchIndexSpec extends FlatSpec with Matchers with ModelAutoDeriva
     it should "fields for v2 document" in {
       // Snake-case-ify the bookkeeping fields.
       val bookkeeping =
-        Seq(ClioDocument.UpsertIdFieldName, ClioDocument.EntityIdFieldName).map { name =>
+        Seq(
+          ElasticsearchIndex.UpsertIdElasticsearchName,
+          ElasticsearchIndex.EntityIdElasticsearchName
+        ).map { name =>
           keywordField(ElasticsearchUtil.toElasticsearchName(name))
```
danxmoran: Shouldn't need the `toElasticsearchName` anymore.

danxmoran: Also, since we seem to use the `Seq(UpsertIdName, EntityIdName)` in a bunch of places, it might be worth making another `val BookkeepingNames` to store it, for other code to use.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -28,7 +39,8 @@ class RecoveryIntegrationSpec
   private def updateDoc(json: Json, pathFieldName: String): Json = {
     val oldUri = json.hcursor.get[String](pathFieldName).fold(throw _, identity)
     val updates = Seq(
-      ClioDocument.UpsertIdElasticSearchName -> Json.fromString(UpsertId.nextId().id),
+      ElasticsearchIndex.UpsertIdElasticsearchName -> Json
+        .fromString(UpsertId.nextId().id),
```
danxmoran: This could be `UpsertId.nextId().asJson`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -28,7 +39,8 @@ class RecoveryIntegrationSpec
...
       pathFieldName -> Json.fromString(s"$oldUri/$randomId")
```
danxmoran: This one could flip to using `asJson` too, now that I'm looking at it.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -37,50 +49,71 @@ class RecoveryIntegrationSpec
...
+      location = location
+    ).asJson
+    val metadataJson = TransferUbamV1Metadata(
       ubamPath = Some(randomUri(i)),
       documentStatus = Some(DocumentStatus.Normal)
-    ).asJson(ElasticsearchIndex.WgsUbam.encoder)
+    ).asJson
+    val upsertIdJson = Map(ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId()).asJson
+    val entityIdJson = Map(
+      ElasticsearchIndex.EntityIdElasticsearchName -> Symbol(
```
danxmoran: Doesn't need to be a `Symbol` any more.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -37,50 +49,71 @@ class RecoveryIntegrationSpec
...
+        s"$flowcellBarcode.$i.$libraryName.${location.entryName}"
+      )
+    ).asJson
+    keyJson.deepMerge(metadataJson).deepMerge(upsertIdJson).deepMerge(entityIdJson)
```
danxmoran: This is pretty much the body of `ElasticsearchDocumentMapper`. Could we move that class into the `dataaccess-model` project, so we can reuse its logic here?

danxmoran: Same comments for gvcf and cram setup below.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -117,12 +118,12 @@ trait GvcfTests {
...
-        val storedDocument = getJsonFrom[DocumentGvcf](returnedUpsertId)
-        storedDocument.location should be(expected.location)
-        storedDocument.project should be(expected.project)
-        storedDocument.sampleAlias should be(expected.sampleAlias)
-        storedDocument.version should be(expected.version)
-        storedDocument.gvcfPath should be(expected.gvcfPath)
+        val storedDocument = getJsonFrom(returnedUpsertId)(ElasticsearchIndex.Gvcf)
+        getLocation(storedDocument) should be(expected.location)
+        getStringByName(storedDocument, "project") should be(expected.project)
+        getStringByName(storedDocument, "sampleAlias") should be(expected.sampleAlias)
```
danxmoran: These field names will need to be snake case. It might be easier to just strip the bookkeeping fields from `storedDocument`, convert `expected` to JSON, and check their equality.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -117,12 +118,12 @@ trait GvcfTests {
...
+        getIntByName(storedDocument, "version") should be(expected.version)
+        getUriByName(storedDocument, "gvcfPath") should be(expected.gvcfPath)
```
danxmoran: If you do keep these `get` methods, you should be able to collapse them all into one generic:
`def getByName[A : Decoder](doc: Json, name: String): A`

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,52 @@
...
+    * @param metadata The TransferMetadata from which to create the document.
+    * @return
     */
-  def empty(key: ModelKey): Document
+  def document(key: Key, metadata: Metadata): Json = {
+    val upsertIdMap = Map(
+      ElasticsearchIndex.UpsertIdElasticsearchName -> genId()
+    )
+    val entityIdMap = Map(
+      ElasticsearchIndex.EntityIdElasticsearchName -> Symbol(
```
danxmoran: Doesn't need to be a `Symbol` any more. Why two separate `Map`s?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,52 @@
...
   * Maps metadata to an Elasticsearch document.
   *
-  * @tparam ModelKey      The type of key.
-  * @tparam ModelMetadata The type of the metadata.
-  * @tparam Document      The type of the document.
+  * @tparam Key      The type of the TransferKey.
+  * @tparam Metadata The type of the TransferMetadata.
   */
-abstract class ElasticsearchDocumentMapper[ModelKey, ModelMetadata, Document] {
+class ElasticsearchDocumentMapper[Key <: Product: Encoder, Metadata <: TransferMetadata[
```
danxmoran: We can probably remove the `<: TransferMetadata[Metadata]` type constraint. For what you're doing here, all that matters is that we've got an `Encoder` for it.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -1,29 +1,52 @@
...
+    val metadataJson = metadata.asJson
+    val upsertIdJson = upsertIdMap.asJson
+    val entityIdJson = entityIdMap.asJson
+    keyJson.deepMerge(metadataJson).deepMerge(upsertIdJson).deepMerge(entityIdJson)
+  }
+}
+
+object ElasticsearchDocumentMapper {
+
+  def apply[Key <: Product: Encoder, Metadata <: TransferMetadata[Metadata]: Encoder]: ElasticsearchDocumentMapper[
```
danxmoran: Relaxing the above constraint will let this be less insane too.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -57,27 +55,30 @@ class ServerService private (
...
+        .batch(
+          ServerService.RecoveryMaxBulkSize,
+          json => Map(ElasticsearchIndex.getEntityId(json) -> json)
+        ) { (idMap, json) =>
+          val id = ElasticsearchIndex.getEntityId(json)
+          val newJson = idMap.get(id) match {
+            case Some(oldJson) => {
+              logger.debug(
+                s"Merging upserts ${ElasticsearchIndex.getUpsertId(oldJson)} and ${ElasticsearchIndex
+                  .getUpsertId(json)} for id $id"
```
danxmoran: Curse ye, scalafmt... I think `logger.debug` can be passed a block, if you want to clean up a bit:
```scala
logger.debug {
  val oldId = ElasticsearchIndex.getUpsertId(oldJson)
  val newId = ...
  s"..."
}
```

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAOSpec.scala
```diff
@@ -112,41 +105,49 @@ class HttpElasticsearchDAOSpec
...
+    val index = new ElasticsearchIndex[ModelMockIndex](
+      ModelMockIndex("docs-" + UUID.randomUUID()),
       ElasticsearchFieldMapper.NumericBooleanDateAndKeywordFields
     )
 
     val documents =
       (1 to 4)
         .map("document-" + _)
-        .map(s => Document(UpsertId.nextId(), Symbol(s)))
+        .map(
```
danxmoran: Oof... this should get slightly less wonky if you use block-form for the arg to `map`, and build the bookkeeping as a single `Map`

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAOSpec.scala
```diff
@@ -112,41 +105,49 @@ class HttpElasticsearchDAOSpec
...
+          s =>
+            Map(
+              ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId()
+            ).asJson
+              .deepMerge(
+                Map(
+                  ElasticsearchIndex.EntityIdElasticsearchName -> Symbol(s)
```
danxmoran: Doesn't need to be a `Symbol` any more.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -42,27 +50,80 @@ class PersistenceDAOSpec
...
+      .deepMerge(metadata1.asJson)
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId
+            .nextId()
+        ).asJson
+      )
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.EntityIdElasticsearchName -> Symbol(
```
danxmoran: Doesn't need to be a `Symbol` any more, and should be able to get fused with above into a single `Map`.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -42,27 +50,80 @@ class PersistenceDAOSpec
...
+      .deepMerge(metadata2.asJson)
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId
+            .nextId()
+        ).asJson
+      )
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.EntityIdElasticsearchName -> Symbol(
```
danxmoran: Same comments

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -71,7 +132,39 @@ class PersistenceDAOSpec
...
+      .deepMerge(mockMetadataJson)
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId
+            .nextId()
+        ).asJson
+      )
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.EntityIdElasticsearchName -> Symbol(
```
danxmoran: etc

danxmoran (commented):
Looking good so far!

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
pshapiro4broad: OK with me, I didn't realize we no longer had a use for `UpsertIdFieldName` other than for making the ES name.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: If it's not ok to do it now, we can save this for a future refactor.

pshapiro4broad (approved):

als364:
Unit tests are passing, but integration tests are failing. I'll address the latest set of comments once I've got all the integration tests passing too.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -222,14 +225,22 @@ class RecoveryIntegrationSpec
...
           }))
+        }
+
+        // Sort before comparing so we can do a pairwise comparison, which saves a ton of time.
+        // Remove null and internal values before comparing.
+        val sortedActual = docs.asArray.value.map(mapper).sorted
+
+        // Already sorted by construction. Remove null and internal values here too.
+        val sortedExpected = expected.map(mapper)
+//          .map(_.mapObject(_.filter({
```
als364: Reminder to self to get rid of commented out dead code later.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
als364: UpsertIds are a different type than EntityIds and circe doesn't handle encoding `Any`s well.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
als364: Tried that - compiler doesn't like not knowing the type ahead of time.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
danxmoran: I think you can do this:
```scala
Json.fromFields(
  UpsertIdName -> genId().asJson,
  EntityIdName -> <the id>.asJson
)
```

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
danxmoran: Hwot! This should definitely be doable... what's the error you hit if you try it?

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
als364: I think I do need the `<: Product` in order for `productIterator` to work on line 32

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -0,0 +1,50 @@
...
+    val metadataJson = metadata.asJson
+    keyJson
+      .deepMerge(metadataJson)
+      .deepMerge(bookkeeping)
+  }
+}
+
+object ElasticsearchDocumentMapper {
+
+  def apply[Key <: Product: Encoder, Metadata: Encoder]: ElasticsearchDocumentMapper[
```
danxmoran: Might be able to massage this into slightly better formatting:
```scala
def apply[
  Key <: Product: Encoder, 
  Metadata: Encoder
]: ElasticsearchDocumentMapper[Key, Metadata] = ...
```

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -58,45 +58,56 @@ class ElasticsearchIndex[Document: FieldMapper](
...
+
+  val UpsertIdElasticsearchName = "upsert_id"
+
+  val BookkeepingNames = Seq(
+    UpsertIdElasticsearchName,
+    EntityIdElasticsearchName
+  )
+
+  def getEntityId(json: Json): String =
+    getByName[String](json, "entity_id")
```
danxmoran: Should use the `val` instead of the `String`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -58,45 +58,56 @@ class ElasticsearchIndex[Document: FieldMapper](
...
+
+  def getUpsertId(json: Json): UpsertId = {
+    getByName[UpsertId](json, "upsert_id")
```
danxmoran: Same here

clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapperSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+      mockFieldDouble = mockFieldDouble,
+      mockFieldInt = mockFieldInt,
+      mockFieldDate = None
+    )
+
+    val document = mapper.document(key, metadata)
+    ElasticsearchIndex.getUpsertId(document) should be(expectedId)
+    ElasticsearchIndex.getEntityId(document) should be(s"$keyLong.$keyString")
+    ElasticsearchIndex.getByName[Double](document, "mock_field_double") should be(
+      mockFieldDouble.get
```
danxmoran: If you wrap `mockFieldDouble` in the `Option` at the point it's passed into the metadata constructor, not the point where it's declared, then you won't need the `.get`

clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapperSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+    )
+    ElasticsearchIndex.getByName[Int](document, "mock_field_int") should be(
+      mockFieldInt.get
```
danxmoran: Same here

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -152,7 +172,10 @@ class RecoveryIntegrationSpec
   }
 
   private val keysToDrop =
-    Set(ClioDocument.UpsertIdElasticSearchName, ClioDocument.EntityIdElasticSearchName)
+    Set(
+      ElasticsearchIndex.UpsertIdElasticsearchName,
+      ElasticsearchIndex.EntityIdElasticsearchName
+    )
```
danxmoran, line 178: Could be `ElasticsearchIndex.BookkeepingFields.toSet` (that val could also probably be a `Set` to start, I don't think we rely on it being ordered)

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -186,14 +209,19 @@ class RecoveryIntegrationSpec
...
       } yield {
-        // Sort before comparing so we can do a pairwise comparison, which saves a ton of time.
-        val sortedActual = docs.asArray.value.sorted
-
-        // Remove null & internal values before comparing; already sorted by construction.
-        val sortedExpected = expected
-          .map(_.mapObject(_.filter {
+        // Helper function for filtering out values we don't care about matching.
+        def mapper(json: Json): Json = {
+          json.mapObject(_.filter({
```
danxmoran, line 214: Don't need the `(` `)` around the `{` `}` in the `filter`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -150,13 +161,25 @@ trait GvcfTests {
...
 
-      storedDocument1.copy(upsertId = upsertId2, gvcfPath = gvcfUri2) should be(
+      storedDocument1
+        .deepMerge(
+          Map(
+            ElasticsearchIndex.UpsertIdElasticsearchName -> upsertId2
+          ).asJson
+        )
+        .deepMerge(
+          Map(ElasticsearchUtil.toElasticsearchName("gvcfPath") -> gvcfUri2).asJson
```
danxmoran, line 181: Could just be snake-case to start?

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAOSpec.scala
```diff
@@ -112,41 +106,47 @@ class HttpElasticsearchDAOSpec
...
       ElasticsearchFieldMapper.NumericBooleanDateAndKeywordFields
     )
 
+    def generateBookkeeping(s: String): Json =
+      Map(
+        ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId()
+      ).asJson
+        .deepMerge(
+          Map(ElasticsearchIndex.EntityIdElasticsearchName -> s).asJson
+        )
```
danxmoran, line 129: Could be simplified with `Json.fromFields`

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -42,27 +52,75 @@ class PersistenceDAOSpec
...
+        Map(
+          ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId
+            .nextId()
+        ).asJson
+      )
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.EntityIdElasticsearchName -> s"$key1Long.$key1String"
+        ).asJson
+      )
```
danxmoran, line 82: Could be simplified with `Json.fromFields`

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -42,27 +52,75 @@ class PersistenceDAOSpec
...
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId
+            .nextId()
+        ).asJson
+      )
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.EntityIdElasticsearchName -> s"$key2Long.$key2String").asJson
       )
```
danxmoran: Same

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/PersistenceDAOSpec.scala
```diff
@@ -71,7 +129,37 @@ class PersistenceDAOSpec
...
+          ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId
+            .nextId()
+        ).asJson
+      )
+      .deepMerge(
+        Map(
+          ElasticsearchIndex.EntityIdElasticsearchName -> s"$mockKeyLong.$mockKeyString"
+        ).asJson
+      )
+  )
```
danxmoran, line 161: Same

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapperSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+      Some(1234L)
+    )
+    val fields = key.asJson.deepMerge(metadata.asJson)
+    val output = fields
+      .deepMerge(
+        Map(ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId()).asJson
+      )
+      .deepMerge(
+        Map(ElasticsearchIndex.EntityIdElasticsearchName -> s"$keyLong.$keyString").asJson
+      )
```
danxmoran, line 100: Same

clio-server/src/test/scala/org/broadinstitute/clio/server/service/GvcfServiceSpec.scala
```diff
@@ -92,15 +98,20 @@ class GvcfServiceSpec extends TestKitSuite("GvcfServiceSpec") {
...
-          GvcfService.v1DocumentConverter.empty(transferKey),
+        .document(
+          transferKey,
           transferMetadata.copy(documentStatus = expectedDocumentStatus)
         )
-        .copy(upsertId = returnedUpsertId)
+        .deepMerge(
+          Map(
+            ElasticsearchUtil
+              .toElasticsearchName(ElasticsearchIndex.UpsertIdElasticsearchName) -> returnedUpsertId
```
danxmoran, line 108: Doesn't need the `toElasticsearchName` wrapper

clio-server/src/test/scala/org/broadinstitute/clio/server/service/GvcfServiceSpec.scala
```diff
@@ -92,15 +98,20 @@ class GvcfServiceSpec extends TestKitSuite("GvcfServiceSpec") {
...
     } yield {
       val expectedDocument = GvcfService.v1DocumentConverter
-        .withMetadata(
-          GvcfService.v1DocumentConverter.empty(transferKey),
+        .document(
+          transferKey,
           transferMetadata.copy(documentStatus = expectedDocumentStatus)
         )
-        .copy(upsertId = returnedUpsertId)
+        .deepMerge(
```
danxmoran, line 105: Instead of `deepMerge`, could we use `mapObject` with `add`?

clio-server/src/test/scala/org/broadinstitute/clio/server/service/GvcfServiceSpec.scala
```diff
@@ -53,6 +57,8 @@ class GvcfServiceSpec extends TestKitSuite("GvcfServiceSpec") {
         Seq(
           GvcfService.v1QueryConverter.buildQuery(
             transferInput.copy(documentStatus = Option(DocumentStatus.Normal))
+          )(
+            ElasticsearchIndex.Gvcf
```
danxmoran, line 61: Doesn't need to be on newline

clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala
```diff
@@ -87,26 +103,63 @@ class ServerServiceSpec extends TestKitSuite("ServerServiceSpec") {
     val app =
       MockClioApp(persistenceDAO = persistenceDAO, searchDAO = searchDAO)
 
-    val numDocs = 1000
+    val numDocs = 5
```
danxmoran, line 106: Put back to 1000?

clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala
```diff
@@ -87,26 +103,63 @@ class ServerServiceSpec extends TestKitSuite("ServerServiceSpec") {
...
+      document
+        .deepMerge(
+          Map(
+            ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId()
+          ).asJson
+        )
+        .deepMerge(
+          Map(
+            ElasticsearchIndex.EntityIdElasticsearchName -> s"$keyLong.$keyString-$counter"
+          ).asJson
```
danxmoran, line 146: `Json.fromFields`

clio-server/src/test/scala/org/broadinstitute/clio/server/service/WgsCramServiceSpec.scala
```diff
@@ -53,6 +59,8 @@ class WgsCramServiceSpec extends TestKitSuite("WgsCramServiceSpec") {
         Seq(
           WgsCramService.v1QueryConverter.buildQuery(
             transferInput.copy(documentStatus = Option(DocumentStatus.Normal))
+          )(
+            ElasticsearchIndex.WgsCram
```
danxmoran: Doens't need newline

clio-server/src/test/scala/org/broadinstitute/clio/server/service/WgsCramServiceSpec.scala
```diff
@@ -92,15 +100,20 @@ class WgsCramServiceSpec extends TestKitSuite("WgsCramServiceSpec") {
...
-          WgsCramService.v1DocumentConverter.empty(transferKey),
+        .document(
+          transferKey,
           transferMetadata.copy(documentStatus = expectedDocumentStatus)
         )
-        .copy(upsertId = returnedUpsertId)
+        .deepMerge(
+          Map(
+            ElasticsearchUtil
+              .toElasticsearchName(ElasticsearchIndex.UpsertIdElasticsearchName) -> returnedUpsertId
```
danxmoran: No `toElasticsearchName`, `mapObject` with `add`

clio-server/src/test/scala/org/broadinstitute/clio/server/service/WgsUbamServiceSpec.scala
```diff
@@ -50,6 +56,8 @@ class WgsUbamServiceSpec extends TestKitSuite("WgsUbamServiceSpec") {
         Seq(
           WgsUbamService.v1QueryConverter.buildQuery(
             transferInput.copy(documentStatus = Option(DocumentStatus.Normal))
+          )(
+            ElasticsearchIndex.WgsUbam
```
danxmoran: Doesn't need newline

clio-server/src/test/scala/org/broadinstitute/clio/server/service/WgsUbamServiceSpec.scala
```diff
@@ -87,15 +95,20 @@ class WgsUbamServiceSpec extends TestKitSuite("WgsUbamServiceSpec") {
...
-          WgsUbamService.v1DocumentConverter.empty(transferKey),
+        .document(
+          transferKey,
           transferMetadata.copy(documentStatus = expectedDocumentStatus)
         )
-        .copy(upsertId = returnedUpsertId)
+        .deepMerge(
+          Map(
+            ElasticsearchUtil
+              .toElasticsearchName(ElasticsearchIndex.UpsertIdElasticsearchName) -> returnedUpsertId
```
danxmoran, line 103: No `toElasticsearchName`, `mapObject` with `name`

clio-util/src/main/scala/org/broadinstitute/clio/util/model/UpsertId.scala
```diff
@@ -15,6 +15,13 @@ final class UpsertId private (val id: String) extends Ordered[UpsertId] {
   }
   override def hashCode(): Int = id.hashCode
   override def toString: String = id
+
+  /**
+    * The filename used for persisting a document's upsert data.
+    *
+    * @return the filename where the upsert with the given ID's data is stored
+    */
+  def persistenceFilename: String = s"${id}.json"
```
danxmoran: Doesn't need `{` `}` around `id`

danxmoran (commented):
Looking good! Just small comments.

Best of luck with the merge 🤞

clio-server/src/test/scala/org/broadinstitute/clio/server/service/WgsCramServiceSpec.scala
als364: And yet sbt thinks it does.

danxmoran: Hwot

als364: I can try to move things around and see if I can get the autoformatter to not ruin everything

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -28,59 +39,68 @@ class RecoveryIntegrationSpec
...
-      ClioDocument.UpsertIdElasticSearchName -> Json.fromString(UpsertId.nextId().id),
-      pathFieldName -> Json.fromString(s"$oldUri/$randomId")
+      ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId().asJson,
+      pathFieldName -> s"$oldUri/$randomId".asJson
     )
     json.deepMerge(Json.fromFields(updates))
   }
 
+  private val ubamMapper =
+    ElasticsearchDocumentMapper[TransferUbamV1Key, TransferUbamV1Metadata]
```
pshapiro4broad, line 49: Looks like this is another case where we could refactor to reduce code duplication, in a future PR.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/SearchDAO.scala
```diff
@@ -36,15 +32,10 @@ trait SearchDAO {
   /**
     * Update-or-insert (upsert) metadata into an index.
     *
-    * @param document A (potentially partial) metadata document containing
-    *                 new fields to set on the document in the index.
+    * @param documents One or more(potentially partial) metadata documents containing
```
pshapiro4broad: typo missing space before `(`

danxmoran (approved):
:shipit: nice work!

PR #221, opened 2018-03-02
DSDEGP-2009: Create key, metatdata, query-input, and query-output classes for Arrays

tbl3rd:
The TransferIndex classes for Arrays compile now.
I'm sure they are wrong, but don't know how to make them right.

### Description

----

* Review cycle:


clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,111 @@
...
+import org.broadinstitute.clio.transfer.model.TransferMetadata
+import org.broadinstitute.clio.util.model.DocumentStatus
+
+case class TransferArraysV1Metadata(
+  autocallCallRate: Option[Double] = None,
+  autocallDate: Option[OffsetDateTime] = None,
+  autocallGender: Option[Symbol] = None,
+  autocallPf: Option[Boolean] = None,
+  autocallVersion: Option[Int] = None,
+  beadPoolManifestFile: Option[URI] = None,
```
danxmoran: In other indexes we usually suffix these with `Path` instead of `File`

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,111 @@
...
+  callRate: Option[Double] = None,
+  callRateThreshold: Option[Float] = None,
```
danxmoran: Curious: why `Float` for this one, instead of `Double`?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,111 @@
...
+      extendedChipManifestFile,
+      fileOfIdatFilenames,
+      genderClusterFile,
+      haplotypeDatabaseFile,
+      paramsFile,
+      refDict,
+      refFasta,
+      refFastaIndex,
+      zcallThresholdsFile
+    )
```
danxmoran: Do we want all of these to get wiped out by the delete CLP, or do we want to leave some around (like we do with metrics files for the cram / gvcf indexes)?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,111 @@
...
+      pathMapper(fileOfIdatFilenames, ArraysExtensions.IdatsTxtExtension),
+    genderClusterFile = pathMapper(genderClusterFile, ArraysExtensions.EgtExtension),
+    haplotypeDatabaseFile =
+      pathMapper(haplotypeDatabaseFile, ArraysExtensions.TxtExtension),
+    paramsFile = pathMapper(paramsFile, ArraysExtensions.TxtExtension),
+    refDict = pathMapper(refDict, ArraysExtensions.DictExtension),
+    refFasta = pathMapper(refFasta, ArraysExtensions.FastaExtension),
+    refFastaIndex = pathMapper(refFastaIndex, ArraysExtensions.FastaFaiExtension),
+    zcallThresholdsFile =
+      pathMapper(zcallThresholdsFile, ArraysExtensions.EgtThresholdsTxtExtension)
```
danxmoran: An extra twist: this `copy` is used to build the set of moves that should happen for data-delivery. Not sure if there's anything we should leave out, like we do with cram metrics.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  val EgtExtension = ".egt"
+
+  val FastaExtension = ".fasta"
+
+  val FaiExtension = ".fai"
+
+  val GzExtension = ".gz"
+
+  val IdatsExtension = ".idats"
```
gbggrant: should be idat extension.  There are two idat files that we need to store here.  The workflow uses the idats file, but it's contents are the paths to the red and green idats.  We don't need to store anything about the idats file if you get what I mean.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
```diff
@@ -0,0 +1,41 @@
...
+/**
+  *  Extensions for files in the arrays index.
+  */
+object ArraysExtensions {
+
+  val BpmExtension = ".bpm"
+
+  val CsvExtension = ".csv"
+
+  val DictExtension = ".dict"
```
gbggrant, line 12: Many of these extensions are not arrays specific.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  val VcfExtension = ".vcf"
+
+  val VcfGzExtension = s"${VcfExtension}${GzExtension}"
+
+  val VcfGzTbiExtension = s"${VcfGzExtension}${TbiExtension}"
+
+  val EgtThresholdsTxtExtension = s"${EgtExtension}${ThresholdsExtension}${TxtExtension}"
+
+  val IdatsTxtExtension = s"${IdatsExtension}${TxtExtension}"
```
gbggrant: idats -> idat

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,151 @@
...
+
+import java.net.URI
+import java.time.OffsetDateTime
+
+import org.broadinstitute.clio.transfer.model.TransferMetadata
+import org.broadinstitute.clio.util.model.DocumentStatus
+
+case class TransferArraysV1Metadata(
+  autocallCallRate: Option[Double] = None,
+  autocallDate: Option[OffsetDateTime] = None,
```
gbggrant: why OffsetDateTime?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
gbggrant: Don't need to store the callRateThreshold.  It's used by the workflow to make a QC decision on the workflow.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,151 @@
...
+  autocallGender: Option[Symbol] = None,
+  autocallPf: Option[Boolean] = None,
+  autocallVersion: Option[Int] = None,
+  beadPoolManifestFile: Option[URI] = None,
+  callRate: Option[Double] = None,
+  callRateThreshold: Option[Float] = None,
+  chipType: Option[String] = None,
+  clusterFile: Option[URI] = None,
+  controlSampleName: Option[String] = None,
+  createdAt: Option[OffsetDateTime] = None,
```
gbggrant: These are timestamps in the database for when the row was created.  Don't we have this in clio

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
```diff
@@ -0,0 +1,151 @@
...
+  genotypeConcordanceSummaryMetrics: Option[URI] = None,
+  grnIdat: Option[URI] = None,
+  gtcFile: Option[URI] = None,
+  haplotypeDatabaseFile: Option[URI] = None,
+  hetHomvarRatio: Option[Double] = None,
+  hetPct: Option[Double] = None,
+  imagingDate: Option[OffsetDateTime] = None,
+  isLatest: Option[Boolean] = None,
+  isZcalled: Option[Boolean] = None,
+  modifiedAt: Option[OffsetDateTime] = None,
```
gbggrant: These are timestamps in the database for when the row was created.  Don't we have this in clio?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1QueryInput.scala
```diff
@@ -0,0 +1,30 @@
...
+case class TransferArraysV1QueryInput(
+  documentStatus: Option[DocumentStatus] = None,
+  location: Option[Location] = None,
+  project: Option[String] = None,
+  sampleAlias: Option[String] = None,
+  version: Option[Int] = None,
+  pipelineVersion: Option[Symbol] = None,
+  workflowStartDate: Option[OffsetDateTime] = None,
+  workflowEndDate: Option[OffsetDateTime] = None,
+  cramMd5: Option[Symbol] = None,
```
gbggrant: No Crams in arrays

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1QueryInput.scala
```diff
@@ -0,0 +1,30 @@
...
+  cramSize: Option[Long] = None,
+  cramPath: Option[URI] = None,
+  cromwellId: Option[UUID] = None,
+  readgroupMd5: Option[Symbol] = None,
+  workspaceName: Option[String] = None,
+  regulatoryDesignation: Option[RegulatoryDesignation] = None,
```
gbggrant: I don't think we use regulatoryDesignation in arrays

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
pshapiro4broad: I would just make this `FileExtension` and not try to organize extensions by index type. Using singular rather than plural and dropping the redundant `Extension` from the name would make them easier to read, IMO, e.g. `FileExtension.Bpm` instead of `FileExtensions.BpmExtension`.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
pshapiro4broad: Is the idea here to preserve the creation time from arrays database row? Not sure what the use of this field is. Anyway, elasticsearch itself doesn't automatically create a 'created at' property for you, and I don't think clio has this concept either.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1QueryInput.scala
pshapiro4broad: We don't think we'll ever do clinical arrays? Or are you saying that we don't have that functionality now, so why add it now?

pshapiro4broad (approved):

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Key.scala
```diff
@@ -0,0 +1,20 @@
...
+  location: Location,
+  chipwellBarcode: String,
+  analysisVersionNumber: Int,
+  version: Int
+) extends TransferKey {
+
+  override def getUrlSegments: Seq[String] =
+    Seq(
+      location.entryName,
+      chipwellBarcode,
```
jacarey: We may want to separate chip barcode from chip position. It could help with chip based queries. Thoughts?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
jacarey: I'm not sure we need the precision of a Double on any of these metrics to TBH

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1QueryInput.scala
```diff
@@ -0,0 +1,30 @@
...
+  sampleAlias: Option[String] = None,
+  version: Option[Int] = None,
+  pipelineVersion: Option[Symbol] = None,
+  workflowStartDate: Option[OffsetDateTime] = None,
+  workflowEndDate: Option[OffsetDateTime] = None,
+  cramMd5: Option[Symbol] = None,
+  cramSize: Option[Long] = None,
+  cramPath: Option[URI] = None,
+  cromwellId: Option[UUID] = None,
+  readgroupMd5: Option[Symbol] = None,
```
jacarey: no readgroups either.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
tbl3rd: Not sure I get what you mean about which files to store. I was purposely redundant here to make it easier to just delete what we don't want rather than have to discover what we do want. We obviously don't have to store both the file containing the idat paths and the idat paths themselves. I think it would be simpler to just store the two .idat files, but want confirmation.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
tbl3rd: Aside from mistakes: the fields all come from some Arrays data source I consulted. CREATED_AT is a column in the ARRAYS_QC table in the cloud metrics database. (I'll drop the fields derived from the *_AT entries unless someone objects.)

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
tbl3rd: Why not OffsetDateTime?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
tbl3rd: I don't know. Probably. Does anyone know the answer now? Or can it wait?

danxmoran: @ktibbett might know... but it can also wait, as long as we remember to fix it up before we release the CLPs.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
danxmoran: @ktibbett probably also knows the list of things that should be delivered and what shouldn't move around. From working on the delivery side of things, it looks like we'll want to deliver:
1. vcf
2. idat(s)
3. gtc

And any indexes for those things (I'm guessing).

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Metadata.scala
pshapiro4broad: `OffsetDateTime` is what we currently use to store dates in elasticsearch.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/TransferArraysV1Key.scala
tbl3rd: Please explain more, Jay.

jacarey: chip well barcode is actually and combination of the chip barcode and the well position.  so 123456789_R01C02 is chip barcode 123456789 row 01 column 02

If someone noticed something was wrong with a sample they might want to go back and look at other samples on that same chip. So we could either keep chipwell barcode as is and look for all keys that are like 123456789_* or we could separate the chip barcode from the well position and just search for all samples with a chip barcode of 123456789.

jacarey (approved):

PR #224, opened 2018-03-08
DSDEGP-1643 remove duplicate code across indices

charlesdyan:
### Description

Refactored individual services and web services to extend a base IndexService and IndexWebService respectively to eliminate duplicate code. Also refactored tests to use the new code and changed web service tests to test the webservice instead of the dao/service.

----

* Review cycle:


clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
```diff
@@ -67,23 +62,36 @@ object ClioServer
...
+      new WgsCramService(persistenceService, searchService)
+    )
+
+  private val apiRoutes: Route =
+    concat(wgsUbamWebService.routes, gvcfWebService.routes, wgsCramWebService.routes)
+
+  private val httpServerDAO = AkkaHttpServerDAO(wrapperDirectives, infoRoutes, apiRoutes)
+
+  private val serverService = ServerService(app, httpServerDAO)
+  override val statusService = StatusService(app, httpServerDAO)
```
danxmoran: Unrelated, but if we're waiting to merge this... could you make whatever this is `override`-ing a class instead of a trait, then add it in through composition? Similar to what you've done with the web services. It makes setting up an isolated instance for testing much easier.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -0,0 +1,80 @@
...
+
+abstract class IndexService[TI <: TransferIndex, D <: ClioDocument: ElasticsearchIndex](
+  persistenceService: PersistenceService,
+  searchService: SearchService,
+  val transferIndex: TI,
+  elasticsearchIndex: ElasticsearchIndex[D]
+)(
+  implicit
+  executionContext: ExecutionContext,
+  documentTag: ClassTag[D]
```
danxmoran: This could also be a context bound above:
```scala
D <: ClioDocument : ElasticsearchIndex: ClassTag
```

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -33,13 +33,14 @@ class PersistenceService private (persistenceDAO: PersistenceDAO, searchDAO: Sea
   def upsertMetadata[TK, TM, D <: ClioDocument: ElasticsearchIndex](
     transferKey: TK,
     transferMetadata: TM,
-    documentMapper: ElasticsearchDocumentMapper[TK, TM, D]
+    documentMapper: ElasticsearchDocumentMapper[TK, TM, D],
+    elasticsearchIndex: ElasticsearchIndex[D]
```
danxmoran: Does it work if `elasticsearchIndex` is `implicit` too? Then we won't have to pass things explicitly below.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/SearchService.scala
```diff
@@ -27,13 +27,14 @@ class SearchService private (searchDAO: SearchDAO) {
     */
   def queryMetadata[TI, TO, D <: ClioDocument: ElasticsearchIndex](
     transferInput: TI,
-    queryMapper: ElasticsearchQueryMapper[TI, TO, D]
+    queryMapper: ElasticsearchQueryMapper[TI, TO, D],
+    elasticsearchIndex: ElasticsearchIndex[D]
```
danxmoran: Same question about making this implicit.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.broadinstitute.clio.util.model.{DocumentStatus, UpsertId}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.reflect.ClassTag
+
+abstract class IndexService[TI <: TransferIndex, D <: ClioDocument: ElasticsearchIndex](
+  persistenceService: PersistenceService,
+  searchService: SearchService,
+  val transferIndex: TI,
+  elasticsearchIndex: ElasticsearchIndex[D]
```
danxmoran: Could we make this `implicit`? (I think) it should thread through nicely if we do.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -163,13 +163,14 @@ object ServerService {
...
 
   def apply(
-    app: ClioApp
+    app: ClioApp,
+    httpServerDAO: HttpServerDAO
   )(implicit executionContext: ExecutionContext, mat: Materializer): ServerService = {
     new ServerService(
       app.serverStatusDAO,
-      app.httpServerDAO,
       app.persistenceDAO,
```
danxmoran: Do you think it'd be a ton of work to scrap `ClioApp` entirely, and just pass the `DAO`s around explicitly? I feel like the whole app setup obfuscates more than it helps.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.broadinstitute.clio.server.service.IndexService
+import org.broadinstitute.clio.transfer.model.TransferIndex
+
+abstract class IndexWebService[TI <: TransferIndex, D <: ClioDocument: ElasticsearchIndex](
+  val indexService: IndexService[TI, D]
+) extends JsonWebService {
+
+  import indexService.transferIndex.implicits._
+
+  val servicePathPrefix: String
```
danxmoran: We shouldn't need this; we can use [this](https://github.com/broadinstitute/clio/blob/develop/clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferIndex.scala#L17) instead.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  lazy val routes: Route = {
+    pathPrefix("v1") {
+      pathPrefix(servicePathPrefix) {
+        concat(getSchema, postMetadata, query, queryall)
+      }
+    }
+  }
+
+  private[webservice] val pathPrefixKey: Directive1[indexService.transferIndex.KeyType]
```
danxmoran: Can you make this a `def`? It's general Scala best-practice, because it lets the individual implementations choose whether the implementation should use a `def`, `lazy val`, or `val`.

clio-server/src/test/scala/org/broadinstitute/clio/server/service/IndexServiceSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+  val dummyInput: indexService.transferIndex.QueryInputType
+
+  def getDummyMetadata(
+    documentStatus: Option[DocumentStatus]
+  ): indexService.transferIndex.MetadataType
+
+  def getService(
+    persistenceService: PersistenceService,
+    searchService: SearchService
+  ): IndexService[TI, D]
```
danxmoran: Newline after this pls

clio-server/src/test/scala/org/broadinstitute/clio/server/service/IndexServiceSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+  def copyDocumentWithUpsertId(originalDocument: D, upsertId: UpsertId): D
+  def documentToJson(document: D): Json
```
danxmoran: Do we need this? I think we should be able to replace it with `document.asJson(index.encoder)`.

clio-server/src/test/scala/org/broadinstitute/clio/server/service/IndexServiceSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+            dummyInput.withDocumentStatus(Option(DocumentStatus.Normal))
+          )
+        )
+      )
+    }
+  }
+
+  private def upsertMetadataTest(
+    documentStatus: Option[DocumentStatus],
+    expectedDocumentStatus: Option[DocumentStatus]
```
danxmoran: We could consolidate these into 1 arg for `documentStatus`, then calculate expected as: `documentStatus.orElse(DocumentStatus.Normal)`. It'll make the call-sites more clear (I was wondering if the duplicate `Option(DocumentStatus.Normal)` was a typo).

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/IndexWebServiceSpec.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  val memorySearchDAO = new MemorySearchDAO()
+
+  def webServiceName: String
+  val mockService: MockIndexService[TI, D]
+  val webService: IndexWebService[TI, D]
+  val onPremKey: webService.indexService.transferIndex.KeyType
+  val cloudKey: webService.indexService.transferIndex.KeyType
+  val badMetadataMap: Map[String, String]
+  val badQueryInputMap: Map[String, String]
```
danxmoran: Can you make all these `def`s?

danxmoran (commented):
🎉

clio-server/src/main/scala/org/broadinstitute/clio/server/service/StatusService.scala
```diff
@@ -13,8 +13,8 @@ import scala.util.{Success, Try}
 
 class StatusService private (
   serverStatusDAO: ServerStatusDAO,
-  httpServerDAO: HttpServerDAO,
-  searchDAO: SearchDAO
+  searchDAO: SearchDAO,
```
pshapiro4broad: was there a reason to swap the order of the args here?

clio-server/src/test/scala/org/broadinstitute/clio/server/webservice/IndexWebServiceSpec.scala
charlesdyan: Looks like I can't access members of webService if I make it a def, so leaving that one

clio-server/src/test/scala/org/broadinstitute/clio/server/service/IndexServiceSpec.scala
charlesdyan: tried that initially, but can't access .asJson from a non-specific ClioDocument.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
charlesdyan: probably not too bad? We'll see!

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
charlesdyan: These three implicifications result in ambiguous implicit value errors... but it actually looks like we can just strip this parameter out entirely. And in PersistenceService and SearchService as well

clio-server/src/test/scala/org/broadinstitute/clio/server/service/IndexServiceSpec.scala
danxmoran: `asJson` is pulled in from a circe import, it's not tied to the specific document types. Try with
`import io.circe.syntax._`

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
charlesdyan: Not sure what you mean by this. ClioServer extends a JsonWebService with several traits. If the StatusWebService trait that this val comes from is an abstract Class.. then in order to preserve everything else, it would have to extend a JsonWebService. Is that what you mean? I also don't entirely understand why it's an override in the first place, since it's just an abstract def in the trait

danxmoran: I mean, make `ClioServer` no longer extend any of these traits, and instead have it include one member of each. So it'd have a `StatusWebService` member, similar to how it now has a `wgsUbamWebService` member. Then we don't have to go through all the headaches of making sure the names of fields within the traits don't clash, and it's easier to instantiate a stand-alone version of each web-service for testing.

`JsonWebService` can still be the parent trait of all the individual web-services, but it shouldn't be mixed into `ClioServer` any more.

(This could be another chunky refactor on its own; just a suggestion for something to tear into if you're waiting on the other refactor to merge)

danxmoran:
retest this please

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebService.scala
```diff
@@ -2,25 +2,16 @@ package org.broadinstitute.clio.server.webservice
...
+    extends IndexWebService[WgsUbamIndex.type, DocumentWgsUbam](wgsUbamService) {
 
-  lazy val wgsUbamRoutes: Route = {
-    pathPrefix("v1") {
-      pathPrefix("wgsubam") {
-        concat(getSchema, postMetadata, query, queryall)
-      }
-    }
-  }
+  val servicePathPrefix = "wgsubam"
```
danxmoran: Leftover from old code, not used

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/GvcfWebService.scala
```diff
@@ -2,72 +2,23 @@ package org.broadinstitute.clio.server.webservice
...
+    extends IndexWebService[GvcfIndex.type, DocumentGvcf](gvcfService) {
 
-  lazy val gvcfRoutes: Route = {
-    pathPrefix("v1") {
-      pathPrefix("gvcf") {
-        concat(gvcfGetSchema, gvcfPostMetadata, gvcfQuery, gvcfQueryall)
-      }
-    }
-  }
+  val servicePathPrefix = "gvcf"
```
danxmoran: Leftover from before, not used

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/WgsCramWebService.scala
```diff
@@ -2,72 +2,23 @@ package org.broadinstitute.clio.server.webservice
...
+    extends IndexWebService[WgsCramIndex.type, DocumentWgsCram](wgsCramService) {
 
-  lazy val wgsCramRoutes: Route = {
-    pathPrefix("v1") {
-      pathPrefix("wgscram") {
-        concat(cramGetSchema, cramPostMetadata, cramQuery, cramQueryall)
-      }
-    }
-  }
+  val servicePathPrefix = "wgscram"
```
danxmoran: Leftover from before, not used

clio-server/src/test/scala/org/broadinstitute/clio/server/service/MockIndexService.scala
```diff
@@ -0,0 +1,60 @@
...
+)(
+  implicit
+  executionContext: ExecutionContext,
+  documentTag: ClassTag[D]
+) extends IndexService(
+      PersistenceService(app),
+      SearchService(app),
+      transferIndex
+    ) {
+  val queryCalls: ArrayBuffer[transferIndex.QueryInputType]
```
danxmoran: Hrm, does assigning the `ArrayBuffer.empty[transferIndex.QueryInputType]` here (etc) not work?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/TransferInput.scala
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.clio.transfer.model
+
+import org.broadinstitute.clio.util.model.DocumentStatus
+
+trait TransferInput[IP <: TransferInput[IP]] { self: IP =>
```
danxmoran: Should be more specific, i.e. `TransferQueryInput`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/StatusService.scala
charlesdyan: Not sure. Possibly just a consequence of multiple passes when taking HttpServerDAO out of the app class. Does seem like it aligns with order of use this way

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
charlesdyan: Were you thinking all of the Directives that are mixed in to ClioServer as well? I've done it for StatusWebService... but it involved a little funkiness with lazification and forward referencing

clio-server/src/main/scala/org/broadinstitute/clio/server/service/GvcfService.scala
```diff
@@ -21,57 +12,10 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  }
-
-  def queryAllMetadata(
-    transferInput: TransferGvcfV1QueryInput
-  ): Source[Json, NotUsed] = {
-    searchService.queryMetadata(
-      transferInput,
-      GvcfService.v1QueryConverter
+)(implicit executionContext: ExecutionContext)
+    extends IndexService[GvcfIndex.type](
```
danxmoran: Do you need the explicit type param any more? I think it was required before because you also had `D`, which wasn't inferrable from any of the type parameters.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  import transferIndex.implicits._
+
+  private[service] val v1DocumentConverter =
+    ElasticsearchDocumentMapper[
+      transferIndex.KeyType,
+      transferIndex.MetadataType
+    ]
+
+  val v1QueryConverter: ElasticsearchQueryMapper[
```
danxmoran: Should / can this also be `private[service]`?

danxmoran: Can drop the explicit type annotation w/o getting yelled at by IntelliJ if so

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -48,7 +47,7 @@ class PersistenceService private (persistenceDAO: PersistenceDAO, searchDAO: Sea
 
 object PersistenceService {
 
-  def apply(app: ClioApp): PersistenceService = {
-    new PersistenceService(app.persistenceDAO, app.searchDAO)
+  def apply(persistenceDAO: PersistenceDAO, searchDAO: SearchDAO): PersistenceService = {
```
danxmoran: I'd just delete this apply method, it's not buying us anything

clio-server/src/main/scala/org/broadinstitute/clio/server/service/SearchService.scala
```diff
@@ -43,7 +42,7 @@ class SearchService private (searchDAO: SearchDAO) {
 
 object SearchService {
 
-  def apply(app: ClioApp): SearchService = {
-    new SearchService(app.searchDAO)
+  def apply(searchDAO: SearchDAO): SearchService = {
```
danxmoran: Would also scrap this apply method

clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala
```diff
@@ -155,13 +155,16 @@ object ServerService {
   val RecoveryMaxBulkSize: Long = 1000L
 
   def apply(
```
danxmoran: Can scrap this one too; `ClioServer` can just pass `ClioServerConfig.Version.value` to the constructor when it builds the service

clio-server/src/main/scala/org/broadinstitute/clio/server/service/StatusService.scala
```diff
@@ -25,16 +20,21 @@ class StatusService private (
...
+      VersionInfo(
+        ClioServerConfig.Version.value
+      )
+    )
   }
 }
 
 object StatusService {
 
   def apply(
```
danxmoran: This one too

clio-server/src/main/scala/org/broadinstitute/clio/server/service/WgsCramService.scala
```diff
@@ -21,57 +12,10 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  }
-
-  def queryAllMetadata(
-    transferInput: TransferWgsCramV1QueryInput
-  ): Source[Json, NotUsed] = {
-    searchService.queryMetadata(
-      transferInput,
-      WgsCramService.v1QueryConverter
+)(implicit executionContext: ExecutionContext)
+    extends IndexService[WgsCramIndex.type](
```
danxmoran: Same question about explicit type param

clio-server/src/main/scala/org/broadinstitute/clio/server/service/WgsUbamService.scala
```diff
@@ -21,57 +12,10 @@ import scala.concurrent.{ExecutionContext, Future}
...
-  }
-
-  def queryAllMetadata(
-    transferInput: TransferUbamV1QueryInput
-  ): Source[Json, NotUsed] = {
-    searchService.queryMetadata(
-      transferInput,
-      WgsUbamService.v1QueryConverter
+)(implicit executionContext: ExecutionContext)
+    extends IndexService[WgsUbamIndex.type](
```
danxmoran: Same question

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/GvcfWebService.scala
```diff
@@ -2,72 +2,20 @@ package org.broadinstitute.clio.server.webservice
...
-
-  lazy val gvcfRoutes: Route = {
-    pathPrefix("v1") {
-      pathPrefix("gvcf") {
-        concat(gvcfGetSchema, gvcfPostMetadata, gvcfQuery, gvcfQueryall)
-      }
-    }
-  }
+class GvcfWebService(gvcfService: IndexService[GvcfIndex.type])
+    extends IndexWebService[GvcfIndex.type](gvcfService) {
```
danxmoran: Same type-param question

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/WgsCramWebService.scala
```diff
@@ -2,72 +2,20 @@ package org.broadinstitute.clio.server.webservice
...
-
-  lazy val wgsCramRoutes: Route = {
-    pathPrefix("v1") {
-      pathPrefix("wgscram") {
-        concat(cramGetSchema, cramPostMetadata, cramQuery, cramQueryall)
-      }
-    }
-  }
+class WgsCramWebService(wgsCramService: IndexService[WgsCramIndex.type])
+    extends IndexWebService[WgsCramIndex.type](wgsCramService) {
```
danxmoran: Same type-param question

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/WgsUbamWebService.scala
```diff
@@ -2,25 +2,13 @@ package org.broadinstitute.clio.server.webservice
...
-
-  lazy val wgsUbamRoutes: Route = {
-    pathPrefix("v1") {
-      pathPrefix("wgsubam") {
-        concat(getSchema, postMetadata, query, queryall)
-      }
-    }
-  }
+class WgsUbamWebService(wgsUbamService: IndexService[WgsUbamIndex.type])
+    extends IndexWebService[WgsUbamIndex.type](wgsUbamService) {
```
danxmoran: Same type-param question

clio-server/src/test/scala/org/broadinstitute/clio/server/service/MockWgsCramService.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import org.broadinstitute.clio.server.dataaccess.elasticsearch.ElasticsearchIndex
+import org.broadinstitute.clio.transfer.model.WgsCramIndex
+import org.broadinstitute.clio.transfer.model.wgscram.TransferWgsCramV1QueryOutput
+import org.broadinstitute.clio.util.model.Location
+
+import scala.concurrent.ExecutionContext
+
+class MockWgsCramService()(implicit executionContext: ExecutionContext)
+    extends MockIndexService[WgsCramIndex.type](
```
danxmoran: Same question about type param (and for all the other mocks)

danxmoran (approved):
Some last ideas for cleanup, but LGTM overall

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
charlesdyan: Just a note: scrapping these requires making each Service class private[server] rather than private or private[service] (to access from ClioServer). Seems like an acceptable change to me.

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
```diff
@@ -44,9 +44,9 @@ object ClioServer extends StrictLogging {
     ClioServerConfig.Persistence.recoveryParallelism
   )
 
-  private val persistenceService = PersistenceService(persistenceDAO, searchDAO)
-  private val searchService = SearchService(searchDAO)
-  private val statusService = StatusService(
+  private val persistenceService = new PersistenceService(persistenceDAO, searchDAO)
```
pshapiro4broad: I thought the scala way was to not use `new`?

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
```diff
@@ -48,42 +36,73 @@ object ClioServer
...
+  private val searchService = new SearchService(searchDAO)
+  private val statusService = new StatusService(
+    serverStatusDAO,
+    searchDAO
+  )
+
+  private val exceptionDirectives = new ExceptionDirectives
+  private val swaggerDirectives = new SwaggerDirectives
+  private val rejectionDirectives =
+    new RejectionDirectives(
```
pshapiro4broad: These seem like very short lines, does scalafmt insist on wrapping them?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -0,0 +1,72 @@
...
+  elasticsearchIndex: ElasticsearchIndex[TI],
+  val transferIndex: TI
+)(
+  implicit
+  executionContext: ExecutionContext
+) {
+
+  import transferIndex.implicits._
+
+  private[service] val v1DocumentConverter =
```
pshapiro4broad: In this general version, does it make sense to have `v1` prefixes on these fields?

pshapiro4broad (approved):
Looks much better! A shame about the `copy` methods but at least there's a lot less boilerplate.

tlangs (approved):

clio-server/src/main/scala/org/broadinstitute/clio/server/ClioServer.scala
charlesdyan: Having scrapped the apply methods for these, I don't think I can get away without `new` . @danxmoran ?

danxmoran: `new` is good. `apply` is useful for hiding details, but we don't have anything useful to hide here.

PR #232, opened 2018-03-28
Speculative conversion of gsutil to using the google cloud java APIs

pshapiro4broad:
NOT FOR MERGING

Prototype code to replace gsutil in clio-client with google cloud APIs

- code compiles
- haven't set up creds yet
- haven't handled exceptions yet
- removed some unused APIs instead of porting them

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -155,5 +155,3 @@ lazy val `clio-integration-test` = project
   .disablePlugins(AssemblyPlugin)
   .settings(commonSettings)
   .settings(libraryDependencies ++= Dependencies.IntegrationTestDependencies)
-
-libraryDependencies += "com.google.cloud" % "google-cloud-storage" % "1.14.0"
--- a/build.sbt
+++ b/build.sbt
@@ -155,3 +155,5 @@ lazy val `clio-integration-test` = project
   .disablePlugins(AssemblyPlugin)
   .settings(commonSettings)
   .settings(libraryDependencies ++= Dependencies.IntegrationTestDependencies)
+
+libraryDependencies += "com.google.cloud" % "google-cloud-storage" % "1.14.0"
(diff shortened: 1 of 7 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -124,55 +84,47 @@ object IoUtil extends IoUtil {
     */
   class GsUtil {
```
danxmoran: If we're doing cleanup, I'd rather get rid of this nested class entirely. It was only separate so we could inject mock standard output for testing our regexes.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -62,54 +59,12 @@ trait IoUtil {
     gsUtil.exists(path.toString) == 0
```
danxmoran: Too far away to comment on directly, but I'd also want to change `copyGoogleObject` and `deleteGoogleObject` to return `Unit` instead of `Int`, now that we can report failure through meaningful exceptions

build.sbt
```diff
@@ -155,3 +155,5 @@ lazy val `clio-integration-test` = project
   .disablePlugins(AssemblyPlugin)
   .settings(commonSettings)
   .settings(libraryDependencies ++= Dependencies.IntegrationTestDependencies)
+
+libraryDependencies += "com.google.cloud" % "google-cloud-storage" % "1.14.0"
```
danxmoran: What error were you seeing when you added this through the normal way?

pshapiro4broad: the library wasn't visible to my code, e.g. import didn't work

PR #238, opened 2018-04-02
Move overwrite validation to server.

jacarey:
This prevents the client from make two round trips. A query was already necessary to check default document status which would have caused 2 queries and one upsert for every add.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
@@ -38,10 +38,10 @@ class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI])
       )
 
     decodedOrError.fold(
-      Future.failed, { decoded =>
+      Future.failed, {
         addFiles(
           webClient,
-          decoded,
+          _,
           addCommand.force
         )
       }
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
@@ -6,10 +6,6 @@ import org.broadinstitute.clio.client.util.IoUtil
 import org.broadinstitute.clio.client.webclient.ClioWebClient
 import org.broadinstitute.clio.transfer.model.ClioIndex
 import org.broadinstitute.clio.util.ClassUtil
-import org.broadinstitute.clio.util.generic.{
-  CaseClassMapper,
-  CirceEquivalentCamelCaseLexer
-}
 import org.broadinstitute.clio.util.model.UpsertId
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -43,25 +39,11 @@ class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI])
 
     decodedOrError.fold(
       Future.failed, { decoded =>
-        val maybeQuery =
-          if (addCommand.force) {
-            // Don't bother querying if we're going to force-overwrite anyways.
-            Future.successful(None)
-          } else {
-            webClient
-              .getMetadataForKey(addCommand.index)(addCommand.key, includeDeleted = false)
-          }
-
-        for {
-          existingMetadata <- maybeQuery
-          upsertResponse <- addFiles(
-            webClient,
-            decoded,
-            existingMetadata
-          ).fold(Future.failed, identity)
-        } yield {
-          upsertResponse
-        }
+        addFiles(
+          webClient,
+          decoded,
+          addCommand.force
+        )
       }
     )
   }
@@ -69,72 +51,17 @@ class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI])
   private def addFiles(
     client: ClioWebClient,
     newMetadata: addCommand.index.MetadataType,
-    existingMetadata: Option[addCommand.index.MetadataType],
-  )(implicit ec: ExecutionContext): Either[Throwable, Future[UpsertId]] = {
-    val differences = existingMetadata.fold(Iterable.empty[(String, Any, Any)]) {
-      diffMetadata(_, newMetadata)
-    }
-    for {
-      upsertResponse <- Either.cond(
-        differences.isEmpty,
-        client
-          .upsert(addCommand.index)(addCommand.key, newMetadata)
-          .recover {
-            case ex =>
-              throw new RuntimeException(
-                s"An error occurred while adding the $name record in Clio.",
-                ex
-              )
-          }, {
-          val diffs = differences.map {
-            case (field, newVal, oldVal) =>
-              s"Field: $field, Old value: $oldVal, New value: $newVal"
-          }.mkString("\n")
+    force: Boolean
+  )(implicit ec: ExecutionContext): Future[UpsertId] = {
 
-          new IllegalArgumentException(
-            s"""Adding this document will overwrite the following existing metadata:
-               |$diffs
-               |Use '--force' to overwrite the existing data.""".stripMargin
+    client
+      .upsert(addCommand.index)(addCommand.key, newMetadata, force)
+      .recover {
+        case ex =>
+          throw new RuntimeException(
+            s"An error occurred while adding the $name record in Clio.",
+            ex
           )
-        }
-      )
-    } yield {
-      upsertResponse
-    }
-  }
-
-  private def diffMetadata(
-    existingMetadata: addCommand.index.MetadataType,
-    newMetadata: addCommand.index.MetadataType
-  ): Iterable[(String, Any, Any)] = {
-    val mapper = new CaseClassMapper[addCommand.index.MetadataType]
-
-    val existingMetadataValues =
-      mapper
-        .vals(existingMetadata)
-        .asInstanceOf[Map[String, Option[Any]]]
-        .filterNot(_._2.isEmpty)
-
-    val newMetadataValues =
-      mapper
-        .vals(newMetadata)
-        .asInstanceOf[Map[String, Option[Any]]]
-        .filterNot(_._2.isEmpty)
-
-    val differentFields =
-      newMetadataValues.keySet
-        .intersect(existingMetadataValues.keySet)
-        .filterNot { field =>
-          existingMetadataValues(field).equals(newMetadataValues(field))
-        }
-
-    differentFields.map { key =>
-      import s_mach.string._
-      (
-        key.toSnakeCase(CirceEquivalentCamelCaseLexer),
-        newMetadataValues(key).get,
-        existingMetadataValues(key).get
-      )
-    }
+      }
   }
 }
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
@@ -109,7 +109,11 @@ class DeleteExecutor[CI <: ClioIndex](deleteCommand: DeleteCommand[CI])
             )
         }
       upsertResponse <- client
-        .upsert(deleteCommand.index)(deleteCommand.key, markedAsDeleted)
+        .upsert(deleteCommand.index)(
+          deleteCommand.key,
+          markedAsDeleted,
+          deleteCommand.force
+        )
         .recover {
           case ex =>
             throw new RuntimeException(
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
@@ -112,7 +112,7 @@ class DeleteExecutor[CI <: ClioIndex](deleteCommand: DeleteCommand[CI])
         .upsert(deleteCommand.index)(
           deleteCommand.key,
           markedAsDeleted,
-          deleteCommand.force
+          force = true
         )
         .recover {
           case ex =>
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -197,7 +197,7 @@ class ClioWebClient(
 
     dispatchRequest(
       HttpRequest(
-        uri = Uri(path = encodedPath).withQuery(Uri.Query("force" -> force.toString)),
+        uri = Uri(path = encodedPath).withQuery(Uri.Query(forceString -> force.toString)),
         method = HttpMethods.POST,
         entity = entity
       )
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -15,7 +15,7 @@ import org.broadinstitute.clio.transfer.model._
 import org.broadinstitute.clio.util.generic.{CirceEquivalentCamelCaseLexer, FieldMapper}
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
 import org.broadinstitute.clio.util.model.UpsertId
-import org.broadinstitute.clio.util.ApiConstants._
+import ApiConstants._
 
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future, Promise}
@@ -167,12 +167,12 @@ class ClioWebClient(
   }
 
   def getClioServerVersion: Future[Json] = {
-    dispatchRequest(HttpRequest(uri = "/version"), includeAuth = false)
+    dispatchRequest(HttpRequest(uri = s"/$versionString"), includeAuth = false)
       .flatMap(unmarshal[Json])
   }
 
   def getClioServerHealth: Future[Json] = {
-    dispatchRequest(HttpRequest(uri = "/health"), includeAuth = false)
+    dispatchRequest(HttpRequest(uri = s"/$healthString"), includeAuth = false)
       .flatMap(unmarshal[Json])
   }
 
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.clio.transfer.model._
 import org.broadinstitute.clio.util.generic.{CirceEquivalentCamelCaseLexer, FieldMapper}
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
 import org.broadinstitute.clio.util.model.UpsertId
+import org.broadinstitute.clio.util.ApiConstants._
 
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future, Promise}
@@ -192,7 +193,7 @@ class ClioWebClient(
      * in the fields of the key.
      */
     val encodedPath = key.getUrlSegments.foldLeft(
-      Uri.Path(s"/api/v1/${clioIndex.urlSegment}/metadata")
+      Uri.Path(s"/$apiString/v1/${clioIndex.urlSegment}/$metadataString")
     )(_ / _)
 
     dispatchRequest(
@@ -255,7 +256,7 @@ class ClioWebClient(
     input: Json,
     includeDeleted: Boolean
   ): Future[Json] = {
-    val queryPath = if (includeDeleted) "queryall" else "query"
+    val queryPath = if (includeDeleted) queryAllString else queryString
 
     val entity = HttpEntity(
       ContentTypes.`application/json`,
@@ -263,7 +264,7 @@ class ClioWebClient(
     )
     dispatchRequest(
       HttpRequest(
-        uri = s"/api/v1/${clioIndex.urlSegment}/$queryPath",
+        uri = s"/$apiString/v1/${clioIndex.urlSegment}/$queryPath",
         method = HttpMethods.POST,
         entity = entity
       )
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -177,7 +177,8 @@ class ClioWebClient(
 
   def upsert[CI <: ClioIndex](clioIndex: CI)(
     key: clioIndex.KeyType,
-    metadata: clioIndex.MetadataType
+    metadata: clioIndex.MetadataType,
+    force: Boolean = false
   ): Future[UpsertId] = {
     import clioIndex.implicits._
 
@@ -194,9 +195,13 @@ class ClioWebClient(
       Uri.Path(s"/api/v1/${clioIndex.urlSegment}/metadata")
     )(_ / _)
 
+    val uri =
+      if (force) Uri(path = encodedPath).withQuery(Uri.Query("force" -> "true"))
+      else Uri(path = encodedPath)
+
     dispatchRequest(
       HttpRequest(
-        uri = Uri(path = encodedPath),
+        uri = uri,
         method = HttpMethods.POST,
         entity = entity
       )
@@ -215,6 +220,7 @@ class ClioWebClient(
     input: clioIndex.KeyType,
     includeDeleted: Boolean
   ): Future[Option[clioIndex.MetadataType]] = {
+
     import clioIndex.implicits._
     import s_mach.string._
 
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -195,13 +195,9 @@ class ClioWebClient(
       Uri.Path(s"/api/v1/${clioIndex.urlSegment}/metadata")
     )(_ / _)
 
-    val uri =
-      if (force) Uri(path = encodedPath).withQuery(Uri.Query("force" -> "true"))
-      else Uri(path = encodedPath)
-
     dispatchRequest(
       HttpRequest(
-        uri = uri,
+        uri = Uri(path = encodedPath).withQuery(Uri.Query("force" -> force.toString)),
         method = HttpMethods.POST,
         entity = entity
       )
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
@@ -75,14 +75,12 @@ class ElasticsearchQueryMapper[Input: ClassTag: FieldMapper] {
   private def flattenVals(queryInput: Input): Seq[(String, _)] = {
     val vals = inputMapper.vals(queryInput).toSeq
     val flattened = vals flatMap {
-      case (name, valueOption) =>
-        val tpe = inputMapper.types(name)
-        if (!valueOption.isInstanceOf[Option[_]]) {
-          throw new IllegalArgumentException(
-            s"Unexpected non-Option value '$valueOption' for $name: $tpe"
-          )
+      case (name, value) =>
+        // Currently, this throws when encountering a non-option
+        value match {
+          case opt: Option[_] => opt.map(name -> _)
+          case nonOpt         => Some(name -> nonOpt)
         }
-        valueOption.asInstanceOf[Option[_]].map(name -> _)
     }
     flattened sortBy {
       case (name, _) => name
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
@@ -76,7 +76,6 @@ class ElasticsearchQueryMapper[Input: ClassTag: FieldMapper] {
     val vals = inputMapper.vals(queryInput).toSeq
     val flattened = vals flatMap {
       case (name, value) =>
-        // Currently, this throws when encountering a non-option
         value match {
           case opt: Option[_] => opt.map(name -> _)
           case nonOpt         => Some(name -> nonOpt)
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
@@ -2,21 +2,17 @@ package org.broadinstitute.clio.server.service
 
 import akka.NotUsed
 import akka.stream.Materializer
-import akka.stream.scaladsl.{Sink, Source}
+import akka.stream.scaladsl.Source
+import gnieh.diffson.circe._
+import io.circe.Json
 import io.circe.syntax._
-import io.circe.{Json, Printer}
 import org.broadinstitute.clio.server.dataaccess.elasticsearch._
 import org.broadinstitute.clio.server.exceptions.UpsertValidationException
-import org.broadinstitute.clio.transfer.model.ClioIndex
-import org.broadinstitute.clio.util.generic.{
-  CaseClassMapper,
-  CirceEquivalentCamelCaseLexer
-}
+import org.broadinstitute.clio.transfer.model._
 import org.broadinstitute.clio.util.model.DocumentStatus.Normal
 import org.broadinstitute.clio.util.model.{DocumentStatus, UpsertId}
 
-import scala.concurrent.duration._
-import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.{ExecutionContext, Future}
 
 abstract class IndexService[CI <: ClioIndex](
   persistenceService: PersistenceService,
@@ -46,18 +42,43 @@ abstract class IndexService[CI <: ClioIndex](
     metadata: clioIndex.MetadataType,
     force: Boolean = false
   )(implicit materializer: Materializer): Future[UpsertId] = {
-
-    validateUpsert(indexKey, metadata, force) match {
-      case Right(updatedMetadata) =>
-        persistenceService
-          .upsertMetadata(
-            indexKey,
-            updatedMetadata,
-            documentConverter,
-            elasticsearchIndex
-          )
-      case Left(rejection) => Future.failed(rejection)
-    }
+    // query by key to see if the document already exists
+    queryMetadataForKey(indexKey)
+      .runFold[Either[Exception, Json]](
+        // If the query doesn't return a document and the document status is not set default to Normal.
+        Right(
+          metadata.documentStatus
+            .fold(metadata.withDocumentStatus(Some(Normal)))(_ => metadata)
+            .asJson
+        )
+      ) { (queryResult, _) =>
+        // if we are not forcing the we want to run validation
+        if (!force) {
+          queryResult match {
+            case Right(existingMetadata) =>
+              // validation will check to make sure no fields are being overwritten
+              validateUpsert(existingMetadata, metadata.asJson)
+            case _ => queryResult
+          }
+        } else {
+          queryResult
+        }
+      }
+      .flatMap[UpsertId] {
+        case Right(updateJson) =>
+          updateJson.as[clioIndex.MetadataType].map { updateMetadata =>
+            persistenceService.upsertMetadata(
+              indexKey,
+              updateMetadata,
+              documentConverter,
+              elasticsearchIndex
+            )
+          } match {
+            case Right(upsertId) => upsertId
+            case Left(failure)   => Future.failed(failure)
+          }
+        case Left(rejection) => Future.failed(rejection)
+      }
   }
 
   def queryMetadata(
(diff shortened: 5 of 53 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -43,98 +39,29 @@ class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI])
...
-
-    differentFields.map { key =>
-      import s_mach.string._
-      (
-        key.toSnakeCase(CirceEquivalentCamelCaseLexer),
-        newMetadataValues(key).get,
-        existingMetadataValues(key).get
-      )
-    }
+      }
```
danxmoran, line 65: :tada:

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -109,7 +109,11 @@ class DeleteExecutor[CI <: ClioIndex](deleteCommand: DeleteCommand[CI])
             )
         }
       upsertResponse <- client
-        .upsert(deleteCommand.index)(deleteCommand.key, markedAsDeleted)
+        .upsert(deleteCommand.index)(
+          deleteCommand.key,
+          markedAsDeleted,
+          deleteCommand.force
```
danxmoran: I think we might always want to force here too, since we'll be overwriting the existing document status.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -194,9 +195,13 @@ class ClioWebClient(
       Uri.Path(s"/api/v1/${clioIndex.urlSegment}/metadata")
     )(_ / _)
 
+    val uri =
+      if (force) Uri(path = encodedPath).withQuery(Uri.Query("force" -> "true"))
```
danxmoran: Could we unconditionally add `Uri.Query("force" -> force.toString)`? A bit cleaner

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+      ElasticsearchIndex.EntityIdElasticsearchName,
+    )
+
+  def validateUpsert(
+    indexKey: clioIndex.KeyType,
+    metadata: clioIndex.MetadataType,
+    force: Boolean
+  )(
+    implicit materializer: Materializer
+  ): Either[Exception, clioIndex.MetadataType] = {
```
danxmoran: This is going to need to return a `Future` since it hits ES

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
       elasticsearchIndex
     )
   }
+  private val keysToDrop =
+    Set(
+      ElasticsearchIndex.UpsertIdElasticsearchName,
+      ElasticsearchIndex.EntityIdElasticsearchName,
+    )
```
danxmoran: These should already be getting stripped out by the `ElasticsearchQueryMapper`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+    force: Boolean
+  )(
+    implicit materializer: Materializer
+  ): Either[Exception, clioIndex.MetadataType] = {
+
+    // drop housekeeping and key fields (evaluate only metadata)
+    val dropKeys = keysToDrop ++ indexKey.getClass.getDeclaredFields
+      .map(field => ElasticsearchUtil.toElasticsearchName(field.getName))
+
+    val storedDocsForKey = Await
```
danxmoran: Pls no `Await`, we should chain with `map` / `flatMap` instead.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+    indexKey: clioIndex.KeyType,
+    metadata: clioIndex.MetadataType,
+    force: Boolean
+  )(
+    implicit materializer: Materializer
+  ): Either[Exception, clioIndex.MetadataType] = {
+
+    // drop housekeeping and key fields (evaluate only metadata)
+    val dropKeys = keysToDrop ++ indexKey.getClass.getDeclaredFields
+      .map(field => ElasticsearchUtil.toElasticsearchName(field.getName))
```
danxmoran: I don't trust `getDeclaredFields`... the non-reflection way we use [elsewhere](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala#L222) is to go through `FieldMapper`.

We should probably extract that logic into a common spot.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+
+    val storedDocsForKey = Await
+      .result(
+        searchService
+          .simpleStringQueryMetadata(indexKey.asJson.pretty(Printer.noSpaces))(
+            elasticsearchIndex
+          )
+          .runWith(Sink.seq),
```
danxmoran: `Sink.seq` should be safe in practice, but if we majorly mess up it might collect a huge list in-memory and crash the server. In my client-side streaming branch I used a [stream fold](https://github.com/broadinstitute/clio/blob/dm-client-streaming-DSDEGP-1985/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala#L160) to avoid building up the huge list.

Or, we could lean on the key invariant and use `Sink.headOption`.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+  ): Either[Exception, clioIndex.MetadataType] = {
+
+    // drop housekeeping and key fields (evaluate only metadata)
+    val dropKeys = keysToDrop ++ indexKey.getClass.getDeclaredFields
+      .map(field => ElasticsearchUtil.toElasticsearchName(field.getName))
+
+    val storedDocsForKey = Await
+      .result(
+        searchService
+          .simpleStringQueryMetadata(indexKey.asJson.pretty(Printer.noSpaces))(
```
danxmoran: Ah, if we're JSON-ifying the key anyways then we don't need to use `FieldMapper` to get the names-to-strip-out. We can do something like:
```scala
val keyJson = indexKey.asJsonObject
val fieldsToDrop = keyJson.keys
```

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+            elasticsearchIndex
+          )
+          .runWith(Sink.seq),
+        1.second
+      )
+      .map(
```
danxmoran: Nitpick: better formatting if we use `{` instead of `(`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+        json =>
+          json.mapObject(_.filterKeys({ key =>
```
danxmoran: We don't need the inner braces on `filterKeys`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+          _ => metadata
+      )
+    )
+
+    updatedMetadata
+  }
+
+  private def diffMetadata(
+    existingMetadata: clioIndex.MetadataType,
+    newMetadata: clioIndex.MetadataType
```
danxmoran: Now that we're doing this in the server, I think we can set it up so both of these args come in as `Json`. Then we won't have to do the hacky casts / camel-case conversions.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -23,10 +25,13 @@ abstract class IndexWebService[CI <: ClioIndex](
...
   private[webservice] val postMetadata: Route = {
     pathPrefix("metadata") {
-      pathPrefixKey { key =>
-        post {
-          entity(as[indexService.clioIndex.MetadataType]) { metadata =>
-            complete(indexService.upsertMetadata(key, metadata))
+      parameter('force.as[Boolean] ? false) { force =>
+        pathPrefixKey { key =>
+          post {
+
```
danxmoran: Extre newline

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -194,9 +195,13 @@ class ClioWebClient(
...
+      else Uri(path = encodedPath)
```
pshapiro4broad: any way to avoid calling `Uri()` constructor on two different code paths here? e.g. something like `Uri(path = encodedPath).map(if (force) _.withQuery(...) else _)`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+      .map(field => ElasticsearchUtil.toElasticsearchName(field.getName))
+
+    val storedDocsForKey = Await
+      .result(
+        searchService
+          .simpleStringQueryMetadata(indexKey.asJson.pretty(Printer.noSpaces))(
+            elasticsearchIndex
+          )
+          .runWith(Sink.seq),
+        1.second
```
pshapiro4broad: What happens if the reply doesn't come in 1second?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
jacarey: ok.. no reason to have a force flag on the delete command then right?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
pshapiro4broad: Why `pretty()` here?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+      )
+      .map(
+        json =>
```
pshapiro4broad: can you replace `json => json` with `_` ?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: (I think) it's still needed; it's also used to say "delete the record in Clio even though the record is pointing at nonexistent files"

danxmoran: (A bad backfill run filled a bunch of records with bad paths which were then "deleted", but the files were still sitting around racking up the $$$)

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
pshapiro4broad: It seems like this work (computing `dropKeys` and this filtering) can be done after the checks on line 115 and line 123. Doing it afterwords would make it clearer that the filtering done here doesn't affect the top level collection, just the json contents of the collection. (If that is what it is doing, it's a little hard to tell)

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
jacarey: hmm ok.. i thought I saw them in there but i'll double check.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+        )(
+          // otherwise use the metadata passed in
+          _ => metadata
+      )
+    )
+
+    updatedMetadata
+  }
+
+  private def diffMetadata(
```
pshapiro4broad: It looks like this is a static method. If so, maybe it should be moved to a class more specific to its definition, such as `ClioIndex`. That way if we ever move this check again, we won't have to move this diff operation again.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
danxmoran: It might be because you're going around the existing query API... if they are coming through, I think `SearchService` would be the right place to strip them out based on how we've got things laid out now.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
...
+      newMetadataValues.keySet
+        .intersect(existingMetadataValues.keySet)
+        .filterNot { field =>
+          existingMetadataValues(field).equals(newMetadataValues(field))
+        }
+
+    differentFields.map { key =>
+      import s_mach.string._
+      (
+        key.toSnakeCase(CirceEquivalentCamelCaseLexer),
```
pshapiro4broad: Isn't this `toElasticsearchName()`?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +78,133 @@ abstract class IndexService[CI <: ClioIndex](
       elasticsearchIndex
     )
   }
+  private val keysToDrop =
+    Set(
+      ElasticsearchIndex.UpsertIdElasticsearchName,
+      ElasticsearchIndex.EntityIdElasticsearchName,
+    )
+
+  def validateUpsert(
```
pshapiro4broad: I think this is complicated enough that it needs a comment that describes what the code is doing and why.

pshapiro4broad (commented):
A few comments, nothing major. I think we really need a test for this code, since it's complex enough that we had to redo it. It would probably have to be written as an integration test.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
jacarey: Since we are just diffing json does it make sense to use a library like https://github.com/gnieh/diffson which has circe support.

danxmoran: 👍 from me

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -34,19 +39,46 @@ abstract class IndexService[CI <: ClioIndex](
...
-        updatedMetadata,
-        documentConverter,
-        elasticsearchIndex
-      )
+    metadata: clioIndex.MetadataType,
+    force: Boolean = false
+  )(implicit materializer: Materializer): Future[UpsertId] = {
+    // query by key to see if the document already exists
+    queryMetadataForKey(indexKey)
+      .runFold[Either[Exception, Json]](
```
danxmoran: If we change the output type of `upsertMetadata` to `Source[UpsertId, NotUsed]` then we could do `fold` instead of `runFold` here, and have this all be one big stream. That should let us drop the implicit materializer too.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +99,58 @@ abstract class IndexService[CI <: ClioIndex](
...
+    if (differences.ops.nonEmpty) {
+      Right(newMetadata)
+    } else {
+      val diffs: String = differences.ops.map { op =>
+        op.toString
+      }.mkString("\n")
+      Left(
+        UpsertValidationException(
+          s"""Adding this document will overwrite the following existing metadata:
+               | $diffs. Use 'force=true' to overwrite the existing data.""".stripMargin
```
danxmoran: Indentation slightly off for the 2nd line, not sure what it'll do to the margin-stripper

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +99,58 @@ abstract class IndexService[CI <: ClioIndex](
...
+  def validateUpsert(
+    existingMetadata: Json,
+    newMetadata: Json
+  ): Either[Exception, Json] = {
+    val differences = JsonDiff.diff(existingMetadata, newMetadata, remember = true)
+
+    if (differences.ops.nonEmpty) {
+      Right(newMetadata)
+    } else {
+      val diffs: String = differences.ops.map { op =>
```
danxmoran: Could be `map(_.toString)`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -43,98 +39,29 @@ class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI])
 
     decodedOrError.fold(
       Future.failed, { decoded =>
```
danxmoran: I think we can drop the `{ decoded => ... }` and do `addFiles(webClient, _, addCommand.force)`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -34,19 +39,46 @@ abstract class IndexService[CI <: ClioIndex](
...
+        Right(
+          metadata.documentStatus
+            .fold(metadata.withDocumentStatus(Some(Normal)))(_ => metadata)
+            .asJson
+        )
+      ) { (queryResult, _) =>
+        // if we are not forcing the we want to run validation
+        if (!force) {
+          queryResult match {
+            case Right(existingMetadata) =>
```
danxmoran: Hrm, puzzling over this fold... it doesn't look like it'll guard against > 1 returned metadata. (It'll throw if the > 1 metadatas have different fields, but in the ID fix case the records were an exact match apart from the `entity_id`).

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -34,19 +39,46 @@ abstract class IndexService[CI <: ClioIndex](
...
+              indexKey,
+              updateMetadata,
+              documentConverter,
+              elasticsearchIndex
+            )
+          } match {
+            case Right(upsertId) => upsertId
+            case Left(failure)   => Future.failed(failure)
+          }
+        case Left(rejection) => Future.failed(rejection)
```
danxmoran: We should be able to `flatMap` over the `Either`s to avoid the nested matches / duplicate handling of `Left`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +99,58 @@ abstract class IndexService[CI <: ClioIndex](
       elasticsearchIndex
     )
   }
+
+  def validateUpsert(
+    existingMetadata: Json,
+    newMetadata: Json
+  ): Either[Exception, Json] = {
+    val differences = JsonDiff.diff(existingMetadata, newMetadata, remember = true)
```
danxmoran: 🎉

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +99,58 @@ abstract class IndexService[CI <: ClioIndex](
...
+  }
+
+  def queryMetadataForKey(
+    indexKey: clioIndex.KeyType
+  ): Source[Json, NotUsed] = {
+    val keyJson = indexKey.asJsonObject
+    val fieldsToDrop = keyJson.keys.toSeq
+
+    searchService
+      .simpleStringQueryMetadata(indexKey.asJson.toString())(
```
danxmoran: I'm not sure `toString()` will do what we want. I think `ModelAutoDerivation.defaultPrinter.noSpaces(indexKey.asJson)` will do the trick.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +99,58 @@ abstract class IndexService[CI <: ClioIndex](
...
+      )
+      .fold[Either[Exception, Option[Json]]](Right(None)) { (acc, storedDocs) =>
+        acc match {
+          case Right(None) =>
+            Right(Some(storedDocs.mapObject(_.filterKeys(!fieldsToDrop.contains(_)))))
+          case Right(Some(_)) =>
+            Left(
+              UpsertValidationException(
+                s"""Got > 1 ${clioIndex.name}s from Clio for key:
+                 |${keyJson.toString}""".stripMargin
```
danxmoran: Same comment here about `toString`

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +99,58 @@ abstract class IndexService[CI <: ClioIndex](
...
+    val keyJson = indexKey.asJsonObject
+    val fieldsToDrop = keyJson.keys.toSeq
+
+    searchService
+      .simpleStringQueryMetadata(indexKey.asJson.toString())(
+        elasticsearchIndex,
+        queryConverter
+      )
+      .fold[Either[Exception, Option[Json]]](Right(None)) { (acc, storedDocs) =>
+        acc match {
```
danxmoran: Since the `Left` case just leaves `acc` as-is, we could
```scala
acc.flatMap {
  case None => ???
  case Some(_) => ???
}
```
to save some boilerplate

clio-server/src/main/scala/org/broadinstitute/clio/server/service/SearchService.scala
```diff
@@ -15,6 +16,21 @@ import org.broadinstitute.clio.transfer.model.ClioIndex
   */
 class SearchService private[server] (searchDAO: SearchDAO) {
 
+  /**
+    * Run a simple string query.
+    * @param simpleString The simple json string to query.
+    * @tparam CI The type of the TransferKey to query.
+    * @return The result of the query.
+    */
+  def simpleStringQueryMetadata[Input, CI <: ClioIndex](simpleString: String)(
```
danxmoran: Would it make sense for this to take `Json`, and handle the stringification? The fact that it's needed as a `String` feels like a detail of `SimpleStringQueryDefinition` that might be worth hiding.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
```diff
@@ -64,6 +66,7 @@ trait ClioIndex {
     implicit val mt: ClassTag[MetadataType] = metadataTag
     implicit val qit: ClassTag[QueryInputType] = queryInputTag
     implicit val ke: Encoder[KeyType] = keyEncoder
+    implicit val oke: ObjectEncoder[KeyType] = objectKeyEncoder
```
danxmoran: Is `ObjectEncoder <: KeyEncoder`? If so we can get rid of the existing `keyEncoder` field + implicit.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
jacarey: hmm weird.. it seemed to work fine when i was testing it

danxmoran: TIL! https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Json.scala#L154

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +95,70 @@ abstract class IndexService[CI <: ClioIndex](
...
+    /* Types of differences:
+        Replace - the existing value is different from what we want to set. We ignore nulls. This will overwrite.
+        Remove - the existing value doesn't exist in the new meta data. This will not overwrite.
+        Add - the existing value doesn't exist. This will not overwrite.
+        Move - effectively an Add and Remove. This will not overwrite.
+        Copy - effectively an Add. This will not overwrite.
+     */
+    val replaceOperations = differences.ops
+      .filter(
+        op => op.isInstanceOf[Replace] && op.asInstanceOf[Replace].value != Json.Null
```
danxmoran: I think we can use `flatMap` and a pattern-match to avoid the casting.
```
val replaceOperations: Seq[Replace] = differences.ops.flatMap {
  case replace: Replace => Some(replace).filter(_.value != Json.Null)
  case _ => None
}
```
And then we shouldn't need the cast below, either.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
```diff
@@ -35,7 +35,7 @@ trait ClioIndex {
 
   val queryInputTag: ClassTag[QueryInputType]
 
-  val keyEncoder: Encoder[KeyType]
+  val objectKeyEncoder: ObjectEncoder[KeyType]
```
danxmoran: We could leave this named as it was, and just change the type (doesn't matter to me either way, just a less noisy diff)

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -23,10 +25,12 @@ abstract class IndexWebService[CI <: ClioIndex](
 
   private[webservice] val postMetadata: Route = {
     pathPrefix("metadata") {
-      pathPrefixKey { key =>
-        post {
-          entity(as[indexService.clioIndex.MetadataType]) { metadata =>
-            complete(indexService.upsertMetadata(key, metadata))
+      parameter('force.as[Boolean] ? false) { force =>
```
danxmoran: Yuck, it would take it as a `Symbol`... do you think it'd be worth sticking `"force"` in a constant in the `transfer-model` project, since it's scattered around a few places? We could (hopefully) still use it here with `Symbol(the-string-constant).as[Boolean]`

danxmoran (commented):
Almost there!

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
jacarey: :+1:

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/package.scala
```diff
@@ -0,0 +1,5 @@
+package org.broadinstitute.clio.transfer
+
+package object model {
+  def forceString: String = "force"
```
jacarey: is there a better place for this?

danxmoran: Not yet, but I'd say we should create one. We could make an `ApiConstants` object to hold magic strings like this (and we could toss `"api"` in there too). package objects are usually used for type aliases / implicits.

clio-util/src/main/scala/org/broadinstitute/clio/util/ApiConstants.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.clio.util
```
danxmoran: Not sure this makes sense to get packaged into the `util` project, `transfer-model` is effectively the client<->server API module.

danxmoran (approved):
A final nitpick, but LGTM. Would be nice to have unit tests for the new code, but if it's a pain to set up with our hand-rolled mocks then I'm willing to trust the integration tests for now.

jacarey (commented):
Yeah was going to work on unit tests once I was happy with changes. Any other thoughts on changing upsert to return Stream ? Worth doing?

danxmoran:
I think it'd be worth trying the stream for upsert, it shouldn't be too many changes from what you've got now.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -67,4 +90,69 @@ abstract class IndexService[CI <: ClioIndex](
...
+      .fold[Either[Exception, Option[Json]]](Right(None)) { (acc, storedDocs) =>
+        acc.flatMap {
+          case None =>
+            Right(Some(storedDocs.mapObject(_.filterKeys(!fieldsToDrop.contains(_)))))
+          case Some(_) =>
+            Left(
+              UpsertValidationException(
+                s"""Got > 1 ${clioIndex.name}s from Clio for key:
+                 |${indexKey.asJson
+                     .pretty(ModelAutoDerivation.defaultPrinter)}""".stripMargin
```
danxmoran: Would pull this out of the string interpolation and into a local val, to avoid the weird line-break.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/PersistenceService.scala
```diff
@@ -36,14 +38,14 @@ class PersistenceService private[server] (
     metadata: M,
     documentMapper: ElasticsearchDocumentMapper[Key, M],
     index: ElasticsearchIndex[_]
-  )(implicit ec: ExecutionContext): Future[UpsertId] = {
+  )(implicit ec: ExecutionContext): Source[UpsertId, NotUsed] = {
```
danxmoran: Looking at this more, I think it'd be OK for this to return `Future` and have the calling code in `IndexService` do the `Source.fromFuture` wrapping. There's nothing inherently stream-y about this method so having it return `Source` can be misleading.

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/JsonWebService.scala
```diff
@@ -1,9 +1,22 @@
...
 
 import akka.http.scaladsl.common.{EntityStreamingSupport, JsonEntityStreamingSupport}
+import akka.stream.scaladsl.Flow
+import akka.util.ByteString
 import de.heikoseeberger.akkahttpcirce.ErrorAccumulatingCirceSupport
 import org.broadinstitute.clio.util.json.ModelAutoDerivation
 
+object JsonWebService {
+
+  def singleElememtJsonStreamingSupport: JsonEntityStreamingSupport =
```
danxmoran: This could be a `val`.

It also feels weird to have this one explicit, but the other implicit. I guess it's nice to have the other one implicit since multiple endpoints use it...

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ApiConstants.scala
```diff
@@ -0,0 +1,24 @@
...
+  def apiString: String = "api"
+  def queryString: String = "query"
+  def queryAllString: String = "queryall"
+  def metadataString: String = "metadata"
+  def healthString: String = "health"
+  def versionString: String = "version"
+}
+sealed trait HealthStatus extends EnumEntry
+
+object HealthStatus extends Enum[HealthStatus] {
```
danxmoran, line 18: Nice

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/JsonWebService.scala
jacarey: yeah I mean we could be explicit as well.. just seems a bit verbose.. Maybe a comment explaining that one is default and the other should be used when you don't want an array?

danxmoran (approved):
:shipit:

clio-server/src/main/scala/org/broadinstitute/clio/server/service/SearchService.scala
```diff
@@ -15,6 +16,23 @@ import org.broadinstitute.clio.transfer.model.ClioIndex
   */
 class SearchService private[server] (searchDAO: SearchDAO) {
 
+  /**
+    * Run a simple string query.
+    * @param simpleJson The simple json to query.
+    * @tparam CI The type of the TransferKey to query.
+    * @return The result of the query.
+    */
+  def simpleStringQueryMetadata[Input, CI <: ClioIndex](simpleJson: Json)(
```
danxmoran: After playing with the branch, I don't think we need this method. We can reuse the existing `queryMetadata` as long as we have a `ElasticsearchQueryMapper[clioIndex.KeyType]`, and we already have all the implicits we need to build one of those.

The only change we need to make in the query-mapper is to change `flattenVals` to support non-options:
```scala
private def flattenVals(queryInput: Input): Seq[(String, _)] = {
    val vals = inputMapper.vals(queryInput).toSeq
    val flattened = vals flatMap {
      case (name, value) =>
       // Currently, this throws when encountering a non-option
        value match {
          case opt: Option[_] => opt.map(name -> _)
          case nonOpt         => Some(name -> nonOpt)
        }
    }
    flattened sortBy {
      case (name, _) => name
    }
  }
```

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -22,19 +24,26 @@ abstract class IndexWebService[CI <: ClioIndex](
...
+    implicit val jsonStreamingSupport: JsonEntityStreamingSupport =
+      JsonWebService.singleElememtJsonStreamingSupport
+    pathPrefix(metadataString) {
+      parameter(Symbol(forceString).as[Boolean] ? false) { force =>
+        pathPrefixKey { key =>
+          post {
+            entity(as[indexService.clioIndex.MetadataType]) { metadata =>
+              complete(
+                indexService
+                  .upsertMetadata(key, metadata, force)
```
danxmoran: We can make this return a strict entity to avoid downstream pain with something like:
```scala
private[webservice] val postMetadata: Route = {
    pathPrefix(metadataString) {
      parameter(Symbol(forceString).as[Boolean] ? false) { force =>
        pathPrefixKey { key =>
          post {
            entity(as[indexService.clioIndex.MetadataType]) { metadata =>
              extractRequestContext { ctx =>
                import ctx.materializer

                complete {
                  // Type inference failed for me just trying to return this,
                  // but maybe we can massage it in a different way
                  val fut = indexService
                    .upsertMetadata(key, metadata, force)
                    .runWith(Sink.head[UpsertId])
                  fut
                }
              }
            }
          }
        }
      }
    }
  }
```

danxmoran: Then we don't need the special-case `JsonEntityStreamingSupport`, either

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
```diff
@@ -75,14 +75,12 @@ class ElasticsearchQueryMapper[Input: ClassTag: FieldMapper] {
...
     val vals = inputMapper.vals(queryInput).toSeq
     val flattened = vals flatMap {
-      case (name, valueOption) =>
-        val tpe = inputMapper.types(name)
-        if (!valueOption.isInstanceOf[Option[_]]) {
-          throw new IllegalArgumentException(
-            s"Unexpected non-Option value '$valueOption' for $name: $tpe"
-          )
+      case (name, value) =>
+        // Currently, this throws when encountering a non-option
```
danxmoran: Stale comment (caught in the copy-pasting act!)

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -31,22 +37,53 @@ abstract class IndexService[CI <: ClioIndex](
...
-      metadata.documentStatus.orElse(Some(DocumentStatus.Normal))
-    )
+    metadata: clioIndex.MetadataType,
+    force: Boolean = false
+  ): Source[UpsertId, NotUsed] = {
+    // query by key to see if the document already exists
+    if (!force) {
+      queryMetadataForKey(indexKey)
+        .fold[Either[Exception, clioIndex.MetadataType]](
+          // If the query doesn't return a document and the document status is not set default to Normal.
```
charlesdyan: A little confusing, could use a comma after "is not set"

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -22,19 +25,32 @@ abstract class IndexWebService[CI <: ClioIndex](
...
-    pathPrefix("metadata") {
-      pathPrefixKey { key =>
-        post {
-          entity(as[indexService.clioIndex.MetadataType]) { metadata =>
-            complete(indexService.upsertMetadata(key, metadata))
+    pathPrefix(metadataString) {
+      parameter(Symbol(forceString).as[Boolean] ? false) { force =>
+        pathPrefixKey { key =>
+          post {
+            entity(as[indexService.clioIndex.MetadataType]) { metadata =>
```
charlesdyan, line 32: I don't really understand what is happening in this block or why

charlesdyan: dan explained it

danxmoran: (Probably worth adding a comment about why we force-run the stream here to get a strict `Future` instead of letting `complete` auto-serialize the stream to a chunked entity)

clio-server/src/test/scala/org/broadinstitute/clio/server/service/MockIndexService.scala
```diff
@@ -32,18 +32,25 @@ abstract class MockIndexService[
       elasticsearchIndex,
       clioIndex
     ) {
-  val queryCalls = ArrayBuffer.empty[clioIndex.QueryInputType]
-  val queryAllCalls = ArrayBuffer.empty[clioIndex.QueryInputType]
-  val upsertCalls = ArrayBuffer.empty[(clioIndex.KeyType, clioIndex.MetadataType)]
+
+  val queryCalls: ArrayBuffer[clioIndex.QueryInputType] =
+    ArrayBuffer.empty[clioIndex.QueryInputType]
```
charlesdyan, line 37: Is this just to make intellij be quiet? Have considered doing it myself on passes through here, but the type annotation is so closely mirrored by the invocation that it seems kind of verbose

charlesdyan (approved):
minor comments

PR #239, opened 2018-04-03
DSDEGP-1985 DSDEGP-2158 rewrite clio-client to use streaming

danxmoran:
### Description

Ready for review!

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -10,97 +16,83 @@ import org.broadinstitute.clio.util.generic.{
...
+                s"Field: $field, Old value: $oldVal, New value: $newVal"
+            }.mkString("\n")
+
+            Source.failed(
+              new IllegalArgumentException(
+                s"""Adding this document will overwrite the following existing metadata:
+                   |$diffs
+                   |Use '--force' to overwrite the existing data.""".stripMargin
+              )
+            )
```
danxmoran: This will get much simpler once #238 merges

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -1,132 +1,154 @@
...
+      .flatMapConcat {
+        _.fold(
+          errs => Source.failed(new IllegalStateException(errs.mkString(", "))),
+          paths => Source.single(paths.to[immutable.Iterable])
+        )
+      }
+  }
 
-    val markedAsDeleted = existingMetadata.markDeleted(deleteCommand.note)
+  private def deletePaths(
```
danxmoran: This is also defined in the move executor, so it'd be nice to find a common spot for it. My gut says `IoUtil` but I'm not sure if that class should know about streams... what do people think?

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -54,131 +46,73 @@ class ClioWebClient(
     }
   }
 
-  /**
```
danxmoran: This all felt like overkill for the most common use-case, which is to spin up a web client, send 1 or 2 requests, then tear down.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: (It doesn't seem as weird to have in `IoUtil` if it was defined in terms of `IO` and `fs2.Stream` instead of `Future` and `Source`, so maybe I've just got a weird aversion to Akka proliferation)

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -1,132 +1,154 @@
...
-        logger.info(s"Deleting '$path' in the cloud.")
-        if (ioUtil.deleteGoogleObject(path) != 0) {
-          // We log immediately here because we Future.sequence these together later,
-          // and that operation short-circuits and returns the first error it hits, but
-          // we want messages for all failed deletes to appear in the logs.
-          val err = s"Could not delete '$path' in the cloud."
-          logger.error(err)
-          throw new Exception(err)
+      .fold(Right(Seq.empty[URI]): Either[Seq[Throwable], Seq[URI]]) { (acc, attempt) =>
+        (acc, attempt) match {
```
jacarey: don't think you need (acc, attempt)  => (acc, attempt)

```     
 .fold(Right(Seq.empty[URI]): Either[Seq[Throwable], Seq[URI]]) {
          case (Right(ps), Right(maybePath)) => Right(ps ++ maybePath.toIterable)
          case (Right(_), Left(err))         => Left(Seq(err))
          case (Left(errs), Right(_))        => Left(errs)
          case (Left(errs), Left(err))       => Left(errs :+ err)
      }
```

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -1,132 +1,154 @@
...
-          case ex =>
-            throw new RuntimeException(
-              s"""Failed to delete at least one file associated with $prettyKey on the cloud.
-                 |The record for $prettyKey still exists in Clio. Check the logs to see which
-                 |delete commands failed, and why, then try re-running this command. If this
-                 |can't be done, please contact the Green Team at ${ClioClientConfig.greenTeamEmail}""".stripMargin,
-              ex
-            )
+    // `mapAsync` fails at runtime if constructed with parallelism of 0.
+    if (oldPaths.isEmpty) {
```
jacarey: since this is just setting up a parallelism buffer couldn't we just use oldPaths.size + 1 and avoid this check?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
```diff
@@ -39,25 +36,36 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)
           deliverCommand.destination,
           ArraysExtensions.IdatExtension
         )
+        val movesToPerform = Map(grn -> movedGrnIdat, red -> movedRedIdat).filterNot {
+          case (old, moved) => moved == old
```
jacarey: should be .equals

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -200,59 +134,61 @@ class ClioWebClient(
...
-
-      metadata.fold(throw _, identity)
-    }
   }
 
   private[webclient] def rawQuery(clioIndex: ClioIndex)(
     input: Json,
     includeDeleted: Boolean
-  ): Future[Json] = {
+  ): Source[Json, NotUsed] = {
```
jacarey: Not truly a  raw query since this will get unmarshalled into a QueryInput object right? We will probably add a true raw query endpoint at some point I imagine.

jacarey (commented):
I'm scared of the merge.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
tlangs: I'm in support of pulling it out and putting it into `IoUtil`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: We don't need to, but it makes plugging the pieces together simpler when everything's a stream.

I'm also ~ab~using the streaming constructs to capture & delay side-effects, so we don't accidentally launch a `Future` before it should be run.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
tlangs: Is this going to speed up querying?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
danxmoran: Hrm, it worked to fix a failing test... I think Scala's `==` delegates to `.equals` for reference types, but I'll double-check.

jacarey: Ah ok.. we use .equals in the other delivery executor. If this works that's fine but lets be consistent.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
danxmoran: Hopefully a bit, but the main benefit is simplifying the code here.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -21,22 +21,25 @@ object Dependencies {
...
   private val Slf4jVersion = "1.7.25"
   private val SwaggerUi = "3.1.5"
-  private val TestContainersScalaVersion = "0.15.0"
+  private val TestContainersScalaVersion = "0.16.0"
   private val TypesafeConfigVersion = "1.3.3"
   private val VaultJavaDriverVersion = "3.1.0"
 
   /** Version of Scala to build Clio with. */
-  val ScalaVersion = "2.12.4"
+  val ScalaVersion = "2.12.5"
```
danxmoran, line 40: Had to bump the Scala version to pull in a fix for a bug in `-Xcheckinit`

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/Metadata.scala
```diff
@@ -100,4 +103,43 @@ object Metadata {
...
   }
+
+  /**
+    * Given one of our `Metadata` classes, extract out all fields that
+    * store path-related information into a map from 'fieldName' -> 'path'.
+    *
+    * Used to build a generic before-and-after move comparison to determine
+    * which paths in a metadata object will actually be affected by the move.
+    */
+  def extractPaths[M <: Metadata[M]: ClassTag](metadata: M): Map[String, URI] = {
```
danxmoran, line 114: I think this whole thing could get ripped out & replaced by `JsonDiff`... will play with it

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -52,36 +56,36 @@ object ClioClient extends LazyLogging {
...
+    val resultStreamOrErr = for {
       credentials <- AuthUtil
         .getCredentials(
           ClioClientConfig.accessToken,
           ClioClientConfig.serviceAccountJson
         )
         .leftMap(AuthError.apply)
 
-      webClient = new ClioWebClient(
+      webClient = ClioWebClient(
```
tlangs: Could we put a no-arg constructor in ClioWebClient to automagically fill in all these environment parameters?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -1,136 +1,159 @@
...
+              new IllegalStateException(
+                s"Inconsistent state detected, non-cloud path '$path' registered to $prettyKey"
+              )
+            )
           )
+        } else {
+          Future(Either.catchNonFatal(ioUtil.googleObjectExists(path))).map { attempt =>
+            attempt.flatMap { doesExist =>
+              if (doesExist) {
+                Right(Some(path))
```
tlangs: This should be `Right(Option(path))` just for safety.

tlangs (commented):
I'll have to finish my review tomorrow when my eyes aren't glossing over.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: I disagree! Using `Option(thing)` marks that you think `thing` could be `null`, but if the path was really `null` many other things would have exploded before this point, so I don't want to give the false impression.

Trying to avoid having Zamboni's hyper-paranoid defensive programming styles creep everywhere.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
danxmoran: Yeah, I can pull the config-reading into the `apply` method.

danxmoran: Er, though that'd make it impossible to inject custom hostname / port / https for integration-testing without generating an application.conf. But I can set these values as default args.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: Now that I'm squinting at this though, this is really an abuse of `Option` as a pseudo-`Iterable`. I'll see if I can find a better way to express it.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -1,136 +1,170 @@
...
+            logger.warn(msg)
+            Right(ps)
+          } else {
+            Left(Seq(new IllegalStateException(msg)))
+          }
+        case (Left(errs), (_, true)) => Left(errs)
+        case (Left(errs), (uri, false)) =>
+          Left(
+            errs :+ new IllegalStateException(
+              s"'$uri' associated with $prettyKey does not exist in the cloud."
```
jacarey: Would be nice to extract these identical error messages. (line 148)

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/Executor.scala
```diff
@@ -1,19 +1,45 @@
...
+  /** Extension methods enabling use of for-comprehensions on Akka's `Source` type. */
+  implicit class SourceMonadOps[A, M](val source: Source[A, M]) extends AnyVal {
+
+    /**
+      * Transform every element of the wrapped stream into a new substream using
+      * the given function, an concatenate the resulting substreams in-order.
+      *
+      * Enables using `for` syntax in general.
+      */
+    def flatMap[B](f: A => Source[B, M]): Source[B, M] = source.flatMapConcat(f)
```
jacarey, line 32: OOOO

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/RetrieveAndPrintExecutor.scala
```diff
@@ -1,34 +1,46 @@
...
+        case _: QueryCommand[_] =>
+          /*
+           * For queries we expect a stream of multiple elements, so we inject extra
+           * characters to be sure the printed stream can be parsed as a JSON array
+           * by the caller.
+           */
+          stringify
+            .intersperse(",\n")
+            .prepend(Source.single("["))
+            .concat(Source.single("]"))
```
jacarey: Can't we just use           
```
stringify
            .intersperse("[",",\n", "]")
```

jacarey:
```
          stringify
            .intersperse("[", ",\n", "]")
``` ?

**Review by jacarey — COMMENTED** [review]
Small stuff but mostly :+1:

**Review by jacarey — APPROVED** [review]

File: clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala [hunk 15]
```diff
@@ -1,136 +1,169 @@
...
+          existingMetadata.markDeleted(deleteCommand.note),
+          // Always force because we're purposefully overwriting document status.
+          force = true
+        )
+        .mapError {
+          case ex =>
+            new RuntimeException(
+              s"""Failed to delete the $name record for $prettyKey in Clio.
+                 |The associated files have been deleted in the cloud, but Clio doesn't know.
+                 |Try re-running this command with '--force'.""".stripMargin,
```
tlangs, line 58: This should be the `AddCommand.force` variable

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -34,6 +40,34 @@ trait IoUtil {
...
+  /**
+    * Build a stream which, when pulled, will delete all of the
+    * given cloud objects in parallel.
+    */
+  def deleteCloudObjects(
+    paths: immutable.Iterable[URI]
+  )(implicit ec: ExecutionContext): Source[Unit, NotUsed] = {
+    Source(paths)
+      .mapAsyncUnordered(paths.size + 1) { path =>
+        Future(Either.catchNonFatal(deleteGoogleObject(path)))
```
tlangs, line 52: Why 1 more thread than the size of the iterable?

clio-client/src/test/scala/org/broadinstitute/clio/client/util/IoUtilSpec.scala
```diff
@@ -1,90 +1,147 @@
...
+         |    Update time:            Wed, 24 Jan 2018 23:00:14 GMT
+         |    Storage class:          STANDARD
+         |    Content-Length:         $expectedSize
+         |    Content-Type:           application/octet-stream
+         |    Hash (crc32c):          AAAAAA==
+         |    Hash (md5):             1B2M2Y8AsgTpgAmY7PhCfg==
+         |    ETag:                   CKCYzILa8dgCEAE=
+         |    Generation:             1516834814888992
+         |    Metageneration:         1
+       """.stripMargin
```
tlangs, line 73: What happens if google changes the output of stat?

tlangs (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
danxmoran: The `mapAsync`s fail at runtime if given parallelism 0.

clio-client/src/test/scala/org/broadinstitute/clio/client/util/IoUtilSpec.scala
danxmoran: Our regex will break and this test won't catch it. The proper fix is to stop shelling out to gsutil.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
tlangs: Why? If there's 0 paths, then there should be nothing happening anyways.

danxmoran: ¯\\_(ツ)_/¯ because Akka

PR #242, opened 2018-04-06
Jc doc status fixup

jacarey:
This fixes the case where you have a document status set in the existing document but the incoming document status is null and force is set to true.

----

* Review cycle:

### DIFF
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
@@ -56,16 +56,12 @@ abstract class IndexService[CI <: ClioIndex](
             )
           ) { (_, existingMetadataJson) =>
             //if the incoming document status is not set then set it to the existing status
-            existingMetadataJson
-              .as[clioIndex.MetadataType]
-              .flatMap { existing =>
-                validateUpsert(
-                  existing,
-                  metadata.documentStatus.fold[clioIndex.MetadataType](
-                    metadata.withDocumentStatus(existing.documentStatus)
-                  )(_ => metadata)
-                )
-              }
+            for {
+              existing <- existingMetadataJson.as[clioIndex.MetadataType]
+              validated <- validateUpsert(existing, metadata)
+            } yield {
+              validated.withDocumentStatus(existing.documentStatus)
+            }
           }
       } else {
         //if the incoming document status is not set then set it to the existing status
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
@@ -61,7 +61,9 @@ abstract class IndexService[CI <: ClioIndex](
               .flatMap { existing =>
                 validateUpsert(
                   existing,
-                  metadata.withDocumentStatus(existing.documentStatus)
+                  metadata.documentStatus.fold[clioIndex.MetadataType](
+                    metadata.withDocumentStatus(existing.documentStatus)
+                  )(_ => metadata)
                 )
               }
           }
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
@@ -77,8 +77,8 @@ abstract class IndexService[CI <: ClioIndex](
             //if the incoming document status is not set then set it to the existing status
             existingMetadataJson
               .as[clioIndex.MetadataType]
-              .flatMap { existing =>
-                Right(metadata.withDocumentStatus(existing.documentStatus))
+              .map { existing =>
+                metadata.withDocumentStatus(existing.documentStatus)
               }
           }
         ) { _ =>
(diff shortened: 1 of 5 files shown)


clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -108,7 +133,7 @@ abstract class IndexService[CI <: ClioIndex](
   }
 
   def validateUpsert(
-    existingMetadata: Json,
+    existingMetadata: clioIndex.MetadataType,
     newMetadata: clioIndex.MetadataType
   ): Either[Exception, clioIndex.MetadataType] = {
     val differences = JsonDiff.diff(existingMetadata, newMetadata, remember = true)
```
danxmoran, line 135: Woah, didn't notice this before... `JsonDiff.diff` works with case classes?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -45,26 +45,51 @@ abstract class IndexService[CI <: ClioIndex](
...
+              setDefaultDocumentStatus(metadata)
+            )
+          ) { (_, existingMetadataJson) =>
+            //if the incoming document status is not set then set it to the existing status
+            existingMetadataJson
+              .as[clioIndex.MetadataType]
+              .flatMap { existing =>
+                validateUpsert(
+                  existing,
+                  metadata.withDocumentStatus(existing.documentStatus)
```
danxmoran: Hrm... will this cause silent failures to overwrite document status? i.e. if the existing doc has status `Normal` and we upsert status `Deleted` with `force=false`, this will overwrite `Deleted` -> `Normal` before validation, so validation will pass, but the status-change will fail to be applied.

jacarey: ugh.. good catch.

jacarey: I must be missing a unit test that would catch this.

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
jacarey: yup!

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -45,26 +45,53 @@ abstract class IndexService[CI <: ClioIndex](
...
+          ) { (_, existingMetadataJson) =>
+            //if the incoming document status is not set then set it to the existing status
+            existingMetadataJson
+              .as[clioIndex.MetadataType]
+              .flatMap { existing =>
+                validateUpsert(
+                  existing,
+                  metadata.documentStatus.fold[clioIndex.MetadataType](
+                    metadata.withDocumentStatus(existing.documentStatus)
+                  )(_ => metadata)
```
danxmoran: I think putting off the status-setting until after validation will be simpler, with the same result:
```scala
for {
  existing <- existingMetadataJson.as[clioIndex.MetadataType]
  validated <- validateUpsert(existing, metadata)
} yield {
  validated.withDocumentStatus(existing.documentStatus)
}
```
Since if the new metadata had a different document status from existing, it would have failed validation.

jacarey: much simpler

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -45,26 +45,49 @@ abstract class IndexService[CI <: ClioIndex](
...
-          validateUpsert(existingMetadataJson, metadata)
-        }
-        .flatMapConcat[UpsertId, NotUsed] {
-          case Right(updatedMetadata) =>
-            upsertToStream(indexKey, updatedMetadata)
-          case Left(rejection) => Source.failed(rejection)
+    // query by key to see if the document already exists if force is false
+    val validateOrError =
+      if (!force) {
+        queryMetadataForKey(indexKey)
```
danxmoran: Akka streams are lazy & pluggable, so I think we could extract the query -> fold -> decoding portion into a local `val`. Then do something like:
```scala
val theQuery = queryMetadataForKey(indexKey).fold(...) { existingMetadataJson.as[Metadata] }
val validateOrError =
  if (!force) {
    theQuery.map { existingOrErr =>
      for {
        existing <- existingOrErr
        ... validate ...
      } yield {
        ...
      }
    }
  } else {
    metadata.documentStatus.fold(theQuery)(_ => Source.single(...))
  }
```

danxmoran (approved):

danxmoran:
@jacarey opportunistic fix: some integration tests are failing because of changes from the last PR that ripped a bunch of logic out of the add CLP. Can you change the expected exception types [here](https://github.com/broadinstitute/clio/blob/develop/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/UbamTests.scala#L84) and [here](https://github.com/broadinstitute/clio/blob/develop/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala#L129) to `RuntimeException`?

als364 (approved):

PR #245, opened 2018-04-11
DSDEGP-1985 Don't time out a request while processing its response.

danxmoran:
### Description

In my test query of all-the-ubams, I saw the client hit a completion timeout after 30 seconds (our default response timeout) while in the midst of printing results. As far as I can tell, this is because Akka HTTP sets up its connection streams so that they don't complete until _after_ the entity stream of the HTTP response completes. Since the client was (correctly) still pulling from the response stream after 30 seconds, the HTTP connection failed to complete within 30 seconds, triggering our timeout.

This PR changes the timeout to be less strict. If the server fails to send any sort of response after 30 seconds, the client will still bail out, but once a response comes in the client will pull the response body for as long as it takes.

I tried to reproduce the weird stream-completion setup in a unit test with no success. I also tried to munge our LoadIntegrationSpec to simulate a query that takes longer than the request timeout to print, but lowering the timeout there caused the spec to fail during its mass-upload phase. I made [DSDEGP-2269](https://broadinstitute.atlassian.net/browse/DSDEGP-2269) to track fixing the upsert bottleneck.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -108,7 +108,7 @@ class ClioWebClient private[client] (
     retriedResponse.flatMapConcat { response =>
       if (response.status.isSuccess()) {
         logger.debug(s"Successfully completed request: $request")
-        response.entity.withoutSizeLimit().dataBytes
+        response.entity.withoutSizeLimit().dataBytes.idleTimeout(requestTimeout)
       } else {
         response.entity.dataBytes.reduce(_ ++ _).flatMapConcat { bytes =>
           Source.failed {
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -84,13 +84,13 @@ class ClioWebClient private[client] (
 
     /*
      * Reusable `Source` which will run the given `HttpRequest` over the connection
-     * flow to the Clio server, erroring out early if the request fails to complete
+     * flow to the Clio server, erroring out early if the response fails to arrive
      * within the request timeout.
      */
     val responseSource = Source
       .single(requestWithCreds)
       .via(connectionFlow)
-      .completionTimeout(requestTimeout)
+      .initialTimeout(requestTimeout)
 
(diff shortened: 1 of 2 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -84,13 +84,13 @@ class ClioWebClient private[client] (
...
      * Reusable `Source` which will run the given `HttpRequest` over the connection
-     * flow to the Clio server, erroring out early if the request fails to complete
+     * flow to the Clio server, erroring out early if the response fails to arrive
      * within the request timeout.
      */
     val responseSource = Source
       .single(requestWithCreds)
       .via(connectionFlow)
-      .completionTimeout(requestTimeout)
+      .initialTimeout(requestTimeout)
```
jacarey, line 93: Would it make sense to set an idle timeout as well?

danxmoran: 👍 I'll try it out. I originally tried setting one here but hit the same issue with the HTTP connection going idle after 30 seconds, but it should(?) work if I stick it after the retry logic.

jacarey:
Not sure the specifics of setting up the tests.. but can we make use of Flow delays to try to simulate expected failures?
https://github.com/akka/akka/blob/master/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDelaySpec.scala

danxmoran:
I (think) flow delays will work for testing the idle timeout, but I couldn't get them to meaningfully distinguish between the completion and initial timeouts when I was playing around with it yesterday.

As far as I can tell, the problem is that the `Flow[HttpResponse, HttpRequest, _]`s that we build for testing complete as soon as the single `HttpResponse` goes through, so the completion time of the connection flow == the time of the `HttpResponse`'s passthrough. IRL, Akka's underlying `Flow` implementation for HTTP connections is somehow tied to the entity-body `Source`s of the `HttpResponse`s it emits, so the completion time of the connection flow == the completion time of the `Source` of the last `HttpResponse` emitted. I peered into Akka's implementation to try to replicate, then backed away slowly.

jacarey (approved):

als364 (approved):

tbl3rd:
I don't see how the response can be considered complete until an HttpEntity is drained,
without re-implementing the HTTP stack.

PR #246, opened 2018-04-12
Introduced DataType enum

tlangs:
### Description

----

* Review cycle:


clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ubam/UbamMetadata.scala
```diff
@@ -14,7 +15,7 @@ case class UbamMetadata(
   analysisType: Option[Symbol] = None,
   baitIntervals: Option[Symbol] = None,
   baitSet: Option[Symbol] = None,
-  dataType: Option[Symbol] = None,
+  dataType: Option[DataType] = None,
```
danxmoran, line 18: If you want a 2-for-1, there's a ticket to add DataType to ubam query inputs, too.

danxmoran (approved):

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ubam/UbamMetadata.scala
tlangs: I'll do it.

pshapiro4broad (approved):
Looks OK. Have we verified that all documents in prod are either `WGS` or `Exome`?

tlangs:
We haven't verified yet, but since everything in Zamboni is controlled ontology, I don't think we'll run into issues. We should query elasticsearch before we merge this just in case though.

danxmoran:
Looks like some other stuff has snuck in:
```
{"took":40,"timed_out":false,"_shards":{"total":5,"successful":5,"failed":0},"hits":{"total":1563241,"max_score":0.0,"hits":[]},"aggregations":{"dts":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[
  {"key":"WGS","doc_count":1561565},
  {"key":"RNA","doc_count":612},
  {"key":"Genome","doc_count":207}
]}}}
```

tlangs:
Should we add those to DataType?

danxmoran:
I'd ask KT. The alternative is to ~delete all the non-WGS records from prod, or to~ not make this an enum.

danxmoran:
Er, nevermind about deleting the non-WGS records. To do that we'd also have to track down all the non-WGS upsert records in GCS, which would be a nightmare.

als364 (approved):

clio-util/src/main/scala/org/broadinstitute/clio/util/model/DataType.scala
```diff
@@ -0,0 +1,15 @@
...
+
+sealed trait DataType extends EnumEntry
+
+object DataType extends Enum[DataType] {
+  override val values: IndexedSeq[DataType] = findValues
+
+  case object WGS extends DataType
+  case object Exome extends DataType
+  case object RNA extends DataType
+  case object Genome extends DataType
```
danxmoran: From slack: it shouldn't cause server failures to limit these to just WGS and Exome, so let's do it and prevent any more RNA/Genome entries from getting loaded in.

danxmoran: Before we merge this we should upsert some RNA records to dev, so we can test querying them post-merge.

PR #249, opened 2018-04-17
Cy dsdegp 2213 expose raw json queries

charlesdyan:
### Description

This branch exposes a raw query endpoint and processes all existing queries as raw queries on the server-side. It also rips out handrolled mocks from some of our server unit tests. [JIRA](https://broadinstitute.atlassian.net/browse/DSDEGP-2213)

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -35,6 +36,11 @@ sealed abstract class QueryCommand[CI <: ClioIndex](val index: CI)
   def includeDeleted: Boolean
 }
 
+sealed abstract class RawQueryCommand[CI <: ClioIndex](val index: CI)
+    extends RetrieveAndPrintCommand {
+  def queryInput: File
```
danxmoran: Would name this `queryInputPath` or something else to distinguish it from the non-raw `QueryInput`

danxmoran: Ooo or! Something like this type-fu:
```scala
sealed abstract class QueryCommand[CI <: ClioIndex, QI](val index: CI) extends RetrieveAndPrintCommand {
  def queryInput: QI
}
sealed abstract class RawQueryCommand[CI <: ClioIndex](val index: CI) extends QueryCommand[CI, File]
sealed abstract class SimpleQueryCommand[CI <: ClioIndex](val index: CI) extends QueryCommand[CI, index.QueryInputType]
```
And then the executor can abstract over the type of input.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -54,6 +58,17 @@ object ClioWebClient {
       extends RuntimeException(
         s"Got an error from the Clio server. Status code: $statusCode. Entity: $entityBody"
       )
+
+  def jsonWithDocumentStatus(input: Json): Json = {
+    import ModelAutoDerivation.encodeEnum
+    input.mapObject(
+      obj =>
```
danxmoran: Could replace `obj => obj.add` with `_.add`

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -54,6 +58,17 @@ object ClioWebClient {
...
+        obj.add(
+          "document_status",
```
danxmoran: We use this in enough places, it's probably worth making a constant for (maybe in the `Metadata` companion object?)

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -54,6 +58,17 @@ object ClioWebClient {
       extends RuntimeException(
         s"Got an error from the Clio server. Status code: $statusCode. Entity: $entityBody"
       )
+
+  def jsonWithDocumentStatus(input: Json): Json = {
```
danxmoran: Not sure it makes sense for `ClioWebClient` to own this method, `Metadata` might be a better spot.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -165,12 +180,34 @@ class ClioWebClient private[client] (
...
+      query(clioIndex)(
+        ClioWebClient.jsonWithDocumentStatus(input.asJson)
+      )
+    }
+  }
+
+  def jsonFileQuery[CI <: ClioIndex](clioIndex: CI)(
+    input: File
+  ): Source[Json, NotUsed] = {
+    parse(input.contentAsString) match {
```
danxmoran: `fold` shorter than `match`:
```scala
parse(...).fold(e => Source.failed(...), query(...)(_, raw = true)
```

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/RetrieveAndPrintExecutorSpec.scala
```diff
@@ -83,8 +83,13 @@ class RetrieveAndPrintExecutorSpec extends BaseClientSpec with AsyncMockFactory
 
       val webClient = mock[ClioWebClient]
       // Type annotations needed for scalamockery.
-      (webClient
-        .query(_: ClioWebClient.QueryAux[UbamQueryInput])(_: UbamQueryInput, _: Boolean))
+      (
+        webClient
```
danxmoran, line 89: Scalafmt pls

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
```diff
@@ -176,6 +177,44 @@ class ClioWebClientSpec extends BaseClientSpec with AsyncMockFactory {
...
+    val client = new ClioWebClient(flow, timeout, 0, generator)
+
+    (for {
+      tempFile <- File.temporaryFile()
+      _ = tempFile.writeText(jsonInput.pretty(defaultPrinter))
+    } yield
+      client
+        .jsonFileQuery(index)(tempFile)
+        .runWith(Sink.head)
+        .map(_ => succeed)).get
```
danxmoran: `for`-`yield` wrapped with parens is ugly... since there's only one `flatMap`, I'd just be explicit with using it. I think the call to `.get` will properly resolve without extra parens that way.

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
```diff
@@ -176,6 +177,44 @@ class ClioWebClientSpec extends BaseClientSpec with AsyncMockFactory {
...
+    }
+
+    val generator = mock[CredentialsGenerator]
+    (generator.generateCredentials _).expects().returning(OAuth2BearerToken("fake"))
+
+    val client = new ClioWebClient(flow, timeout, 0, generator)
+
+    (for {
+      tempFile <- File.temporaryFile()
+      _ = tempFile.writeText(jsonInput.pretty(defaultPrinter))
```
danxmoran: I think you can chain the call to `writeText` onto the call to `temporaryFile()`

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -165,12 +180,34 @@ class ClioWebClient private[client] (
     ).via(parserFlow)
   }
 
-  def query[I](clioIndex: ClioWebClient.QueryAux[I])(
+  def preformattedQuery[I](clioIndex: ClioWebClient.QueryAux[I])(
```
danxmoran: Hrm, not sure if "preformatted" is the right description... maybe "simple"? Since the server re-formats it.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
danxmoran: Also, should this take a `DocumentStatus` as input? If not, I'd rename to be clear it's setting status = Normal

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -54,6 +58,17 @@ object ClioWebClient {
       extends RuntimeException(
         s"Got an error from the Clio server. Status code: $statusCode. Entity: $entityBody"
       )
+
+  def jsonWithDocumentStatus(input: Json): Json = {
+    import ModelAutoDerivation.encodeEnum
+    input.mapObject(
+      obj =>
+        obj.add(
```
danxmoran: This will overwrite any existing document status (which I think isn't the current behavior). I think the old way would be something like `Json.obj("status" -> status).deepMerge(obj)`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -282,6 +282,17 @@ abstract class BaseIntegrationSpec(clioDescription: String)
       .write(obj.asJson.pretty(implicitly[Printer]))
```
danxmoran: Would change this to use `writeLocalTmpFile`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -67,6 +68,140 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+         |    }
+         |  }
+         |}
+      """.stripMargin
+    rawQueryTest(
+      rawJsonQuery,
+      runUpsertGvcf(key, metadata),
+      outputs => {
+        outputs should have length 1
+        outputs.headOption should be(Some(expected))
```
danxmoran: I think these two could be collapsed into `outputs should contain only expected`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -67,6 +68,140 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+         |        }
+         |      ]
+         |    }
+         |  }
+         |}
+      """.stripMargin
+    rawQueryTest(
+      rawJsonQuery,
+      setup(),
+      (outputs: Seq[Json]) => {
```
danxmoran: Is the type annotation needed?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -67,6 +68,140 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+        outputs should have length 2
+        outputs should be(Seq(expected, expected))
```
danxmoran: The length check is redundant since you're also checking with `be`

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
```diff
@@ -47,23 +47,40 @@ class HttpElasticsearchDAO private[dataaccess] (
...
-      .sortByFieldAsc(HttpElasticsearchDAO.DocumentScrollSort)
-      .query(queryDefinition)
+      .source(requestBody)
 
     val responsePublisher = httpClient.publisher(searchDefinition)
     Source
       .fromPublisher(responsePublisher)
       .map(_.to[Json])
+      .alsoTo(
+        Sink.foreach(json => logger.info(json.pretty(ModelAutoDerivation.defaultPrinter)))
```
danxmoran: Should remove (or maybe make `debug`? could be useful I guess)

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
```diff
@@ -18,52 +19,36 @@ import scala.reflect.ClassTag
...
-
-  private val keysToDrop =
-    Set(
-      ElasticsearchIndex.UpsertIdElasticsearchName,
-      ElasticsearchIndex.EntityIdElasticsearchName
+    val queryDefinition = boolQuery must queries
+    JsonObject(
+      (
+        elasticsearchQueryObjectName,
+        parse(BoolQueryBuilderFn(queryDefinition).string).toTry.get
```
danxmoran: `fold(throw _, identity)` will save the to-try conversion.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchQueryMapper.scala
```diff
@@ -85,6 +70,18 @@ class ElasticsearchQueryMapper[Input: ClassTag: FieldMapper] {
...
   }
+
+  /**
+    * Munges the query result document into the appropriate form for the query output.
+    *
+    * @param document A query result document.
+    * @return The query output.
+    */
+  def toQueryOutput(document: Json): Json = {
+    document.mapObject(_.filterKeys({ key =>
```
danxmoran: Shouldn't need the parens around the `{ }` in `filterKeys`

danxmoran: Er, alt: `filterKeys(!keysToDrop.contains(_))` (might not work because of the `!`)

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -60,13 +63,28 @@ abstract class IndexWebService[CI <: ClioIndex](
...
+    path(rawQueryString) {
       post {
-        entity(as[indexService.clioIndex.QueryInputType]) { input =>
-          complete(indexService.queryAllMetadata(input))
+        entity(as[JsonObject]) { json =>
+          extractRequestContext { ctx =>
+            import ctx.materializer
+            import ctx.executionContext
+
+            val primedResponse =
```
danxmoran: Would add a comment here explaining why we do this.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/RetrieveAndPrintExecutorSpec.scala
danxmoran: Clarification: I'm whining about scalafmt

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
charlesdyan: Doesn't seem like I can

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
charlesdyan: Where this function is used, it is to set DocumentStatus to normal for queries. Nothing gets overwritten. The old way was to operate on Input classes and call .withDocumentStatus, which uses copy

charlesdyan: by nothing I mean nothing in Clio. I guess the query could be overwritten if someone is explicitly querying Doc Status not normal, but you can use queryAll for that

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -282,6 +282,17 @@ abstract class BaseIntegrationSpec(clioDescription: String)
...
   }
 
+  /**
+    * Validate that a String is valid JSON and write it
+    * to a local temp file if it is. Throw an exception if it isn't.
```
als364: I don't think this actually throws an exception if the json is invalid.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
charlesdyan: oops

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
charlesdyan: Ah, yeah. It used to. Then I decided I didn't want it to. Will update comment

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -165,12 +168,32 @@ class ClioWebClient private[client] (
...
+    }
+  }
+
+  def jsonFileQuery[CI <: ClioIndex](clioIndex: CI)(
+    input: File
+  ): Source[Json, NotUsed] = {
+    parse(input.contentAsString)
+      .fold(
+        e =>
+          Source.failed(new IllegalArgumentException("File must contain valid Json.", e)),
```
danxmoran: Would include the file path in the error message

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
```diff
@@ -176,6 +177,46 @@ class ClioWebClientSpec extends BaseClientSpec with AsyncMockFactory {
...
+    }
+
+    val generator = mock[CredentialsGenerator]
+    (generator.generateCredentials _).expects().returning(OAuth2BearerToken("fake"))
+
+    val client = new ClioWebClient(flow, timeout, 0, generator)
+
+    File
+      .temporaryFile()
+      .map(tempFile => {
```
danxmoran: `(tempFile => {` cleaner as just `{ tempFile =>`

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
```diff
@@ -324,4 +380,23 @@ class ClioWebClientSpec extends BaseClientSpec with AsyncMockFactory {
...
+
+  it should "raise an error if a raw query input file contains invalid JSON" in {
+    val index = ModelMockIndex()
+
+    // If the request reaches flow, we're doing it wrong
+    val flow = Flow[HttpRequest].map(_ => fail)
+
+    val client = new ClioWebClient(flow, timeout, 0, stub[CredentialsGenerator])
+
+    (for {
```
danxmoran: Would be cleaner without the `for` comprehension

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/RetrieveAndPrintExecutorSpec.scala
```diff
@@ -83,8 +83,13 @@ class RetrieveAndPrintExecutorSpec extends BaseClientSpec with AsyncMockFactory
...
+          .simpleQuery(_: ClioWebClient.QueryAux[UbamQueryInput])(
+            _: UbamQueryInput,
+            _: Boolean
+          )
+        )
         .expects(WgsUbamIndex, query, includeDeleted)
         .returning(Source(keys.map(_.asJson)))
```
danxmoran, line 97: We should add a unit test for the JSON-file query input case, too.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/GvcfTests.scala
```diff
@@ -67,6 +68,134 @@ trait GvcfTests { self: BaseIntegrationSpec =>
...
+    recoverToSucceededIf[FailedResponse] {
+      rawQueryTest(
+        rawJsonQuery,
+        Future.successful(()),
+        _ => succeed
+      )
+    }
+  }
+
+  it should "fail when submitting invalid json" in {
```
danxmoran: I don't think we need this as an integration test, since it's a purely client-side failure covered by the unit tests.

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
```diff
@@ -47,23 +47,41 @@ class HttpElasticsearchDAO private[dataaccess] (
...
-      .query(queryDefinition)
+      .source(requestBody)
 
     val responsePublisher = httpClient.publisher(searchDefinition)
     Source
       .fromPublisher(responsePublisher)
       .map(_.to[Json])
+      .alsoTo(
+        Sink
+          .foreach(json => logger.debug(json.pretty(ModelAutoDerivation.defaultPrinter)))
```
danxmoran, line 83: Does it work as `foreach(logger.debug(_.pretty(...)))`?

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -88,18 +89,25 @@ abstract class IndexService[CI <: ClioIndex](
...
-      )
-    )
+    val document = documentConverter.document(indexKey, updatedMetadata)
+    val futureUpsertId = for {
+      _ <- persistenceDAO.writeUpdate(document, elasticsearchIndex)
+      _ <- searchDAO.updateMetadata(document)(elasticsearchIndex)
+    } yield {
+      ElasticsearchIndex.getUpsertId(document)
+    }
+    Source.fromFuture(futureUpsertId)
```
danxmoran: Since the `Source.fromFuture` is just a convenience for the calling code, I think it'd be more clear to push the call there and have this method return a `Future`

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -60,13 +63,34 @@ abstract class IndexWebService[CI <: ClioIndex](
...
-          complete(indexService.queryAllMetadata(input))
+        entity(as[JsonObject]) { json =>
+          extractRequestContext { ctx =>
+            import ctx.materializer
+            import ctx.executionContext
+
+            // we "prime" the response by pulling its head because
+            // otherwise, an initial successful response is sent to the client
+            // along with a stream to pull on no matter what. If we cannot then get a successful
+            // response from elasticsearch, the stream is truncated and an unhelpful exception
```
danxmoran, line 77: Would clarify that the exception happens on the client-side.

danxmoran (commented):
Getting there!

clio-server/src/main/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAO.scala
charlesdyan: doesn't seem to

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ApiConstants.scala
```diff
@@ -8,7 +8,7 @@ object ApiConstants {
   def forceString: String = "force"
   def apiString: String = "api"
   def queryString: String = "query"
-  def queryAllString: String = "queryall"
+  def rawQueryString: String = "rawquery"
```
danxmoran, line 11: Ergh, just occurred to me that ripping this out will break the backfill (it queries with include-deleted=true). We should probably keep the web endpoints until we get the chance to bump the version of the client used there.

danxmoran (approved):

charlesdyan:
Make sure to update clio client in backfill before this is released

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -233,38 +264,44 @@ object ClioCommand extends ClioParsers {
 
   val addPrefix = "add-"
   val queryPrefix = "query-"
```
tlangs: this should be named to `simpleQueryPrefix`

tlangs (approved):

PR #251, opened 2018-04-17
Tl dsdegp 2120 defaults for backfill

tlangs:
### Description

Added defaults for Ubam `dataType` for recovery.

----

* Review cycle:


clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/SemiAutoClioIndex.scala
```diff
@@ -70,6 +73,7 @@ case object WgsUbamIndex
   override val name: String = "WgsUbam"
   override val commandName: String = "wgs-ubam"
   override val elasticsearchIndexName: String = "wgs-ubam"
+  override val defaults: Json = UbamMetadata(dataType = Option(DataType.WGS)).asJson
```
danxmoran: The ticket is for crams & gvcfs, not ubams. data type isn't part of the key for ubams so it's OK for it to be missing on some upserts.

danxmoran: (You'll need to add data type to those metadata classes)

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
```diff
@@ -51,6 +51,8 @@ trait ClioIndex {
 
   val queryInputMapper: FieldMapper[QueryInputType]
 
+  val defaults: Json
```
danxmoran: I'm not convinced this should go into `ClioIndex`. All this code deals with client <-> server API, but this is a detail of recovery buried in the backend. (Though we do have `elasticsearchIndexName` in this trait, but I continue to object to that too).

danxmoran:
Should also add a new unit test to check that defaults are applied (something [like this](https://github.com/broadinstitute/clio/blob/develop/clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala#L110)).

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/SemiAutoClioIndex.scala
danxmoran: Sorry for scope creep, but could you also bundle https://broadinstitute.atlassian.net/browse/DSDEGP-2121 into this PR? Ideally we should only need to rebuild the cram & gvcf indexes once, so it'd be great if we started including data_type in the keys for crams & gvcfs as part of this.

tlangs: Oh wow I misunderstood the ticket. My bad!

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -20,6 +26,7 @@ import org.broadinstitute.clio.util.model.UpsertId
   */
 class ElasticsearchIndex[CI <: ClioIndex](
   val clioIndex: CI,
+  val defaults: Json,
```
tlangs: @danxmoran Is this more what you were envisioning?

danxmoran: Ye!

If we want to constrain it even more, we could make `defaults` be `clioIndex.MetadataType` and then do the `asJson` in the class body. But I think there's some arbitrary restriction on constructor parameters referring to one another that might block it...

tlangs: I think there is. I'm not too happy about defining the actual default values in `ElasticSearchIndex` either. I don't think there should be data in our mapping logic.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/WgsCramMetadata.scala
```diff
@@ -4,8 +4,19 @@ import java.net.URI
...
-import org.broadinstitute.clio.util.model.{DocumentStatus, RegulatoryDesignation}
+import org.broadinstitute.clio.util.json.ModelAutoDerivation
+import org.broadinstitute.clio.util.model.{
+  DataType,
+  DocumentStatus,
+  RegulatoryDesignation
+}
+
+object WgsCramMetadata extends ModelAutoDerivation {
+  val defaults: Json = WgsCramMetadata(dataType = Option(DataType.WGS)).asJson
```
danxmoran: Am not convinced by this either. Having it here with the other client <-> server API stuff is pretty wide-open, and it's easy to think these defaults should be used in spots other than just recovery. Scoping the code to the only spot that it should be used (in the server backend) prevents abuse.

tlangs: I'm concerned that putting it in `ElasticSearchIndex` muddles the definition of the data with the actual values of the data. `ElasticSearchIndex` just defines the indices and doesn't seem like the place to store values.

danxmoran: You could make the same argument about `TransferIndex` / `XYZMetadata`. Ideally they'd all just be declarations of the data's shape and not include any real values, but we need a spot to shim in the defaults. I'd rather keep the shimming as isolated as possible.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +57,19 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
   }
 
+  /**
+    * The ID of a json record in elasticsearch. By default, each record's ID consists
+    * of the concatenated values of the Key fields.
+    * Additional fields can be added to the ID using the ElasticsearchIndex constructor.
+    */
+  def getId(json: Json): String = {
+    // Very hacky way of getting key fields without instantiating dummy keys
+    val keyFields = clioIndex.keyTag.runtimeClass.getConstructors.head.getParameters
```
danxmoran: Would prefer to use the `FieldMapper` for the key type instead of the class tag. It's shapeless-fu instead of reflection.

danxmoran: We should cache this calculation at the instance level too, instead of doing it on every `getId` call.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -20,9 +26,12 @@ import org.broadinstitute.clio.util.model.UpsertId
   */
 class ElasticsearchIndex[CI <: ClioIndex](
   val clioIndex: CI,
-  private[elasticsearch] val fieldMapper: ElasticsearchFieldMapper
+  val defaults: Json = Json.obj(),
+  private[elasticsearch] val fieldMapper: ElasticsearchFieldMapper,
+  private[elasticsearch] val additionalIdFields: Seq[String] = Seq.empty
```
danxmoran: I wonder if we need this, or if we could compute it from `defaults`... get all the keys with non-null values. WDYT?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -93,24 +115,28 @@ object ElasticsearchIndex extends ModelAutoDerivation {
...
       WgsUbamIndex,
+      UbamMetadata().asJson,
       ElasticsearchFieldMapper.StringsToTextFieldsWithSubKeywords
     )
 
   val Gvcf: ElasticsearchIndex[GvcfIndex.type] =
     new ElasticsearchIndex(
       GvcfIndex,
+      GvcfMetadata(dataType = Option(DataType.WGS)).asJson,
       ElasticsearchFieldMapper.StringsToTextFieldsWithSubKeywords
```
danxmoran: Needs to include `data_type` as an additional key field, if we're not deriving it from defaults.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -93,24 +115,28 @@ object ElasticsearchIndex extends ModelAutoDerivation {
...
     )
 
   val WgsCram: ElasticsearchIndex[WgsCramIndex.type] =
     new ElasticsearchIndex(
       WgsCramIndex,
+      WgsCramMetadata(dataType = Option(DataType.WGS)).asJson,
       ElasticsearchFieldMapper.StringsToTextFieldsWithSubKeywords
```
danxmoran: Also needs the additional key field.

clio-dataaccess-model/src/test/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndexSpec.scala
```diff
@@ -11,12 +12,14 @@ class ElasticsearchIndexSpec extends FlatSpec with Matchers with ModelAutoDeriva
   it should behave like aV1Index(
     new ElasticsearchIndex[ModelMockIndex](
       ModelMockIndex("mock"),
+      Map.empty[String, String].asJson,
```
danxmoran: Why not just `Json.obj()`? Here and elsewhere.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
tlangs: What if we have something defaulted that we don't want in the id?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +57,17 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
 
+  /**
+    * The ID of a json record in elasticsearch. By default, each record's ID consists
+    * of the concatenated values of the Key fields.
+    * Additional fields can be added to the ID using the ElasticsearchIndex constructor.
+    */
+  def getId(json: Json): String = {
+    val keyFields = FieldMapper[clioIndex.KeyType].fields.keys
+      .map(snakeCaseTransformation)
+    (keyFields ++ additionalIdFields).map(json.unsafeGet[String]).mkString
```
danxmoran: I'm not sure there's a generic way to decode this... you might need to go through circe's api for working with JSON and handle the cases for string and number explicitly (and I guess throw for all the other cases).

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
danxmoran: Hopefully we'll never do that, but fair enough.

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAOSpec.scala
```diff
@@ -115,23 +121,23 @@ class HttpElasticsearchDAOSpec
...
 
-    def generateBookkeeping(s: String): Json =
+    def generateBookkeeping: Json =
       Map(
         ElasticsearchIndex.UpsertIdElasticsearchName -> UpsertId.nextId()
       ).asJson
-        .deepMerge(
-          Map(ElasticsearchIndex.EntityIdElasticsearchName -> s).asJson
-        )
+        .deepMerge(ModelMockKey(keyLong, keyString).asJson)
```
danxmoran: Not sure if this is the root cause, but reusing the key here will cause problems. All the documents will merge together since they share an ID.

danxmoran: (Where before they all had unique IDs)

tlangs:
@danxmoran The unit and integration tests have been modified to test for this new functionality.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +58,21 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
 
+  /**
+    * The ID of a json record in elasticsearch. By default, each record's ID consists
+    * of the concatenated values of the Key fields.
+    * Additional fields can be added to the ID using the ElasticsearchIndex constructor.
+    */
+  def getId(json: Json): String = {
+    val keyFields = FieldMapper[clioIndex.KeyType].fields.keys
+      .map(snakeCaseTransformation)
+    val foo =
```
danxmoran: Leftover `foo`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +58,21 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
+      (keyFields ++ additionalIdFields).toSeq.sorted
+        .map(f => json.getAsString(f))
```
danxmoran: Should just be able to do `json.getAsString`

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +58,21 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
+  /**
+    * The ID of a json record in elasticsearch. By default, each record's ID consists
+    * of the concatenated values of the Key fields.
+    * Additional fields can be added to the ID using the ElasticsearchIndex constructor.
+    */
+  def getId(json: Json): String = {
+    val keyFields = FieldMapper[clioIndex.KeyType].fields.keys
+      .map(snakeCaseTransformation)
+    val foo =
+      (keyFields ++ additionalIdFields).toSeq.sorted
```
danxmoran: Using `sorted` will change the ordering of the ID pieces for ubams, which we weren't planning on rebuilding as part of this... will poll the group about it.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -20,9 +27,12 @@ import org.broadinstitute.clio.util.model.UpsertId
   */
 class ElasticsearchIndex[CI <: ClioIndex](
   val clioIndex: CI,
-  private[elasticsearch] val fieldMapper: ElasticsearchFieldMapper
+  private[elasticsearch] val defaultFields: Json = Json.obj(),
+  private[elasticsearch] val fieldMapper: ElasticsearchFieldMapper,
+  private[elasticsearch] val additionalIdFields: Seq[String] = Seq.empty
```
danxmoran: I think we should derive this from `defaultFields`, instead of leaving it wide open. The only time we should be shimming in defaults is to modify the `_id` value (otherwise we can use normal upserts), so I'd rather it be locked down.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala
```diff
@@ -56,5 +56,18 @@ object ElasticsearchUtil {
...
+      json.asObject
+        .flatMap(_.apply(key))
+        .map {
+          case s if s.isString =>
+            s.asString.getOrElse("")
+          case a if a.isArray =>
+            a.asArray.map(_.mkString(".")).getOrElse("")
+          case j => j.toString()
+        }
+        .getOrElse("")
```
danxmoran: `getOrElse("")` worries me, it's too easy to silently corrupt data. I'd rather we explode if we get an unexpected value.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala
```diff
@@ -56,5 +56,18 @@ object ElasticsearchUtil {
...
     def unsafeGet[T: Decoder](key: String): T =
       json.hcursor.get[T](key).fold(throw _, identity)
+
+    def getAsString(key: String): String = {
+      json.asObject
+        .flatMap(_.apply(key))
+        .map {
+          case s if s.isString =>
+            s.asString.getOrElse("")
+          case a if a.isArray =>
```
danxmoran: I think we should explode on an array / object field, it doesn't make sense to have it as part of the key.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala
```diff
@@ -56,5 +56,18 @@ object ElasticsearchUtil {
       */
     def unsafeGet[T: Decoder](key: String): T =
       json.hcursor.get[T](key).fold(throw _, identity)
+
+    def getAsString(key: String): String = {
+      json.asObject
+        .flatMap(_.apply(key))
+        .map {
+          case s if s.isString =>
+            s.asString.getOrElse("")
```
danxmoran: Can avoid scattering `getOrElse`s throughout the cases by making this a `flatMap` (and the last case gets wrapped in a `Some`).

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
```diff
@@ -233,7 +252,20 @@ class RecoveryIntegrationSpec
         // Already sorted by construction. Remove null and internal values here too.
```
danxmoran, line 252: Can't comment on it directly, but instead of adding a separate loop check for the expected keys I'd just change `mapper` (above) to inject the default values. Then the existing `should contain` will cover the default-adding (and it'll give a better message than "false was not equal to true" when it fails).

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -20,9 +27,12 @@ import org.broadinstitute.clio.util.model.UpsertId
   */
 class ElasticsearchIndex[CI <: ClioIndex](
   val clioIndex: CI,
-  private[elasticsearch] val fieldMapper: ElasticsearchFieldMapper
+  private[elasticsearch] val defaultFields: Json = Json.obj(),
```
danxmoran: Defaults usually go after non-defaults (otherwise you're forced to pass them to get to the trailing non-defaults, or use named arguments).

clio-server/src/test/scala/org/broadinstitute/clio/server/dataaccess/HttpElasticsearchDAOSpec.scala
```diff
@@ -41,11 +41,13 @@ class HttpElasticsearchDAOSpec
     val indexVersion1: ElasticsearchIndex[_] =
       new ElasticsearchIndex[ModelMockIndex](
         ModelMockIndex("test_index_update_type"),
+        Json.obj(),
```
danxmoran: Shuffling the arg order so this goes after the `ElasticsearchFieldMapper` will let you avoid specifying it here (and everywhere else).

clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala
```diff
@@ -170,7 +168,19 @@ class ServerServiceSpec
...
+
+      upsertedDocs.map(_.mapObject(_.filter {
+        case (k, _) => k != "mock_default_field"
+      })) should contain theSameElementsAs initStoredDocuments
+
+      upsertedDocs
+        .drop(initInSearch)
+        .forall(
+          js => js.asObject.getOrElse(JsonObject.empty).contains("mock_default_field")
+        ) should be(true)
```
danxmoran: Same comment as in the integration tests, I think it'd be better to instead inject `mock_default_field` into the expected documents and then use one `should contain`.

clio-util/src/main/scala/org/broadinstitute/clio/util/json/ModelAutoDerivation.scala
```diff
@@ -117,6 +118,10 @@ trait ModelAutoDerivation extends AutoDerivation {
 
   implicit val decodeSymbol: Decoder[Symbol] =
     Decoder.decodeString.map(Symbol.apply)
+
+  def dropNullsFromJson(json: Json): Json = {
+    parse(json.pretty(implicitly)).getOrElse(Json.obj())
+  }
```
danxmoran: I'd put this in [JsonOps](https://github.com/broadinstitute/clio/blob/develop/clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala#L50) as an extension method, and instead of jumping through the printer I'd use a `filter` [like this](https://github.com/broadinstitute/clio/blob/develop/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala#L225), to avoid dependencies on tangential implicits / to (probably) be more efficient.

clio-client/src/test/scala/org/broadinstitute/clio/client/commands/ClioParsersSpec.scala
```diff
@@ -18,9 +18,13 @@ class ClioParsersSpec extends FlatSpec with Matchers with ClioParsers {
   DocumentStatus.values.foreach {
     it should behave like enumParser(_)
   }
+  DataType.values.foreach {
+    it should behave like enumParser(_)
```
pshapiro4broad: Do we need to test every enum separately here? E.g. is there something special that needs to be done for each enum for this to work, or would it be sufficient to test a single enum?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchDocumentMapper.scala
```diff
@@ -25,10 +25,7 @@ class ElasticsearchDocumentMapper[Key <: IndexKey: Encoder, Metadata: Encoder](
   def document(key: Key, metadata: Metadata): Json = {
     val bookkeeping = Json.fromFields(
       Seq(
-        ElasticsearchIndex.UpsertIdElasticsearchName -> genId().asJson,
-        ElasticsearchIndex.EntityIdElasticsearchName -> key.getUrlSegments
-          .mkString(".")
-          .asJson
+        ElasticsearchIndex.UpsertIdElasticsearchName -> genId().asJson
```
pshapiro4broad: Now that there's only one element do you still need a `Seq`?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala
tlangs: I'm not sure if exploding is the right idea because this code isn't specific to fields used for IDs

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: Not sure what the motivation of using `sorted` here is. Is the intent to alpha sort all key fields for each ID? I think we'd want to have the key fields in the same order for all IDs in an index.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
pshapiro4broad: Also it appears to be unused in this commit; I'm not sure what purpose it serves. What is the case where need a field that's part of the ID but is not part of the key?

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala
danxmoran: We shouldn't be using this method outside of ID generation (now that I'm re-looking at it, I'd even rather move it to ElasticsearchIndex as a private method there). The whole setup of concatenating pieces with `.` and not double-quoting strings is designed to fit the ID-gen use case, and doesn't make sense in other contexts.

pshapiro4broad:
I think we should look at adding dataType as a key field before doing this, so that we don't need temporary code to manually add dataType to the id token.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchUtil.scala
tlangs: that makes sense. I'll move it into EsI.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
tlangs: Since the mapper is used for the actual values as well, wouldn't that not be testing anything?

tlangs:
The scope of this branch has already creeped a lot. I've implemented this in a way that moving the ID from metadata to key shouldn't really change much.

tlangs:
Integration tests pass.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/RecoveryIntegrationSpec.scala
danxmoran: Good point, it shouldn't go in `mapper`. I'd still say injecting the default fields into the expected and doing a single `contains` check is clearer than the double-check setup, and covers more (right now you're testing that the default keys get added to the JSON, but not that the keys have the expected values). I'd tack on a second `map` call in the `val sortedExpected =` statement.

clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala
```diff
@@ -170,7 +167,18 @@ class ServerServiceSpec
...
+        .take(initInSearch)
+
+      upsertedDocs.map(_.mapObject(_.filter {
+        case (k, _) => k != "mock_default_field"
+      })) should contain theSameElementsAs initStoredDocuments
+
+      upsertedDocs
+        .drop(initInSearch)
+        .map(_.asObject.map(_.keys.toSet).getOrElse(Set.empty))
+        .reduce(_ intersect _) should contain("mock_default_field")
```
danxmoran: Same comments as with `mapper` in the recovery spec.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -76,41 +100,58 @@ object ElasticsearchIndex extends ModelAutoDerivation {
...
+    UpsertIdElasticsearchName
   )
 
-  def getEntityId(json: Json): String =
-    json.unsafeGet[String](EntityIdElasticsearchName)
-
   def getUpsertId(json: Json): UpsertId =
     json.unsafeGet[UpsertId](UpsertIdElasticsearchName)
 
+  implicit class IdJson(val json: Json) extends AnyVal {
```
danxmoran: There's no need to use an implicit class for this; `getAsString` should only ever be used in `getId` above, so a private method would do (or even inlining the logic in the `map` call).

clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala
```diff
@@ -132,29 +134,24 @@ class ServerServiceSpec
       Some(URI.create("/mock")),
       Some(1L)
```
danxmoran, line 136: We should also add a test with `mockDefaultField` set to something other than `"mock-default"`, and check that it doesn't get overwritten by recovery.

danxmoran (commented):
Getting there!

danxmoran:
The scope creep was an accident of dependencies not being clear between tickets. I think the ideal thing to do would be to extract out the bits for 2121 (killing entity_id) into a separate PR to merge on its own, and save adding data_type until we can do the right thing with adding it to a new key, with new APIs. Then we can lock down the default-injection to only operate on key values and save ourselves from adding more `TODO delete this later` code.

clio-server/src/test/scala/org/broadinstitute/clio/server/service/ServerServiceSpec.scala
tlangs: Done

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +56,33 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
+          case a if a.isArray || a.isObject =>
+            throw new RuntimeException("Arrays and objects cannot be used as ID fields")
+          case j => Option(j.toString())
+        }
+        .getOrElse(throw new RuntimeException(s"Could not get $key from json"))
+    }
+
+    val keyFields = FieldMapper[clioIndex.KeyType].fields.keys
+      .map(snakeCaseTransformation)
+    (keyFields ++ defaults.asObject.map(_.keys).getOrElse(Iterable.empty)).toSeq.sorted
```
danxmoran: The more I think about it, the more convinced I am we shouldn't do `sorted`. It'll break the existing ubam index. Last time we rebuilt it it took the better half of a day, and we've backfilled significantly more records since then. We shouldn't take Clio out for a full day unless we really need to.

danxmoran (changes requested):
I also still think we should split this in 2: A merge-able PR to remove `entity_id`, then a follow-up to add `data_type` to the keys for crams and gvcfs, w/ defaults set on recovery & API migration. @ApChagi's got a PR for migrating the ubam APIs at #254 that we could mimic.

clio-dataaccess-model/src/main/scala/org/broadinstitute/clio/server/dataaccess/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -48,6 +56,33 @@ class ElasticsearchIndex[CI <: ClioIndex](
...
+
+    def getAsString(key: String): String = {
+      json.asObject
+        .flatMap(_.apply(key))
+        .flatMap {
+          case s if s.isString =>
+            s.asString
+          case a if a.isArray || a.isObject =>
+            throw new RuntimeException("Arrays and objects cannot be used as ID fields")
+          case j => Option(j.toString())
```
pshapiro4broad: Not sure what the purpose of this is, you can call `toString` on any object (other than `null` of course).

tlangs: Calling `toString` on a JSON String results in the quotes being included in the string. Other JSON data types don't have this issue.

pshapiro4broad: OK, and `Option()` is required because `toString` might return `null`?

tlangs: `Option` is just to conform to the flatMap. `Json.asString` returns an `Option`, so we need to return an `Option` here too.

danxmoran: Re-stating pet peeve about using `Option` instead of `Some` when you know the thing can't be `null`

danxmoran (dismissed):
The epic saga completes!

danxmoran:
Oh, but there be merge conflicts. Will re-look after they're resolved.

tlangs:
Already fixed.

danxmoran (approved):

pshapiro4broad (approved):

PR #253, opened 2018-04-24
DSDEGP-1770: replace gsutil with java api

pshapiro4broad:
DSDEGP-1770: replace gsutil shell subprocess calls with native java API

- remove unused APIs and tests from IoUtil
- fix error reporting code so exception types, stack crawls aren't lost
- update readme to match current integration test names

----

* Review cycle:


clio-client/src/test/scala/org/broadinstitute/clio/client/util/IoUtilSpec.scala
```diff
@@ -1,101 +1,37 @@
...
+  class TestIoUtil extends IoUtil {
 
-  it should "read a metadata file from file location" in {
-    val contents = "I'm a file!"
-
-    File.temporaryFile() { f =>
-      IoUtil.readMetadata(f.write(contents).uri) should be(contents)
-    }
-  }
+    override def readGoogleObjectData(location: URI): String = { " " }
```
tlangs: The braces here are unnecessary

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -54,163 +52,94 @@ trait IoUtil {
...
-    }
-  }
+  def copyGoogleObject(from: URI, to: URI): Unit
 
-  def googleObjectExists(path: URI): Boolean = {
-    gsUtil.exists(path.toString) == 0
-  }
+  def deleteGoogleObject(path: URI): Unit
 
-  def getMd5HashOfGoogleObject(path: URI): Option[Symbol] = {
```
tlangs: Other projects depend on md5 hashes.

tlangs (changes requested):

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -54,163 +52,94 @@ trait IoUtil {
...
-    * TODO: We should consider moving this to google-cloud-java
-    * api usage instead of gsutil. We pay significant overhead
-    * on the startup time of every gsutil call.
-    */
-  class GsUtil {
-
-    def ls(path: String): Seq[String] = {
-      runGsUtilAndGetStdout(Seq("ls", path)).split("\n")
-    }
+class GsUtil(credentials: Credentials) extends IoUtil {
```
danxmoran: Would still be a fan of what we talked about earlier:
1. Fold `GsUtil` into `IoUtil`
2. Pass a `Storage` instance into the `IoUtil` constructor
3. Move the existing `Credentials -> Storage` logic into an `apply` method on `IoUtil`

Once `Storage` is exposed in the constructor we can mock it out for unit tests, and get that coverage % up. The `apply` logic would be the only bit leaned on by the integration tests.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -54,163 +52,94 @@ trait IoUtil {
...
+  override def copyGoogleObject(from: URI, to: URI): Unit = {
+    getBlob(from).copyTo(toBlobId(to))
+    ()
+  }
 
-    def stat(path: String): String = {
-      runGsUtilAndGetStdout(Seq("stat", path))
+  override def deleteGoogleObject(path: URI): Unit = {
+    if (!getBlob(path).delete()) {
+      throw new RuntimeException(s"Cannot delete '$path' because it does not exist.")
```
danxmoran: Maybe an `IllegalArgumentException` instead, to be more specific?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -54,163 +52,94 @@ trait IoUtil {
...
+  override def writeGoogleObjectData(data: String, location: URI): Unit = {
+    getBlob(location) match {
+      case blob: Blob =>
+        val channel = blob.writer()
+        try {
+          channel.write(ByteBuffer.wrap(data.getBytes()))
+        } finally {
+          channel.close()
+        }
+      case _ =>
```
danxmoran: What value could `_` be here? If it's `null`, then `getBlob` should be changed to return an `Option`.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -54,163 +52,94 @@ trait IoUtil {
...
-      File
-        .temporaryDirectory("gsutil-state")
-        .map { tmp =>
-          runner(Process(Seq("gsutil", "-o", s"GSUtil:state_dir=$tmp") ++ gsUtilArgs))
-        }
-        .get()
+  override def googleObjectExists(path: URI): Boolean = {
+    getBlob(path) match {
+      case blob: Blob => blob.exists()
+      case _          => false
```
danxmoran: Same comment about `getBlob`'s return type. With an `Option` this'd be pretty concise.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorSpec.scala
```diff
@@ -166,7 +167,9 @@ class DeleteExecutorSpec extends BaseClientSpec with AsyncMockFactory {
       executor.execute(webClient, ioUtil).runWith(Sink.ignore)
     }.map { ex =>
       paths.foreach { uri =>
-        ex.getMessage should include(uri.toString)
+        val sw = new StringWriter
+        ex.printStackTrace(new PrintWriter(sw))
```
danxmoran: Could move the printing out of the `foreach`

clio-client/src/test/scala/org/broadinstitute/clio/client/util/IoUtilSpec.scala
```diff
@@ -104,44 +40,56 @@ class IoUtilSpec extends BaseClientSpec with AsyncMockFactory {
...
+    val stream = gsutil.deleteCloudObjects(urisToFail + uriToSucceed)
 
-    recoverToExceptionIf[IOException](stream.runWith(Sink.ignore)).map { ex =>
-      urisToFail.foreach { uri =>
-        ex.getMessage should include(uri)
+    for {
+      _ <- recoverToExceptionIf[IOException](stream.runWith(Sink.ignore)).map { ex =>
+        urisToFail.foreach { uri =>
+          val sw = new StringWriter
+          ex.printStackTrace(new PrintWriter(sw))
```
danxmoran: Could move the printing out of the `foreach`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
```diff
@@ -59,10 +60,25 @@ abstract class BaseIntegrationSpec(clioDescription: String)
     */
   def clioWebClient: ClioWebClient
 
+  private lazy val googleCredential: Credentials =
```
danxmoran, line 63: An `IoUtil.apply` taking whatever type `serviceAccount` is would save us from needing to duplicate this logic.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
pshapiro4broad: There is an API to get it, so it's easy to implement if we need it. What is using it? I looked at backfill and workflow_tests and didn't see it used there. It's not used in clio. The backfill does override this method in `NioUtil` but it's not called, as far as I can tell.

tlangs: I thought the backfill uses it when backfilling wgs ubams. maybe we stopped doing that?

danxmoran (commented):
Oops forgot to submit the review

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
danxmoran: There's a separate API that gets existence, size, and md5 all at once.

Either way I think it's OK to remove things here that the main clio code doesn't use, as long as the backfill code can extend `IoUtil` with new methods.

pshapiro4broad: The code in backfill uses `getGoogleObjectInfo()`, which is implemented using the storage API, not gsutil or the NIO API. That seems OK although there are some small overlaps with the code here and the code in backfill; we can decide what to do about backfill in a later checkin. Once this settles down I'll see about updating backfill.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
pshapiro4broad: OK now that it's working I'll give this a try.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
pshapiro4broad: OK. Actually I'm unsure how this can occur in practice since `getBlob()` seems to return null if an object doesn't exit.

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
pshapiro4broad: Yeah, I thought about that but didn't want to change `read` to return `Option[String]`, etc. I'll look at it again.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutorSpec.scala
pshapiro4broad: good point, `ex` doesn't change inside the  `foreach`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -47,7 +47,7 @@ class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI]) extends Executor
   ): Source[addCommand.index.MetadataType, NotUsed] = {
     Source
       .single(location)
-      .map(ioUtil.readMetadata)
+      .map(ioUtil.readFileData)
```
danxmoran, line 50: This change busted the expectations for the `IoUtil` mock in the unit tests.

project/src/main/scala/org/broadinstitute/clio/sbt/Dependencies.scala
```diff
@@ -35,6 +35,7 @@ object Dependencies {
   private val TestContainersScalaVersion = "0.16.0"
   private val TypesafeConfigVersion = "1.3.3"
   private val VaultJavaDriverVersion = "3.1.0"
+  private val GoogleCloudStorageVersion = "1.14.0"
```
danxmoran: Pls put in ABC order

danxmoran (approved):
Final nitpicks, LGTM once unit tests are fixed up

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/BaseIntegrationSpec.scala
pshapiro4broad: I think this is the only code that does this, at least before calling `IoUtil()`.

tlangs (approved):

PR #256, opened 2018-04-27
DSDEGP-1770 Follow-up fixes to using the SDK for I/O

danxmoran:
### Description

`copyTo` returns a `CopyWriter`. Based on the [javadoc](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java#L544) it sounded like the copy might not actually complete unless [`getResult`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/CopyWriter.java#L74) is called on the writer. Based on [this](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/CopyWriter.java#L42) ~I think our tests didn't catch any truncated copies because the integration tests only run against a single bucket, so all copies complete in a single call.~

EDIT:  I manually tested moving a cram between buckets & projects and it worked fine without `getResult`, so I'm not sure what qualifies as "separate locations" requiring multiple RPC calls.

A different problem cropped up in the meantime, causing nightly tests to [fail](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/zamboni-nightly-workflows-test/112/). Right now we use the `GoogleCredential` built to talk past the openIDC proxy to also build our `StorageOptions`. This fails because the openIDC credential is missing the required auth scopes for talking to storage. We hit the same problem while porting the backfill to NIO, and the integration tests didn't catch it because they have their own logic for setting up credentials.

While fixing the problem, I took the chance to trim a lot of cruft from our auth-related code. I also tried to push more logic into places covered by the integration tests. The [nightly](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/dsde-pipelines-nightly-data-delivery-test/239/) [tests](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/zamboni-nightly-workflows-test/111/) pass when using this branch.

----

* Review cycle:


clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/AuthRefreshTests.scala
```diff
@@ -1,58 +0,0 @@
...
-import scala.concurrent.duration._
-import scala.util.Success
-
-trait AuthRefreshTests { self: EnvIntegrationSpec =>
-
-  behavior of "Clio Client"
-
-  it should "refresh the auth token after it expires" in {
-
-    googleCredential.addChangeListener(
```
danxmoran, line 20: All the shuffling around of auth-related code made it harder to access the `googleCredential` for this test. I had something nearly working, but something about the clock-shifting started causing the gcloud web services to return errors warning that the requester's clock wasn't accurate (which was true... just not sure why we haven't seen it before). Since the code on our side that this test was meant to cover is also [fully covered](https://codecov.io/gh/broadinstitute/clio/src/803f4087eab79376f9594531718d36412de245f6/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/GoogleCredentialsGenerator.scala) by unit tests, I thought it'd be OK to just scrap this. I can try restoring it if others disagree.

charlesdyan (approved):

als364 (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -91,7 +91,7 @@ class IoUtil(storage: Storage) {
   }
 
   def copyGoogleObject(from: URI, to: URI): Unit = {
-    requireBlob(from).copyTo(toBlobId(to))
+    requireBlob(from).copyTo(toBlobId(to)).getResult
```
pshapiro4broad, line 94: DOH

PR #263, opened 2018-05-09
Make ordering of paths-to-delete deterministic.

danxmoran:
### Description

This should prevent spurious test failures.

Making scalamock ignore `Iterable` ordering is doable, but ugly. We don't actually rely on the map-async being unordered here because the next step in the stream is a `fold`, so we can't progress until all of the paths are checked anyways. We also set up the `mapAsync` with parallelism `paths + 1` so all the checks are guaranteed to launch at the same time.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
@@ -139,7 +139,7 @@ class DeleteExecutor[CI <: ClioIndex](deleteCommand: DeleteCommand[CI])(
 
     val paths = metadata.pathsToDelete.to[immutable.Iterable]
     Source[URI](paths)
-      .mapAsyncUnordered(paths.size) { path =>
+      .mapAsync(paths.size) { path =>
         Future(path -> ioUtil.googleObjectExists(path))
       }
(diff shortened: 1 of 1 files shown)


jacarey (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -139,7 +139,7 @@ class DeleteExecutor[CI <: ClioIndex](deleteCommand: DeleteCommand[CI])(
 
     val paths = metadata.pathsToDelete.to[immutable.Iterable]
     Source[URI](paths)
-      .mapAsyncUnordered(paths.size) { path =>
```
als364, line 142: Why were we doing this in the first place?

als364 (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: The paths don't need to be reassembled in-order after this stage of the stream for the mainline code to work (scalamock is really being too strict).

jacarey: supposedly scalamock as an implicit inAnyOrder but that seems to not be the case with AsyncMockFactory vs MockFactory

PR #267, opened 2018-05-14
Added amend functionality

tlangs:
### Description

Implemented the changes discussed in office hours. The new `{index}-amend` client command takes a file with metadata in it and applies that metadata to every document in the index. If a document already has that field, then the original value is left intact. If a document has all metadata defined, then it is skipped and not upserted again. This PR includes both unit and integration tests!

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -144,6 +149,13 @@ final case class DeleteGvcf(
   force: Boolean = false
 ) extends DeleteCommand(GvcfIndex)
 
+@CommandName(ClioCommand.amendGvcfName)
+final case class AmendGvcf(
+  metadataLocation: URI
+) extends AmendCommand(GvcfIndex) {
+  override def queryInput: GvcfQueryInput = GvcfQueryInput()
```
danxmoran: Here and elsewhere, `queryInput` should be in the parameter list, annotated with `@Recurse`. No default value needed.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AddExecutor.scala
```diff
@@ -17,7 +13,7 @@ import org.broadinstitute.clio.transfer.model.ClioIndex
   * it to the clio-server, optionally overwriting existing fields.
   */
 class AddExecutor[CI <: ClioIndex](addCommand: AddCommand[CI]) extends Executor {
-  import addCommand.index.implicits._
+  import Executor.readMetadata
```
danxmoran: Doesn't need to be an import since it's only used in one place.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+import org.broadinstitute.clio.client.commands.AmendCommand
+import org.broadinstitute.clio.client.util.IoUtil
+import org.broadinstitute.clio.client.webclient.ClioWebClient
+import org.broadinstitute.clio.transfer.model.ClioIndex
+import org.broadinstitute.clio.JsonUtils.JsonOps
+
+class AmendExecutor[CI <: ClioIndex](amendCommand: AmendCommand[CI]) extends Executor {
+
+  import amendCommand.index.implicits._
+  import Executor.{SourceMonadOps, readMetadata}
```
danxmoran: Don't need to import `readMetadata` here.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+    // zip new metadatas with documents
+    newMetadataSource
+      .zip(
+        webClient
+          .simpleQuery(amendCommand.index)(queryIndexInput, includeDeleted = false)
+          .map(_.dropNulls)
+      )
+      // Munge metadata for documents that have partial data
+      .map { tup =>
+        val (newMetadataJson, documentJson) = tup
```
danxmoran: Can cut out the `tup` midpoint with `.map { case (newMetadataJson, documentJson) => ... }`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+          documentJson.asObject.map(_.keys.toSeq).getOrElse(Seq.empty)
+        val newMetadataMinusExisting =
+          newMetadataJson.mapObject(_.filterKeys(key => !existingKeys.contains(key)))
+        (newMetadataMinusExisting, documentJson)
+      }
+      // filter out things that already have no new metadata to upsert
+      .filter { tup =>
+        val (newMetadataJson, _) = tup
+        newMetadataJson.asObject.exists(obj => !obj.isEmpty)
+      }
```
danxmoran: Why do you need this whole dance instead of just `documentJson.deepMerge(newMetadataJson)`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+      // Upsert the stuff!
+      .flatMapConcat { tup =>
+        val (newMetadataJson, documentJson) = tup
```
danxmoran: Can also skip the `tup`, as above.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+        val key = getKeyFromDocument(documentJson)
+        val metadata = newMetadataJson
+          .as[amendCommand.index.MetadataType]
+          .fold(
+            ex =>
+              throw new RuntimeException(
+                "Could not encode new metadata json into class",
+                ex.getCause
```
danxmoran: Would just do `ex`, not `ex.getCause`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+        val (newMetadataJson, _) = tup
+        newMetadataJson.asObject.exists(obj => !obj.isEmpty)
+      }
+      // Upsert the stuff!
+      .flatMapConcat { tup =>
+        val (newMetadataJson, documentJson) = tup
+        val key = getKeyFromDocument(documentJson)
+        val metadata = newMetadataJson
+          .as[amendCommand.index.MetadataType]
+          .fold(
```
danxmoran: If you import `cats.syntax.either._` you can use `valueOr` instead of `fold`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+        webClient.upsert(amendCommand.index)(key, metadata, force = false)
+
+      }
+  }
+
+  private def getKeyFromDocument(json: Json): amendCommand.index.KeyType = {
+    val keyJson =
+      json.mapObject(_.filterKeys(amendCommand.index.keyFieldNames.contains))
+
+    val key: amendCommand.index.KeyType = keyJson.as[amendCommand.index.KeyType] match {
```
danxmoran: I think this `match` is also a `valueOr`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+      case Left(ex: DecodingFailure) =>
+        throw new RuntimeException(
+          s"Could not create ${amendCommand.index.keyTag} from ${json.pretty(implicitly)}",
+          ex.getCause
```
danxmoran: `ex`, no `getCause`

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private def getKeyFromDocument(json: Json): amendCommand.index.KeyType = {
+    val keyJson =
+      json.mapObject(_.filterKeys(amendCommand.index.keyFieldNames.contains))
+
+    val key: amendCommand.index.KeyType = keyJson.as[amendCommand.index.KeyType] match {
+      case Left(ex: DecodingFailure) =>
+        throw new RuntimeException(
+          s"Could not create ${amendCommand.index.keyTag} from ${json.pretty(implicitly)}",
```
danxmoran: Would use `defaultPrinter` explicitly

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/Executor.scala
```diff
@@ -42,4 +47,32 @@ object Executor {
       */
     def withFilter(p: A => Boolean): Source[A, M] = source.filter(p)
   }
+
+  /**
+    * Build a stream which, when pulled, will read JSON from a URI and decode it
+    * into the metadata type associated with `index`.
+    */
+  def readMetadata[CI <: ClioIndex](index: ClioIndex)(
```
danxmoran: Why in `Executor` instead of in `IoUtil`?

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AmendExecutorSpec.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  private val zipped = keys.zip(metadatas)
+
+  val newMetadata = ArraysMetadata(sampleAlias = Some("amended_sample_alias"))
+
+  private def setupQueryReturn(
+    webClient: ClioWebClient,
+    documents: immutable.Iterable[(ArraysKey, ArraysMetadata)]
+  ): Unit = {
+    val _ = (
```
danxmoran: Do you need the `val _ =`? I don't think we use it elsewhere when scala-mocking.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/AmendExecutor.scala
```diff
@@ -0,0 +1,89 @@
...
+    newMetadataSource
+      .zip(
+        webClient
+          .simpleQuery(amendCommand.index)(queryIndexInput, includeDeleted = false)
+          .map(_.dropNulls)
+      )
+      // Munge metadata for documents that have partial data
+      .map { tup =>
+        val (newMetadataJson, documentJson) = tup
+        val existingKeys =
```
charlesdyan: can we call this existing fields and fields since we use key to refer to the clio key just a few lines down?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -1192,4 +1193,69 @@ trait ArraysTests { self: BaseIntegrationSpec =>
...
+
+      val storedDocument1 = amended1.headOption.getOrElse(fail)
+      storedDocument1.unsafeGet[String]("chip_type") should be("amended_chip_type")
+      storedDocument1.unsafeGet[String]("sample_alias") should be("amended_sample_alias")
+
+      val storedDocument2 = amended2.headOption.getOrElse(fail)
+      storedDocument2.unsafeGet[String]("chip_type") should be("existing_chip_type")
+      storedDocument2.unsafeGet[String]("sample_alias") should be("amended_sample_alias")
+    }
+  }
```
danxmoran: Should also have a test using the `queryInput` options.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
```diff
@@ -21,8 +22,12 @@ trait ClioIndex {
 
   val commandName: String
 
+  val keyFieldNames: Seq[String]
```
danxmoran: I don't think we need this here, we can always get it from the `keyMapper`.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
```diff
@@ -21,8 +22,12 @@ trait ClioIndex {
...
+
   type KeyType <: IndexKey
 
+  val keyTypeTag: TypeTag[KeyType]
```
danxmoran: Why do we need this?

danxmoran (commented):
Didn't look closely at the unit tests, will check again after main-code review changes.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/ClioIndex.scala
tlangs: Its to get the field names of they keys. You mentioned there's a way to get it from the `keyMapper`?

danxmoran: `keyMapper.fields.keys` will give you camel-cased versions, which you can then snake-case-ify.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AmendExecutorSpec.scala
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  it should "not amend documents that already have metadata defined" in {
+    val ioUtil = mock[IoUtil]
+    (ioUtil.readFileData _)
+      .expects(loc)
+      .returning(newMetadata.asJson.pretty(defaultPrinter))
+
+    val webClient = mock[ClioWebClient]
```
charlesdyan: If you have the spec extend OneInstancePerTest, you can share a single mock between the tests without worrying about expectations from one test interfering with another.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AmendExecutorSpec.scala
```diff
@@ -0,0 +1,174 @@
...
+          )
+        )
+        .expects(ArraysIndex, theKey, metadata, false)
+        .returning(Source.single(UpsertId.nextId().asJson))
+    }
+  }
+
+  it should "not amend documents that already have metadata defined" in {
+    val ioUtil = mock[IoUtil]
+    (ioUtil.readFileData _)
```
charlesdyan: this could be a method that takes metadata to return to reduce duplicate code

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -1192,4 +1193,69 @@ trait ArraysTests { self: BaseIntegrationSpec =>
       storedDocument1.unsafeGet[String]("notes") should be("I'm a note")
     }
   }
+
+  it should "amend documents with new metadata" in {
+
+    val metadataFile: File = File
+      .newTemporaryFile()
+      .overwrite(
```
charlesdyan: Why overwrite instead of just write?

charlesdyan (approved):
Couple maybe suggestions, but lgtm

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AmendExecutorSpec.scala
tlangs: It was complaining about discarding a non-unit value.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/Executor.scala
pshapiro4broad: If it was a method on `IoUtil` then it wouldn't need to accept it as an argument.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -1192,4 +1193,69 @@ trait ArraysTests { self: BaseIntegrationSpec =>
       storedDocument1.unsafeGet[String]("notes") should be("I'm a note")
     }
   }
+
+  it should "amend documents with new metadata" in {
```
pshapiro4broad: Note that this doesn't test the use case, which is, "update all arrays records with regulatory designation X". We'd probably want to only run that as an integration test, since we wouldn't want to run it in dev.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -1192,4 +1193,69 @@ trait ArraysTests { self: BaseIntegrationSpec =>
...
+
+    val metadataFile: File = File
+      .newTemporaryFile()
+      .overwrite(
+        ArraysMetadata(
+          sampleAlias = Some("amended_sample_alias"),
+          chipType = Some("amended_chip_type")
+        ).asJson.pretty(implicitly)
+      )
+      .deleteOnExit()
```
pshapiro4broad: can you use better files here to remove this?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -1192,4 +1193,69 @@ trait ArraysTests { self: BaseIntegrationSpec =>
...
+      _ <- runUpsertArrays(
+        upsertKey2,
+        ArraysMetadata(
+          chipType = Some("existing_chip_type")
+        )
+      )
+      _ <- runIgnore(
+        ClioCommand.amendArraysName,
+        "--metadata-location",
+        metadataFile.path.toAbsolutePath.toString
```
pshapiro4broad: Not sure why `metadatFile.toString` wouldn't work here?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/dataMock.scala
```diff
@@ -39,6 +41,7 @@ case class ModelMockIndex(
     implicitly[FieldMapper[MetadataType]]
   override val queryInputMapper: FieldMapper[ModelMockQueryInput] =
     implicitly[FieldMapper[QueryInputType]]
+  override val keyFieldNames: Seq[String] = Seq("mock_key_long", "mock_key_string")
```
pshapiro4broad: Note that this stuff isn't a mock, it's test data. A mock is a stubbed/faked implementation of an object or API. This, on the other hand, is real object data and is therefore not a mock.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -258,6 +258,7 @@ class ClioClient(webClient: ClioWebClient, ioUtil: IoUtil)(
         case addCommand: AddCommand[_]       => new AddExecutor(addCommand)
         case moveCommand: MoveCommand[_]     => new MoveExecutor(moveCommand)
         case deleteCommand: DeleteCommand[_] => new DeleteExecutor(deleteCommand)
+        case amendCommand: AmendCommand[_]   => new AmendExecutor(amendCommand)
```
pshapiro4broad: I'm not wild about the verb amend here. To me, amend implies editing or changing, but this is an additive only operation.

Perhaps `backfill` would be better? To me it implies affecting a large number of records by adding values to a table. Or maybe just `fill`?

If we did support non-additive operations, other options here are `change`, `alter`, etc.

tlangs: I agree with your, but naming things is hard and I was struggling to come up with the right term. I'm hesitant to use `backfill`, since that already means upserting new documents from data in the existing databases. What about `patch`?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
tlangs: This is better files. But I probably could.

tlangs: Actually, scratch that. The for-comprehension needs to be of type `Future`, not `ManagedResource`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
tlangs: Obsolete because there is not more `queryInput` argument.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/Executor.scala
tlangs: There's already a `readMetadata` method in `IoUtil` that just reads file data from a `URI`. Seems poorly named. I'm thinking that this `readMetadata` is pretty specific to the client code and therefore should stay in the `clio-client` codebase.

danxmoran: `IoUtil` is in the `clio-client` codebase.

Agreed the existing `readMetadata` is poorly named. It should just be `readFile` or `read`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+      }
+  }
+
+  /**
+    * Munge the patch metadata with the document metadata to get a new metadata json
+    * @param patchMetadataJson The new metadata to insert into the index
+    * @param documentJson The old document to insert the metadata into
+    * @return A tuple containing the munged metadata, the old metadata, and the key for this document
+    */
+  private def mungeMetadata(
```
pshapiro4broad: Is this different than `mergeMetadata`? `Munge` isn't very descriptive.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+    (
+      patchMetadataJson.deepMerge(oldMetadataJson),
+      oldMetadataJson,
+      indexKey
+    )
+  }
+
+  /**
+    * Upsert the new metadata into the document.
+    * This method strips any existing keys out of the metadata to be upserted in order to reduce traffic.
```
pshapiro4broad: The term `keys` here is somewhat confusing is that it's not a clio index key, but a metadata field name, which happens to be a `key` in the map that represents the metadata.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+    * @param oldMetadataJson The old, preexisting metadata
+    * @param docKey The key for this metadata
+    * @param webClient A ClioWebClient to upsert with.
+    * @return
+    */
+  private def upsertPatch(
+    mergedMetadataJson: Json,
+    oldMetadataJson: Json,
+    docKey: DocumentKey,
+    webClient: ClioWebClient
```
pshapiro4broad, line 92: It seems odd to me that `ClioWebClient` and `IoUtil` aren't members of `Executor`, e.g. provided at construction time, since they don't vary over the lifetime of the executor, and are needed for the executor to work.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+import org.broadinstitute.clio.JsonUtils.JsonOps
+
+class PatchExecutor[CI <: ClioIndex](patchCommand: PatchCommand[CI]) extends Executor {
+
+  import patchCommand.index.implicits._
+  import Executor.SourceMonadOps
+
+  private type DocumentKey = patchCommand.index.KeyType
+
+  private val queryAllFile: File =
```
pshapiro4broad: can you make this lazy to avoid writing unless the user is requesting a patch?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+    File.newTemporaryFile().deleteOnExit().write(Json.obj().pretty(implicitly))
```
danxmoran: Ergh...

Alt proposal: change the `rawQuery` method in the web client to take a `Json` instead of a `File`, and push the file-reading into the `RetrieveAndPrintExecutor`. There are going to be other use-cases when we want to build a `Json` in-memory and pass it directly to Clio without jumping through a tempfile.

pshapiro4broad (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+      .mapObject(_.filterKeys(key => !oldMetadataKeys.contains(key)))
+      .as[patchCommand.index.MetadataType]
+      .valueOr(
+        ex =>
+          throw new IllegalArgumentException(
+            "Could not encode new metadata json into class",
+            ex
+        )
+      )
+    webClient.upsert(patchCommand.index)(docKey, patchMetadata, force = false)
```
danxmoran, line 106: Would also be good to add an `upsertJson` API to avoid doing the pointless decode-then-encode. We could make it `private[client]` to lock it down.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -0,0 +1,146 @@
...
+      json.mapObject(
+        _.filterKeys(Executor.getJsonKeyFieldNames(patchCommand.index).contains)
+      )
+
+    keyJson
+      .as[DocumentKey]
+      .valueOr(
+        ex =>
+          throw new IllegalArgumentException(
+            s"Could not create ${patchCommand.index.keyTag} from ${json.pretty(implicitly)}",
```
danxmoran: Would prefer `defaultPrinter` to `implicitly`

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/AddExecutorSpec.scala
```diff
@@ -90,21 +80,23 @@ class AddExecutorSpec extends BaseClientSpec with AsyncMockFactory {
...
-      .expects(loc)
-      .returning(otherMetadata.asJson.pretty(defaultPrinter))
+    (ioUtil
+      .readMetadata(_: ArraysIndex.type)(_: URI))
+      .expects(*, loc)
+      .returning(Source.single(otherMetadata))
 
     val executor = new AddExecutor(AddWgsUbam(theKey, loc, false))
-    recoverToSucceededIf[IllegalArgumentException] {
+    recoverToSucceededIf[Exception] {
```
danxmoran: Why? The more-specific exception is useful for making sure we fail in the expected way.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/PatchExecutorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+    setupQueryReturn(webClient, zipped)
+
+    val executor = new PatchExecutor(PatchArrays(loc))
+    executor.execute(webClient, ioUtil).runWith(Sink.seq).map { _ =>
+      succeed
+    }
+  }
+
+  it should "throw an exception if the wrong type of metadata is given" in {
+    recoverToSucceededIf[ClassCastException] {
```
danxmoran: Er... `ClassCastException`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
tlangs: https://broadinstitute.atlassian.net/browse/DSDEGP-2384

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/PatchExecutorSpec.scala
tlangs: Yes. If you're backfilling arrays and you get a cram, its a `ClassCastException`

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/PatchExecutorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+      val ioUtil = mock[IoUtil]
+      (ioUtil
+        .readMetadata(_: GvcfIndex.type)(_: URI))
+        .expects(*, loc)
```
danxmoran: This `*` is the problem. `PatchExecutor` will always pass in the index for its command (`ArraysIndex` in this case), so this is an impossible expectation.

It should instead expect an `ArraysIndex` but return a `Source.failed` simulating one of the [possible](https://github.com/broadinstitute/clio/pull/267/files#diff-c2407722b19c3ff72071ca32bed8428fR58) [failures](https://github.com/broadinstitute/clio/pull/267/files#diff-c2407722b19c3ff72071ca32bed8428fR66) in `IoUtil`.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -254,7 +254,7 @@ class ClioWebClient(
       }
   }
 
-  private[webclient] def query(clioIndex: ClioIndex)(
+  def query(clioIndex: ClioIndex)(
```
danxmoran, line 257: Maybe `private[client]`? It'd be good to keep this somewhat locked-down.

danxmoran (approved):

PR #268, opened 2018-05-15
Add hook to allow umask to be set for clio docker.

jacarey:
This allows us to control the permissions of all the files that clio creates.
Use 0000 umask for integration test so that data can be cleaned up properly
Remove tmpfs for elasticsearch (I believe we are clean up properly between tests now)

### Description

----

* Review cycle:

### DIFF
--- a/project/src/main/scala/org/broadinstitute/clio/sbt/Docker.scala
+++ b/project/src/main/scala/org/broadinstitute/clio/sbt/Docker.scala
@@ -53,12 +53,21 @@ object Docker {
       label("CLIO_VERSION", version.value)
       add(artifact, artifactTargetPath)
       expose(8080, 31757)
+
       /*
        * We need to use `entryPointShell` here to allow the environment variable to be substituted.
        * By default, using the shell form of ENTRYPOINT prevents signals from reaching the docker container.
        * Prepending `exec` fixes the problem.
        */
-      entryPointShell("exec", "java", "$JAVA_OPTS", "-jar", artifactTargetPath)
+      entryPointShell(
+        "umask",
+        "$UMASK;",
+        "exec",
+        "java",
+        "$JAVA_OPTS",
+        "-jar",
+        artifactTargetPath
+      )
     }
   }
 
--- a/clio-integration-test/src/it/resources/org/broadinstitute/clio/integrationtest/clio-compose.yml
+++ b/clio-integration-test/src/it/resources/org/broadinstitute/clio/integrationtest/clio-compose.yml
@@ -12,6 +12,7 @@ services:
         -Dclio.server.elasticsearch.http-hosts.0.hostname=${ELASTICSEARCH_HOST}
         -Dlogback.configurationFile=/app/clio-logback.xml
         -Dclio.server.persistence.root-dir=/persistence
+      UMASK: ${UMASK}
     volumes:
       - ${CONF_DIR}/clio-logback.xml:/app/clio-logback.xml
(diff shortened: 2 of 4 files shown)


project/src/main/scala/org/broadinstitute/clio/sbt/Docker.scala
```diff
@@ -53,12 +53,21 @@ object Docker {
...
+
       /*
        * We need to use `entryPointShell` here to allow the environment variable to be substituted.
        * By default, using the shell form of ENTRYPOINT prevents signals from reaching the docker container.
        * Prepending `exec` fixes the problem.
        */
-      entryPointShell("exec", "java", "$JAVA_OPTS", "-jar", artifactTargetPath)
+      entryPointShell(
+        "umask",
+        "$UMASK;",
```
pshapiro4broad, line 64: Not sure how this works, is this building a command line out of the arguments? I would use `&&` instead of `;` to sequence commands in sh.

Since there are only two states here it might be simpler to use a boolean and have this code hardcode the two different umasks we want to support.

jacarey: We don't really care if umask fails.. if it does it goes back to default behavior.
Likewise if it isn't set it will just print out and use the default umask value. 
The two states are just set it to something or use the default.

pshapiro4broad: OK I get it. Still, it does seem odd to include two different command lines when calling `entryPointShell()`. But if it works, it works. Have you started an integration build with this branch yet?

jacarey: yeah umask has to be set in the shell that is running the subsequent commands... alternately we could create a shell script that has the umask in it first and the following commands.. but i'm not sure that buys us much

pshapiro4broad: I get that, I meant that usually when writing code that creates a command line, you can't just add a `;` and have all the commands run. E.g. it doesn't work that way in Java `ProcessBuilder`. But I guess it's different in docker.

jacarey: Ah yes. Docker entrypoints have two forms .. shell and exec. In this case we are using shell which adds the instruction

[info] Step 7/7 : ENTRYPOINT umask $UMASK; exec java $JAVA_OPTS -jar /app/clio-server.jar

shell entry points are run with /bin/sh -c

danxmoran:
Test-run in progress here: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-integration/426/

jacarey:
Cleaned up docker on the slave node. Reran integration test here:
https://gotc-jenkins.dsp-techops.broadinstitute.org/job/clio-integration/429

Only failure was the version skew.. I think this is ready for prime time.

danxmoran (approved):

project/src/main/scala/org/broadinstitute/clio/sbt/Docker.scala
pshapiro4broad: Given that, I'm not sure why we bother splitting it up into separate arguments.

jacarey: if we don't then it tries to escape the space and umask fails.. <shrug>

pshapiro4broad (approved):

PR #272, opened 2018-05-20
DSDEGP-2090 Deliver idats to a subdirectory of the target path

danxmoran:
### Description

Needed to match the expectations from the picard-private tool which generates the samples.tsv file.

Raises the question (for later): if the clio-client knows that idats should go in a sub-directory, should it also know that all delivered files go in a directory tree for the project / sample / barcode / version? It'd save us an extra query in the wdl.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
@@ -27,14 +27,15 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: Executio
   ): Source[(ArraysMetadata, immutable.Seq[IoOp]), NotUsed] = {
     (deliveredMetadata.grnIdatPath, deliveredMetadata.redIdatPath) match {
       case (Some(grn), Some(red)) => {
+        val idatDestination = deliverCommand.destination.resolve("/idat/")
         val movedGrnIdat = Metadata.findNewPathForMove(
           grn,
-          deliverCommand.destination,
+          idatDestination,
           ArraysExtensions.IdatExtension
         )
         val movedRedIdat = Metadata.findNewPathForMove(
           red,
-          deliverCommand.destination,
+          idatDestination,
           ArraysExtensions.IdatExtension
         )
         val idatCopies = immutable
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
@@ -16,7 +16,7 @@ import scala.concurrent.ExecutionContext
   * Wraps the move CLP with extra IO / upsert logic:
   *
   *   1. Copies the idat files to the target path
-  *   2. Records the updated locations of the idat files, along with the workspace name, in the metadata
+  *   2. Records the updated locations of the idat files in the metadata
   */
 class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: ExecutionContext)
     extends DeliverExecutor(deliverCommand) {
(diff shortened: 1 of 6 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverExecutor.scala
```diff
@@ -33,21 +33,20 @@ abstract class DeliverExecutor[CI <: DeliverableIndex](
...
-      if (deliverCommand.force || workspaceName.isEmpty || workspaceName.contains(
-            deliverCommand.workspaceName
-          )) {
-        Source.single(metadata)
-      } else {
-        Source.failed(
-          new UnsupportedOperationException(
-            s"Cannot deliver ${deliverCommand.index.name} to workspace '${deliverCommand.workspaceName}' " +
-              s"because it has already been delivered to workspace '${workspaceName.get}'. Use --force " +
-              "if you want to override this restriction."
```
danxmoran, line 47: Compulsively reworking to avoid the weird `if` formatting / the `.get`

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/Metadata.scala
```diff
@@ -106,13 +106,12 @@ object Metadata {
     * extension. Rather than make dangerous guesses, we require that the code calling
     * the move operation provide the file extension to retain during the move operation.
     */
-  def findNewPathForMove(
+  def buildFilePath(
```
danxmoran, line 109: Renamed because "find" made it sound like it was searching for an existing path instead of creating a new one, and because it's not just used for file moves.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/Metadata.scala
```diff
@@ -106,13 +106,12 @@ object Metadata {
...
     source: URI,
     destination: URI,
     extension: String,
     newBasename: Option[String] = None
   ): URI = {
-    // TODO: Rewrite this using a Path-based API.
```
danxmoran, line 115: `Paths.get` of a cloud `URI` will explode without the right global NIO config options. Since we're using the SDK instead of NIO, I don't think we'll be reworking this to use `Path`.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
```diff
@@ -74,3 +79,7 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: Executio
     }
   }
 }
+
+object DeliverArraysExecutor {
+  val IdatsDir = "idats/"
+}
```
tbl3rd, line 85: OOK.

tbl3rd (approved):
Like the new names.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -413,8 +413,8 @@ trait ArraysTests { self: BaseIntegrationSpec =>
     val rootSource = rootTestStorageDir / s"arrays/$chipwellBarcode/v$version/"
     val vcfSource = rootSource / vcfName
     val vcfIndexSource = rootSource / vcfIndexName
-    val grnIdatSource = rootSource / grnIdatName
-    val redIdatSource = rootSource / redIdatName
+    val grnIdatSource = rootSource / "idats" / grnIdatName
```
pshapiro4broad, line 416: Why not use `DeliverArraysExecutor. IdatsDir` here?

pshapiro4broad (approved):
This all seems fine to me but

- Where is the spec for what delivery looks like? We didn't have this going in, so it should be a result of the implementation work.

- Why are we generating a tsv file someplace else that mirrors the behavior of code here? We don't want duplication of logic. Either the tsv or something like it should be an input, or this should produce a manifest that can be used to generate the tsv. Making clio-client delivery produce a manifest in general seems like a good idea.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
danxmoran: That constant ends with a `/`, wasn't sure if it'd mess things up when paired with the method `/`.

PR #274, opened 2018-05-21
DSDEGP-2395 Allow writing of reference paths for arrays

als364:
### Description

Include reference dict, fasta, fasta.fai paths in Clio arrays metadata.

----

* Review cycle:


danxmoran (approved):
Not sure we need the new integration tests, but they don't see how they hurt anything either.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysMetadata.scala
```diff
@@ -28,6 +28,9 @@ case class ArraysMetadata(
   pipelineVersion: Option[Symbol] = None,
   project: Option[String] = None,
   redIdatPath: Option[URI] = None,
+  refFastaPath: Option[URI] = None,
+  refFastaIndexPath: Option[URI] = None,
+  refDictPath: Option[URI] = None,
```
jacarey, line 33: do we care about validating that the file prefixes are the same?

jacarey (approved):

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysMetadata.scala
als364: In theory, what files we use as references could change. What is the relationship of the dict file to the fasta file? Do we do this file prefix validation on other files that have indexes?

jacarey: For instance - 

HTSJDK ReferenceFileFactory expects the prefix for the files to all be the same in order to "find" the dict and index.

https://github.com/samtools/htsjdk/blob/854d7365539c63e39a4ada03ec7abbb408b12993/src/main/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java#L147

Same as index file - 

https://github.com/samtools/htsjdk/blob/854d7365539c63e39a4ada03ec7abbb408b12993/src/main/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java#L171

I don't think we necessarily need to do that validation here.. but some tools will rely on them having the same file prefix.

jacarey: I think typically we have only supplied the fasta file and deduced the other file names from that. I'm not saying that is ideal either but is less likely to trip up htsjdk.

als364: I could either do that validation here or change it (here, in picard-private, and in dsde-pipelines) to accept a file prefix and append the extensions.

jacarey: Personally I prefer having them all defined instead of deriving them. There are likely tools that don't care about prefixes. Even the validation feels like it should only be a warning to me, but I wouldn't be surprised if this validation happens upstream of calls to clio anyway. I think it is fine as is.

PR #275, opened 2018-06-04
Various improvements to client code.

danxmoran:
Some have been nagging at me for awhile, some are lingering low-hanging tickets, and some are from seeing errors when trying to patch reference paths into arrays records in prod.

* Push file IO logic out of web-client (leave it to the executors).
* DSDEGP-2384 Add method for upserting Json.
* DSDEGP-2407 Add logging for cloud IO / successful execution.
* Kill reflection-based ClassUtil in favor of Json-based Show.
* Make upsert parallelism of patch commands configurable.
* Make initial query smarter in patch, to avoid pointless work.
* Add scaladoc to ClioWebClient.

### Description

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -69,6 +69,7 @@ sealed abstract class DeliverCommand[+CI <: DeliverableIndex](override val index
 
 sealed abstract class PatchCommand[CI <: ClioIndex](val index: CI) extends ClioCommand {
   def metadataLocation: URI
+  def parallelism: Int
```
pshapiro4broad: To me this name is unclear. The API called below that accepts this argument calls this attribute `breadth` and describes it as "at most `breadth` substreams are being consumed at any given time".

Maybe `maxParallelStreams`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -23,7 +23,9 @@ class DeleteExecutor[CI <: ClioIndex](deleteCommand: DeleteCommand[CI])(
   implicit ec: ExecutionContext
 ) extends Executor {
 
-  private val prettyKey = ClassUtil.formatFields(deleteCommand.key)
+  import deleteCommand.index.implicits._
+
+  private val prettyKey = deleteCommand.key.show
```
pshapiro4broad, line 28: Does `show` do the same thing as `formatFields()` did? If we had a type `sealed trait IndexCommand(val index:CI) extends ClioCommand` we could add `key` and `prettyKey` to that to avoid having to import `show` here.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
```diff
@@ -4,12 +4,12 @@ import java.net.URI
 
 import akka.NotUsed
 import akka.stream.scaladsl.Source
+import cats.syntax.show._
```
pshapiro4broad, line 7: Is the wildcard import needed if we're only calling `show`?

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
```diff
@@ -149,5 +142,29 @@ class PatchExecutor[CI <: ClioIndex](patchCommand: PatchCommand[CI]) extends Exe
...
+
+object PatchExecutor {
 
+  /**
+    * Given a JSON object containing metadata fields which should be "patched" into
+    * Clio documents, build a JSON object containing an Elasticsearch query which
+    * will return only documents with at least one unset value amongst the fields-to-patch.
+    *
+    * Used to avoid pulling documents from Clio that are guaranteed to not be re-
+    * upserted, since patching doesn't overwrite existing values.
```
pshapiro4broad, line 142: Do we handle the case were the before/after data is the same in `mergeMetadata` above?

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
```diff
@@ -107,6 +108,7 @@ class IoUtil(storage: Storage) {
   }
 
   def writeGoogleObjectData(data: String, location: URI): Unit = {
+    logger.info(s"Writing to '$location'...")
```
pshapiro4broad, line 111: To me this seems more like DEBUG than INFO. Do we have a policy on what is INFO vs DEBUG?

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -135,25 +178,46 @@ class ClioWebClient(
...
     dispatchRequest(HttpRequest(uri = s"/$versionString"), includeAuth = false)
       .via(parserFlow)
 
+  /** Get the health reported by the Clio server. */
   def getClioServerHealth: Source[Json, NotUsed] =
     dispatchRequest(HttpRequest(uri = s"/$healthString"), includeAuth = false)
       .via(parserFlow)
 
+  /**
+    * POST new metadata to the Clio server for a document key.
```
pshapiro4broad: Do you think it's important for the caller to know the underlying http request type for each API? To me that seems like an implementation detail that the caller shouldn't need to know about.

pshapiro4broad (approved):
Nice to use json literals, didn't know there was a way to do that.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: It doesn't show exactly the same thing, but it shows the same information. The way I've set it up, it'll show snake-case field names instead of camel-case (which is how the outside world sees the fields, so I think it's an improvement).

clio-client/src/main/scala/org/broadinstitute/clio/client/util/IoUtil.scala
danxmoran: We don't have a good policy on that. I was trying to mimic the output that `gsutil` used to spit out.

gbggrant (approved):
👍

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: There's probably a specific implicit hook we could pull by name here, but this is the recommended style when pulling in implicit syntax.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeleteExecutor.scala
danxmoran: Tried out pulling the key into the command name, and found a concrete benefit of the `show` approach. For commands like `patch` that don't take a single key as input, we can't encode a pretty key within the command object, but we'd still like to be able to use a consistent key->string definition once we've gotten our hands on keys from the server. Backfill is done with now, but the same situation would have applied there.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/PatchExecutor.scala
danxmoran: It was handled in the `for` comprehension, but I've shuffled things around to make it happen in `mergeMetadata`

PR #277, opened 2018-07-20
Implement move-all.

gbggrant:
Modifed arrays move and cram move.
Updated tests and integration tests.

### Description

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
```diff
@@ -78,6 +81,84 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: Executio
...
+    baseStream.flatMapConcat {
+      case (movedMetadata, moveOps) =>
+        buildDelivery(
+          movedMetadata.withWorkspaceName(deliverCommand.workspaceName),
+          moveOps
+        )
+    }
+  }
+
+  private[dispatch] def buildDeliveryMove(
```
gbggrant: This is a wholesale copy from the base class - probably could be done a lot better, but couldn't quickly find a way to only get the pathsForDelivery without filtering them out.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsCramTests.scala
```diff
@@ -479,21 +519,50 @@ trait WgsCramTests { self: BaseIntegrationSpec =>
...
+        fingerprintDetailMetricsDestination,
+        preAdapterSummaryMetricsDestination,
+        preAdapterDetailMetricsDestination,
+        preBqsrSelfSmDestination
+      ).foreach(_ should exist)
 
       // We don't deliver fingerprinting metrics for now because they're based on unpublished research.
-      fingerprintMetricsSource should exist
-      fingerprintMetricsDestination shouldNot exist
+//      fingerprintMetricsSource should exist
```
gbggrant: Need to modify the Cram delivery executor to not deliver the fingerprints (which I don't think it does anyway?).
In any case, I think the comment is inappropriate here as we *can* move fingerprints

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
danxmoran: Rough idea: add a `deliverInto` method to the [`DeliverableMetadata` trait](https://github.com/broadinstitute/clio/blob/develop/clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/DeliverableMetadata.scala), and call it here.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
```diff
@@ -1,41 +1,40 @@
...
-
-  val DictExtension = ".dict"
-
   val EgtExtension = ".egt"
 
-  val FastaExtension = ".fasta"
+  val IdatExtension = ".idat"
 
-  val FaiExtension = ".fai"
+  val GrnIdatExtension = s"_Grn$IdatExtension"
```
pshapiro4broad, line 16: It looks like this is only used in tests. If so, it should be defined in a test class.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutorSpec.scala
```diff
@@ -30,9 +30,9 @@ class DeliverArraysExecutorSpec extends BaseClientSpec with AsyncMockFactory {
       Some(URI.create(s"gs://bucket/the-vcf${ArraysExtensions.VcfGzTbiExtension}")),
     gtcPath = Some(URI.create(s"gs://bucket/the-gct${ArraysExtensions.GtcExtension}")),
     grnIdatPath =
-      Some(URI.create(s"gs://bucket/the-grn${ArraysExtensions.IdatExtension}")),
+      Some(URI.create(s"gs://bucket/the-grn${ArraysExtensions.GrnIdatExtension}")),
```
pshapiro4broad: Not sure what the point of this change is, was this required to get the tests to pass?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/CramTests.scala
```diff
@@ -482,21 +522,50 @@ trait CramTests { self: BaseIntegrationSpec =>
...
+        fingerprintDetailMetricsDestination,
+        preAdapterSummaryMetricsDestination,
+        preAdapterDetailMetricsDestination,
+        preBqsrSelfSmDestination
+      ).foreach(_ should exist)
 
       // We don't deliver fingerprinting metrics for now because they're based on unpublished research.
-      fingerprintMetricsSource should exist
-      fingerprintMetricsDestination shouldNot exist
+//      fingerprintSummaryMetricsSource should exist
```
pshapiro4broad: What is the status of these tests? If they are no longer valid, the should be removed.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutorSpec.scala
gbggrant: It's the actual extension of green idat files, so I thought it more useful that the code handle those.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
```diff
@@ -78,6 +81,84 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: Executio
         )
     }
   }
+
+  override protected[dispatch] def buildMove(
```
danxmoran: This `override` should get dropped, and the existing `override` in the `DeliverExecutor` should be changed to not make its `super` call.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
danxmoran: Walking myself back after time away, a few things:
1. This is going to do more than just move files, so I wouldn't have `Move` in its name
2. The copied logic is set up to avoid too much boilerplate when the concrete type of `moveCommand.index.MetadataType` isn't known. For the delivery executors, we know exactly what the type of the metadata is, so there's nothing stopping us from calling `copy` on the case class / manually building the list of `IoOps`. I think it'd make the code a lot clearer to do the simple thing here (and in the cram delivery executor, when you get to it).

danxmoran: For example, see how the existing `buildDelivery` explicitly accesses the idat paths. Now that I'm looking at it I think all of this logic should go into that method.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsCramTests.scala
danxmoran: Cram delivery only moves the cram + crai, and generates the md5. It should stay that way (at least in this PR).

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/CramMetadata.scala
```diff
@@ -87,7 +86,39 @@ case class CramMetadata(
...
+      fingerprintingSummaryMetricsPath = pathMapper(
+        fingerprintingSummaryMetricsPath,
+        CramExtensions.FingerprintingSummaryMetricsExtension
+      ),
+      fingerprintingDetailMetricsPath = pathMapper(
+        fingerprintingDetailMetricsPath,
+        CramExtensions.FingerprintingDetailMetricsExtension
+      ),
+      fingerprintVcfPath =
+        pathMapper(fingerprintVcfPath, CramExtensions.FingerprintVcfExtension)
```
danxmoran: Reminder: there are more metrics files to fill in here before merging. I think everything `URI` in this class should be getting moved as part of move execution.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutorSpec.scala
danxmoran: You'll need to change the logic in the `DeliverArraysExecutor` which currently uses the generic idat extension for this to have an effect on main code.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/CramMetadata.scala
danxmoran: We could probably come up with some type-fu to extract out stuff-to-move based on field types... not sure if it's worth it.

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/WgsCramTests.scala
gbggrant: check

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/CramTests.scala
gbggrant: removed now.  Superseded by the testing above

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/arrays/ArraysExtensions.scala
gbggrant: It's used in DeliverArraysExecutor now.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutorSpec.scala
gbggrant: check.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/CramMetadata.scala
gbggrant: There are some that are redundant (fingerprintPath vs fingerprintVcfPath) which I guess I can move both of,
and there are some that are probably legacy (preBqsrDepthSmPath) that I can't find in recent workflows and don't know the extension for.

danxmoran:
Confirming question from Slack: ripping out all of the expectations [like this](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveExecutorSpec.scala#L110) from the `MoveExecutorSpec` should fix most of the current failures. [This test](https://github.com/broadinstitute/clio/blob/develop/clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/MoveExecutorSpec.scala#L267) will probably need to be deleted entirely. Not sure what's up with the failure in the `RetrieveAndPrintExecutor`... will look for anything fishy in the code diff.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
```diff
@@ -22,24 +22,57 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: Executio
     extends DeliverExecutor(deliverCommand) {
 
   override protected[dispatch] def buildDelivery(
-    deliveredMetadata: ArraysMetadata,
-    moveOps: immutable.Seq[IoOp]
+    deliveredMetadata: ArraysMetadata
```
danxmoran: I'd call this `existingMetadata` to be clear the method is transforming pre-delivery state into post-delivery state.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
```diff
@@ -22,24 +22,57 @@ class DeliverArraysExecutor(deliverCommand: DeliverArrays)(implicit ec: Executio
...
+          moveCommand.newBasename
+        )
+        val moves = immutable
+          .Seq(
+            MoveOp(vcf, vcfMove),
+            MoveOp(index, vcfIndexMove),
+            MoveOp(gtc, gtcMove)
+          )
+          .filterNot { op =>
+            op.src.equals(op.dest)
```
danxmoran: This filter would be worth defining once in the `MoveExecutor`, in a separate follow-up method. We'd need to make it handle all 3 cases (but that should be OK). Both moves and copies get dropped if src == dest, and writes always happen(?).

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverExecutor.scala
```diff
@@ -51,19 +50,14 @@ abstract class DeliverExecutor[CI <: DeliverableIndex](
...
-    ioUtil: IoUtil
+    metadata: moveCommand.index.MetadataType
   ): Source[(moveCommand.index.MetadataType, immutable.Seq[IoOp]), NotUsed] = {
 
-    val baseStream = super
-      .buildMove(metadata, ioUtil)
-      .orElse(Source.single(metadata -> immutable.Seq.empty))
-
+    val baseStream = Source.single(metadata -> immutable.Seq.empty)
     baseStream.flatMapConcat {
```
danxmoran: Shouldn't need the `baseStream` / `flatMapConcat` wrapper any more; could just call `buildDelivery(metadata.withWorkspaceName(...))`

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/CramTests.scala
```diff
@@ -412,50 +424,88 @@ trait CramTests { self: BaseIntegrationSpec =>
...
+    val crosscheckDestination = rootDestination / s"$endBasename${CramExtensions.CrossCheckExtension}"
+    val duplicateMetricsDestination = rootDestination / s"$endBasename${CramExtensions.DuplicateMetricsExtension}"
+    val fingerprintVcfDestination = rootDestination / s"$endBasename${CramExtensions.FingerprintVcfExtension}"
+    val fingerprintSummaryMetricsDestination = rootDestination / s"$endBasename${CramExtensions.FingerprintingSummaryMetricsExtension}"
+    val fingerprintDetailMetricsDestination = rootDestination / s"$endBasename${CramExtensions.FingerprintingDetailMetricsExtension}"
+    val preAdapterSummaryMetricsDestination = rootDestination / s"$endBasename${CramExtensions.PreAdapterSummaryMetricsExtension}"
+    val preAdapterDetailMetricsDestination = rootDestination / s"$endBasename${CramExtensions.PreAdapterDetailMetricsExtension}"
+    val preBqsrSelfSmDestination = rootDestination / s"$endBasename${CramExtensions.PreBqsrSelfSMExtension}"
+    val preBqsrDepthSmDestination = rootDestination / s"$endBasename${CramExtensions.PreBqsrDepthSMExtension}"
```
danxmoran: I feel bad asking to add more, but I'd also want to check:
1. All readgroup-level metrics (stored as an array in Cram metadata) get moved
2. All logs (pointed to by a top-level directory) get moved

Should be possible to test with 4 extra files.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/CramMetadata.scala
```diff
@@ -87,7 +86,98 @@ case class CramMetadata(
...
+      baitBiasSummaryMetricsPath = pathMapper(
+        baitBiasSummaryMetricsPath,
+        CramExtensions.BaitBiasSummaryMetricsExtension
+      ),
+      baitBiasDetailMetricsPath = pathMapper(
+        baitBiasDetailMetricsPath,
+        CramExtensions.BaitBiasDetailMetricsExension
+      ),
+      wgsMetricsPath = pathMapper(wgsMetricsPath, CramExtensions.WgsMetricsExension)
+//      readgroupLevelMetricsFiles = pathMapper(readgroupLevelMetricsFiles)
```
danxmoran: Test I asked for will make sure this gets uncommented.

I think you want something like:
```scala
readgroupLevelMetricsFiles = readgroupLevelMetricsFiles.map(pathMapper(_, ???))
```
Though as I typed it I realized there's no consistent extension to use...

danxmoran (commented):
Nearly there!

I'm not sure what to do about the readgroup-level metrics files. We could punt on it for now and make a separate ticket specifically for dealing with them.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/wgscram/CramMetadata.scala
danxmoran: Something like this might work:
```scala
readgroupLevelMetricsFiles = readgroupLevelMetricsFiles.map { path =>
  val name = new File(path).getName
  pathMapper(path, name)
}
```
It's a bad hack, but I think it'll fool the existing code. Better would be for us to restructure this logic to distinguish things-which-can-change-their-base-name from things-which-can-only-be-moved, but that's for another ticket.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
gbggrant: check

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverExecutor.scala
gbggrant: Thanks.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutor.scala
gbggrant: check.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
```diff
@@ -156,47 +156,21 @@ class MoveExecutor[CI <: ClioIndex](protected val moveCommand: MoveCommand[CI])(
...
-                   |${nonCloudPaths.mkString(",")}""".stripMargin
-              )
-            )
-          }
-        }
+  protected[dispatch] def filterOutUnneededTransfers(
+    ioOps: immutable.Seq[IoOp]
+  ): immutable.Seq[IoOp] = {
+    ioOps.filterNot {
+      case moveOp: MoveOp => moveOp.src.equals(moveOp.dest)
```
danxmoran: Slightly less code: `case MoveOp(src, dest) => src.equals(dest)`, same for the copy case.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/MetadataMover.scala
```diff
@@ -0,0 +1,61 @@
...
+object MetadataMover {
+
+  /**
+    * Given a `source` path, a `destination` directory, and an extension, figure out what
+    * the new file name should be were the source file to be moved into the destination path,
+    * optionally changing the base-name in the process.
+    *
+    * NOTE: Because we use '.' liberally in both our file names and file extensions,
+    * there's no easy formula at this abstract level for determining which part of a
+    * filename should be replaced by "newBasename" and which should be retained as the
```
pshapiro4broad, line 51: Does this comment still belong here? There is no `newBasename` and there's no code that handles file extensions here.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/MetadataMover.scala
```diff
@@ -0,0 +1,61 @@
...
+        s"Non-directory destination '$destination' given for metadata move"
+      )
+    }
+
+    moveMetadata(src, destination, newBasename) match {
+      case (metadata, ops) => (metadata, ops.to[immutable.Seq])
+    }
+  }
+
+  protected def moveMetadata(
```
pshapiro4broad, line 28: scaladoc please

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutorSpec.scala
```diff
@@ -1,114 +0,0 @@
...
-
-        newMetadata.workspaceName should be(Some(workspaceName))
-        newMetadata.grnIdatPath should be(movedGrn)
-        newMetadata.redIdatPath should be(movedRed)
-
-        ops should contain allElementsOf Seq.concat(grnMove, redMove)
-    }
-  }
-
-  it should "fail if no red idat is registered to a document" in {
```
pshapiro4broad, line 87: it looks like the negative tests weren't ported, was there a reason for that?

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramMover.scala
```diff
@@ -0,0 +1,288 @@
...
+    )
+
+    val dest = src.copy(
+      cramPath = movedCram,
+      craiPath = movedCram.map { cramUri =>
+        // DSDEGP-1715: We've settled on '.cram.crai' as the extension and
+        // want to fixup files with just '.crai' when possible.
+        URI.create(s"$cramUri${CramExtensions.CraiExtensionAddition}")
+      },
+      logPath = src.logPath.map(buildFilePath(_, destination)),
```
pshapiro4broad, line 26: only `logPath` doesn't require handling `newBasename`?

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramMover.scala
```diff
@@ -0,0 +1,288 @@
...
+          _,
+          destination,
+          newBasename.map(_ + CramExtensions.DuplicateMetricsExtension)
+        )
+      ),
+      fingerprintPath = src.fingerprintPath.map(
+        buildFilePath(
+          _,
+          destination,
+          newBasename.map(_ + CramExtensions.FingerprintVcfExtension)
```
pshapiro4broad, line 108: it looks like this will overwrite `fingerprintVcfPath` since it has the same file extension

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramMover.scala
```diff
@@ -0,0 +1,288 @@
...
+      ),
+      wgsMetricsPath = src.wgsMetricsPath.map(
+        buildFilePath(
+          _,
+          destination,
+          newBasename.map(_ + CramExtensions.WgsMetricsExension)
+        )
+      ),
+      readgroupLevelMetricsFiles =
+        src.readgroupLevelMetricsFiles.map(_.map(buildFilePath(_, destination)))
```
pshapiro4broad, line 245: no renaming required here?

pshapiro4broad (approved):
Looks ok, too bad about all the code expansion

danxmoran (approved):
Needs a scalafmt, but LGTM after

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramMover.scala
danxmoran: The fields refer to the same file. Looks like we don't actually use `fingerprintVcfPath` in prod based on querying:
```
{
    "query": {
        "exists": { "field": "fingerprint_vcf_path" }
    }
}
```
And getting no results. We could take the chance to remove the field now.

clio-client/src/test/scala/org/broadinstitute/clio/client/dispatch/DeliverArraysExecutorSpec.scala
danxmoran: Checking these cases got dropped from the main code

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramMover.scala
danxmoran: `logPath` and `readgroupLevelMetrics`. But `logPath` doesn't actually work as-is, so this should get removed for now.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramMover.scala
danxmoran: Nope. If we did rename we'd get naming conflicts all over the place (unless we did something like `newBase + full old name`)

PR #284, opened 2018-08-14
DSDEGP-2469: move idats when delivering from a workspace

tlangs:
### Description

If the workspace name is defined and not empty, then we assume the Array has been delivered and we can move the idats instead of copying them.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/ArrayDeliverer.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/ArrayDeliverer.scala
@@ -1,7 +1,7 @@
 package org.broadinstitute.clio.client.metadata
 import java.net.URI
 
-import org.broadinstitute.clio.client.dispatch.MoveExecutor.{CopyOp, IoOp}
+import org.broadinstitute.clio.client.dispatch.MoveExecutor.{CopyOp, IoOp, MoveOp}
 import org.broadinstitute.clio.transfer.model.arrays.{ArraysExtensions, ArraysMetadata}
 
 class ArrayDeliverer extends MetadataMover[ArraysMetadata] {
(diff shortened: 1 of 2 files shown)


tbl3rd (approved):
Add a workspaceName case to ArrayDelivererSpec?

Idats are not shared across arrays in FireCloud?

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/ArrayDeliverer.scala
```diff
@@ -40,12 +40,18 @@ class ArrayDeliverer extends MetadataMover[ArraysMetadata] {
       redIdatPath = src.redIdatPath.map(buildFilePath(_, idatDestination))
     )
 
+    // If the Array has already been delivered, we want to move the idats instead of copying them.
+    val idatOp = if (src.workspaceName.forall(_.isEmpty)) {
```
pshapiro4broad, line 44: Seems like we do this test in at least one other place in the code, what if we had a `hasBeenDelivered()` API to avoid code/logic duplication?

pshapiro4broad (commented):
This does what it says but am wondering what the decision behind this is; the ticket is still about un-delivering. Unless this is the un-delivering code?

tlangs:
@tbl3rd The idats are copied into the firecloud buckets when delivered. I don't think there's a case where we would deliver an array to two different places

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/ArrayDeliverer.scala
tbl3rd: Phil explained that there is a idat copy per array delivery in FireCloud.
Idat files are not shared across arrays there.

pshapiro4broad:
OK I now get it. Maybe a better way to think about is isn't un-deliver (or re-deliver), it's that the first time an idat is delivered, you must copy since it's shared. Every time after that you can move because it's no longer shared.

Maybe a slightly longer comment would help here.

tlangs:
@pshapiro4broad exactly.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/ArrayDeliverer.scala
tlangs: Since we're only doing this for arrays delivery and for this specific case, I don't see why we need to abstract the logic out.

pshapiro4broad: It looks like it could be done in `DeliverableMetadata` and then shared with the code in `CramMetadata.pathsToDelete`. There is a similar use in `DeliverExecutor.checkPreconditions` but that is testing to see if it has been delivered _and_ the workspace name is different.

pshapiro4broad (approved):

PR #292, opened 2018-08-27
Added idle-timeout config for client

tlangs:
### Description

The aggregation project backfill keeps timing out as the database catches up with supplying records. Being able to specify a long idle-timeout will make things much easier.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -70,7 +70,6 @@ object ClioWebClient {
     new ClioWebClient(
       connectionFlow,
       responseTimeout,
-      idleTimeout,
       maxRequestRetries,
       GoogleCredentialsGenerator(credentials)
     )
@@ -94,7 +93,6 @@ object ClioWebClient {
 class ClioWebClient(
   connectionFlow: Flow[HttpRequest, HttpResponse, _],
   responseTimeout: FiniteDuration,
-  idleTimeout: FiniteDuration,
   maxRequestRetries: Int,
   tokenGenerator: CredentialsGenerator
 ) extends StrictLogging {
@@ -146,7 +144,7 @@ class ClioWebClient(
       .single(requestWithCreds)
       .via(connectionFlow)
       .initialTimeout(responseTimeout)
-      .idleTimeout(idleTimeout)
+      .idleTimeout(responseTimeout)
 
     /*
      * Retry on any connection failures (since Akka's built-in retry mechanisms
@@ -164,7 +162,7 @@ class ClioWebClient(
     retriedResponse.flatMapConcat { response =>
       if (response.status.isSuccess()) {
         logger.debug(s"Successfully completed request: $request")
-        response.entity.withoutSizeLimit().dataBytes.idleTimeout(idleTimeout)
+        response.entity.withoutSizeLimit().dataBytes.idleTimeout(responseTimeout)
       } else {
         response.entity.dataBytes.reduce(_ ++ _).flatMapConcat { bytes =>
           Source.failed {
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -47,6 +47,7 @@ object ClioWebClient {
     clioPort: Int = ClioClientConfig.ClioServer.clioServerPort,
     useHttps: Boolean = ClioClientConfig.ClioServer.clioServerUseHttps,
     requestTimeout: FiniteDuration = ClioClientConfig.responseTimeout,
+    idleTimeout: FiniteDuration = ClioClientConfig.idleTimeout,
     maxRequestRetries: Int = ClioClientConfig.maxRequestRetries
   )(implicit system: ActorSystem): ClioWebClient = {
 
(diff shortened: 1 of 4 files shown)


pshapiro4broad:
Can this be done by using a custom application.conf file instead of using a command line option?

tlangs:
@pshapiro4broad Unfortunately, the `idleTimeout` of the Akka stream is set in code, so code changes are needed.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -143,6 +146,7 @@ class ClioWebClient(
       .single(requestWithCreds)
       .via(connectionFlow)
       .initialTimeout(requestTimeout)
+      .idleTimeout(idleTimeout)
```
pshapiro4broad: If this is omitted, then there is no idle timeout, correct? If so, why is one needed now when it wasn't before?

That is, the code doesn't seem to be making a default timeout longer, but instead adding one where there wasn't one before, in this case at least.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -160,7 +164,7 @@ class ClioWebClient(
     retriedResponse.flatMapConcat { response =>
       if (response.status.isSuccess()) {
         logger.debug(s"Successfully completed request: $request")
-        response.entity.withoutSizeLimit().dataBytes.idleTimeout(requestTimeout)
+        response.entity.withoutSizeLimit().dataBytes.idleTimeout(idleTimeout)
```
pshapiro4broad: So the code before was using the `requestTimeout` as the `idleTimeout` as well. Why not use the same value for both?

pshapiro4broad (commented):
I see, this _is_ adding a new setting to the conf file.

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
```diff
@@ -28,6 +28,7 @@ class ClioWebClientSpec extends BaseClientSpec with AsyncMockFactory {
 
   //private val index = ModelMockIndex()
```
pshapiro4broad: may as well remove this line while you're here

clio-client/src/test/scala/org/broadinstitute/clio/client/webclient/ClioWebClientSpec.scala
```diff
@@ -28,6 +28,7 @@ class ClioWebClientSpec extends BaseClientSpec with AsyncMockFactory {
...
   private val timeout = 1.second
+  private val idleTimeout = 1.second
```
pshapiro4broad: Can you add a test that verifies that the new idle timeout setting is having an effect? Or is there an existing test that already asserts this?

tlangs: Can't comment on it, but it's already tested on line 131. Instead of using the `requestTimeout` to set the idle timeout like we were before, this uses an explicit `idleTimout` to do so.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -47,6 +47,7 @@ object ClioWebClient {
     clioPort: Int = ClioClientConfig.ClioServer.clioServerPort,
     useHttps: Boolean = ClioClientConfig.ClioServer.clioServerUseHttps,
     requestTimeout: FiniteDuration = ClioClientConfig.responseTimeout,
```
pshapiro4broad: It is confusing that these names are different :(

tlangs: Yes. Maybe @danxmoran has some insight?

danxmoran: ¯\\\_(ツ)\_/¯ responseTimeout is more accurate than requestTimeout, I'd say.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
danxmoran: Same question

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
danxmoran: Same question

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
tlangs: I'll change it then.

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -142,7 +142,8 @@ class ClioWebClient(
     val responseSource = Source
       .single(requestWithCreds)
       .via(connectionFlow)
-      .initialTimeout(requestTimeout)
+      .initialTimeout(responseTimeout)
+      .idleTimeout(responseTimeout)
```
danxmoran, line 146: This still seems like the PR is making things more restrictive, not less. If you crank up the existing response timeout parameter to something huge, does that not fix it?

tlangs: It does not. The problem is the idle-timeout on the `Source`

danxmoran (approved):

pshapiro4broad (approved):

PR #293, opened 2018-08-30
Modify clio to use sbt 1.2, in part by removing some settings.

pshapiro4broad:
Things compile but it's not ready to merge.

This is just an experiment to see how much work would be required to migrate to 1.0.


build.sbt
```diff
@@ -4,7 +4,8 @@ import org.broadinstitute.clio.sbt._
 // ## sbt settings
 // For more info, see: https://broadinstitute.atlassian.net/wiki/pages/viewpage.action?pageId=114531509
 
-enablePlugins(GitVersioning)
+lazy val root = (project in file("."))
+  .enablePlugins(GitVersioning)
```
danxmoran, line 8: There's already a root project defined here: https://github.com/broadinstitute/clio/pull/293/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR51. What's the motivation for scoping the `enablePlugins` on it?

project/build.properties
```diff
@@ -1,3 +1,3 @@
 # Set the sbt version used by the project. scala version set in build.sbt.
 # suppress inspection "UnusedProperty"
-sbt.version=0.13.17
+sbt.version=1.2.1
```
danxmoran, line 3: I think it's up to 1.2.3 by now

project/src/main/scala/org/broadinstitute/clio/integrationtest/ClioIntegrationTestSettings.scala
```diff
@@ -58,8 +58,9 @@ object ClioIntegrationTestSettings extends BuildInfoKeys {
     Def.task {
       val generatedResourceDir = (resourceManaged in IntegrationTest).value
 
-      val DockerMajorVersionRegex(dockerMajorVersion) =
-        Process(Seq("docker", "--version")).!!.trim
+//      val DockerMajorVersionRegex(dockerMajorVersion) =
+//        Process(Seq("docker", "--version")).!!.trim
```
danxmoran, line 62: What's the motivation for this? Did `Process` go away?

project/src/main/scala/org/broadinstitute/clio/publish/ArtifactoryFatJarPublishingPlugin.scala
```diff
@@ -30,9 +30,11 @@ object ArtifactoryFatJarPublishingPlugin extends AutoPlugin {
...
     "[organization]/[module]/[revision]/[type]s/ivy-[revision].xml"
   private val ivyArtifactoryPatterns = Patterns(
-    ivyPatterns = Seq(ivyXmlPattern),
-    artifactPatterns = Seq(ivyArtifactPattern),
-    isMavenCompatible = true
+    ivyPatterns = Vector(ivyXmlPattern),
+    artifactPatterns = Vector(ivyArtifactPattern),
+    isMavenCompatible = true,
+    descriptorOptional = true,
+    skipConsistencyCheck = false
```
danxmoran, line 37: Were there equivalent options for `descriptorOptional` and `skipConsistencyCheck` in 0.13 that these are now matching? Or are these brand new?

project/src/main/scala/org/broadinstitute/clio/publish/ArtifactoryFatJarPublishingPlugin.scala
```diff
@@ -49,13 +51,7 @@ object ArtifactoryFatJarPublishingPlugin extends AutoPlugin {
       Seq(
         publishTo := Some(ivyArtifactoryResolver),
         publishMavenStyle := false,
-        isSnapshot := true,
-        moduleSettings := InlineConfigurationWithExcludes(
-          projectID.value,
-          projectInfo.value,
-          Seq.empty,
-          configurations = Seq(Default)
-        )
```
danxmoran, line 58: I think we need to keep the `moduleSettings` so gradle can find the jar in the zamboni nightly test.

build.sbt
pshapiro4broad: I forgot but I think `enablePlugins()` no longer worked at the top level

project/src/main/scala/org/broadinstitute/clio/publish/ArtifactoryFatJarPublishingPlugin.scala
pshapiro4broad: yeah, I couldn't get it to compile so I removed it for now, but I assumed some stuff will be broken until it's added back in

PR #300, opened 2018-10-24
DSDEGP-2829 deliver metrics only with command line option

tlangs:
### Description

Delivering sample level metrics files should only happen when there is no fee for service. This allows for a command line option to turn off sample level metrics delivery from the delivery WDL.
----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
@@ -64,6 +64,7 @@ sealed abstract class DeliverCommand[+CI <: DeliverableIndex](override val index
   final def destination: URI = workspacePath
   def newBasename: Option[String]
   def force: Boolean
+  def deliverSampleLevelMetrics: Boolean
 }
 
 sealed abstract class PatchCommand[CI <: ClioIndex](val index: CI) extends ClioCommand {
@@ -71,9 +72,6 @@ sealed abstract class PatchCommand[CI <: ClioIndex](val index: CI) extends ClioC
   def maxParallelUpserts: Int
 }
 
-sealed abstract class BackCompatibleDeliverCram extends DeliverCommand(CramIndex) {
-  override val metadataMover = new CramDeliverer
-}
 // Generic commands.
 
 @CommandName(ClioCommand.getServerHealthName)
@@ -166,8 +164,11 @@ final case class DeliverCram(
   workspaceName: String,
   workspacePath: URI,
   newBasename: Option[String] = None,
-  force: Boolean = false
-) extends BackCompatibleDeliverCram
+  force: Boolean = false,
+  deliverSampleLevelMetrics: Boolean = false
+) extends DeliverCommand(CramIndex) {
+  override val metadataMover = new CramDeliverer
+}
 
 @CommandName(ClioCommand.patchCramName)
 final case class PatchCram(
@@ -262,6 +263,7 @@ final case class DeliverArrays(
   force: Boolean = false
 ) extends DeliverCommand(ArraysIndex) {
   override val metadataMover = new ArrayDeliverer
+  override def deliverSampleLevelMetrics: Boolean = false
 }
 
 @CommandName(ClioCommand.patchArraysName)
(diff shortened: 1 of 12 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -64,16 +64,14 @@ sealed abstract class DeliverCommand[+CI <: DeliverableIndex](override val index
   final def destination: URI = workspacePath
   def newBasename: Option[String]
   def force: Boolean
+  def deliverSampleLevelMetrics: Boolean
```
pshapiro4broad, line 67: Is there any way to do this without adding this option to the base deliver command, e,g. only add it to the cram delivery command? It seems a waste to add it to every deliver command if it's only used by one.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
```diff
@@ -262,6 +263,7 @@ final case class DeliverArrays(
   force: Boolean = false
 ) extends DeliverCommand(ArraysIndex) {
   override val metadataMover = new ArrayDeliverer
+  override def deliverSampleLevelMetrics: Boolean = false
```
pshapiro4broad, line 266: sample level metrics are specific to crams so this option is not appropriate for arrays

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
```diff
@@ -142,10 +142,16 @@ class MoveExecutor[CI <: ClioIndex](protected val moveCommand: MoveCommand[CI])(
     * more IO operations.
     */
   protected[dispatch] def buildMove(
-    metadata: moveCommand.index.MetadataType
+    metadata: moveCommand.index.MetadataType,
+    deliverMetrics: Boolean = false
```
pshapiro4broad, line 146: According to DSDEGP-2229 the sample level metrics feature is for deliver only, not move.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
```diff
@@ -37,7 +38,8 @@ class CramDeliverer extends MetadataMover[CramMetadata] {
       writeMd5Op.toIterable
     ).flatten
 
-    if (src.regulatoryDesignation.exists(_.equals(RegulatoryDesignation.ResearchOnly))) {
+    if (deliverMetrics &&
+        src.regulatoryDesignation.exists(_.equals(RegulatoryDesignation.ResearchOnly))) {
```
pshapiro4broad, line 42: Now that this feature is opt-in only, does the code still need to check for `ResearchOnly`? We should find out from ops and/or Janki how they expect this feature to be used.

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/MoveExecutor.scala
tlangs: Because of the way clio-client is structured, the deliver stuff uses the move trait, which means we have to pass this option to all move implementers :(

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
tlangs: I agree. As of now, it acts as an extra safeguard

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
tlangs: It will never appear as a command line option for arrays. Since its defined as a `def` and not in the parameter list, it will always be false and cannot be changed at runtime.

clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
tlangs: I don't think so. It needs to be a part of the abstract `DeliverCommand`, or else we can't pass it though to the actual deliver logic.

als364:
Completed in #301.

PR #301, opened 2018-10-29
DSDEGP-2829: add option to control metrics delivery for deliver cram command

pshapiro4broad:
Here's another approach to fix this issue that only affects the cram delivery command.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
@@ -4,9 +4,9 @@ import java.net.URI
 import better.files.File
 import org.broadinstitute.clio.client.dispatch.MoveExecutor.{IoOp, MoveOp, WriteOp}
 import org.broadinstitute.clio.transfer.model.cram.{CramExtensions, CramMetadata}
-import org.broadinstitute.clio.util.model.RegulatoryDesignation
 
-class CramDeliverer extends MetadataMover[CramMetadata] {
+case class CramDeliverer(deliverSampleMetrics: Boolean)
+    extends MetadataMover[CramMetadata] {
   override protected def moveMetadata(
     src: CramMetadata,
     destination: URI,
@@ -37,7 +37,7 @@ class CramDeliverer extends MetadataMover[CramMetadata] {
       writeMd5Op.toIterable
     ).flatten
 
-    if (src.regulatoryDesignation.exists(_.equals(RegulatoryDesignation.ResearchOnly))) {
+    if (deliverSampleMetrics) {
       lazy val oldBaseName =
         src.cramPath.map(
           p => File(p.getPath).name.replace(CramExtensions.CramExtension, "")
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
@@ -1,4 +1,5 @@
 package org.broadinstitute.clio.client.metadata
+
 import java.net.URI
 
 import better.files.File
@@ -38,6 +39,11 @@ case class CramDeliverer(deliverSampleMetrics: Boolean)
     ).flatten
 
     if (deliverSampleMetrics) {
+      if (src.regulatoryDesignation.exists(_.isClinical)) {
+        throw new RuntimeException(
+          s"Cannot delivery sample level metrics for clinical sample $src"
+        )
+      }
       lazy val oldBaseName =
         src.cramPath.map(
           p => File(p.getPath).name.replace(CramExtensions.CramExtension, "")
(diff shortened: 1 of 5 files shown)


tlangs (approved):
Looks good, but we should probably put in some integration tests to verify functionality from the command line. You can pull those test from my branch.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
```diff
@@ -37,7 +37,7 @@ class CramDeliverer extends MetadataMover[CramMetadata] {
       writeMd5Op.toIterable
     ).flatten
 
-    if (src.regulatoryDesignation.exists(_.equals(RegulatoryDesignation.ResearchOnly))) {
+    if (deliverSampleMetrics) {
```
tlangs, line 41: We should talk about where we want to the check for regulatory designation to happen. Doing it in the delivery WDL will be kinda unwieldy. We don't have the Regulatory Designation available on a per-sample basis when we call `MoveSamples` in the WDL to make the decision of whether or not to actually move sample level metrics.

als364: Why not just keep it here?

clio-client/src/test/scala/org/broadinstitute/clio/client/metadata/CramDelivererSpec.scala
```diff
@@ -51,10 +51,12 @@ class CramDelivererSpec extends FlatSpec with Matchers {
...
 
-  private val deliverer = new CramDeliverer
+  private val delivererWithMetrics = CramDeliverer(deliverSampleMetrics = true)
 
-  it should "generate ops to move the cram & crai, and write the cram md5" in {
-    val (delivered, ops) = deliverer.moveInto(metadata, destination)
+  private val delivererNoMetrics = CramDeliverer(deliverSampleMetrics = false)
+
+  it should "generate ops to move the cram & crai, write the cram md5 and copy metrics files" in {
+    val (delivered, ops) = delivererWithMetrics.moveInto(metadata, destination)
```
als364, line 60: Please add a check to verify that the metrics files are in the right place

clio-client/src/test/scala/org/broadinstitute/clio/client/metadata/CramDelivererSpec.scala
```diff
@@ -109,10 +111,11 @@ class CramDelivererSpec extends FlatSpec with Matchers {
     )
   }
 
-  it should "use the new basename for the md5 file, if given" in {
+  it should "use the new basename for the md5 file, if given, and copy metrics files" in {
     val basename = "the-new-basename"
 
-    val (delivered, ops) = deliverer.moveInto(metadata, destination, Some(basename))
+    val (delivered, ops) =
```
als364, line 118: Please add a check to verify that the metrics files are in the right place

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
pshapiro4broad: The original JIRA says
> All sample-level metrics files should be delivered for most samples but not for clinical samples or for fee-for-service projects

Since we don't have a way to test for fee-for-service (and since we don't know yet how users will request this delivery feature) I thought it best to only use the command line argument to toggle this setting.

clio-client/src/test/scala/org/broadinstitute/clio/client/metadata/CramDelivererSpec.scala
pshapiro4broad: This is already done below, see the lines regarding `preAdapterSummaryMetricsPath`, `preAdapterDetailMetricsPath`, etc.

clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/CramDeliverer.scala
tlangs: We know fee-for-service when delivery is started, so that flag can be set easily in the WDL. We don't know the Regulatory Designation of each sample in the WDL though, and its much more effort to get that into the WDL than to just check for it when we're already in Clio.

pshapiro4broad: OK, I can add it back in. The original issue says `not for clinical samples` but the code you added tests for `ResearchOnly`. Is there a reason why you didn't use `!Clinical`?

I don't see why anyone would need to know regulatory designation, though. We aren't doing to deliver these metrics unless someone runs delivery with a command line option. This is a manual step; at that point all the checks should've been done to confirm that this is the right thing to do. If we get to this point and silently omit the metrics the user has asked for, that doesn't seem correct either. I guess we could add code to fail in that case. But now we're adding extra code and we don't yet know when this feature will be used.

clio-client/src/test/scala/org/broadinstitute/clio/client/metadata/CramDelivererSpec.scala
als364: github autocollapse strikes again.

als364 (approved):

PR #308, opened 2019-08-08
GH-358 modify config to allow url encoded slashes

charlesdyan:
### Description

Apache blocks all requests with a urlencoded slash in the path. This causes sample names with slashes to break upsert.

----

* Review cycle:

### DIFF
--- a/clio-server/config/site.conf
+++ b/clio-server/config/site.conf
@@ -25,7 +25,7 @@ Redirect 307 / https://${SERVER_NAME}/
 ServerAlias ${SERVER_ALIAS}
 DocumentRoot /app
 
-AllowEncodedSlashes On
+AllowEncodedSlashes NoDecode
 <Directory "/app">
 AllowOverride All
 Options -Indexes
--- a/clio-server/config/site.conf
+++ b/clio-server/config/site.conf
@@ -25,7 +25,7 @@ Redirect 307 / https://${SERVER_NAME}/
 ServerAlias ${SERVER_ALIAS}
 DocumentRoot /app
 
-AllowEncodedSlashes NoDecode
+AllowEncodedSlashes On
 <Directory "/app">
 AllowOverride All
 Options -Indexes
--- a/clio-server/config/site.conf
+++ b/clio-server/config/site.conf
@@ -25,8 +25,8 @@ Redirect 307 / https://${SERVER_NAME}/
 ServerAlias ${SERVER_ALIAS}
 DocumentRoot /app
 
+AllowEncodedSlashes NoDecode
 <Directory "/app">
-AllowEncodedSlashes On
 AllowOverride All
 Options -Indexes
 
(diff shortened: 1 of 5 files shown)


clio-server/config/site.conf
```diff
@@ -25,6 +25,7 @@ Redirect 307 / https://${SERVER_NAME}/
 ServerAlias ${SERVER_ALIAS}
 DocumentRoot /app
 
+AllowEncodedSlashes NoDecode
```
pshapiro4broad, line 28: Do you have a test that verifies that this fixes the issue? I would've thought that using `NoDecode` would require our code to explicitly decode the URL encoding.

charlesdyan: I've confirmed it works by deploying this branch to of Clio server to dev 
`wm28d-f87:clio yanc$ java -Dclio.server.hostname=clio.gotc-dev.broadinstitute.org -Dclio.server.port=443 -Dclio.server.use-https=true -Dclio.client.service-account-json=clio-dev-account.json -XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -Xmx256m -jar clio-client.jar add-gvcf --project C321 --data-type Exome --sample-alias "CEPH/UTAH PEDIGREE 1463" --location GCP --version 1 --metadata-location gvcf_metadata.json
2019-08-08T21:07:14,280Z [INFO] Successfully updated record for {"location": "GCP", "project": "C321", "data_type": "Exome", "sample_alias": "CEPH/UTAH PEDIGREE 1463", "version": 1}.
`
whereas we would have got a 404 before deploying. But would probably still be good to see if I can add something to the integration test...

charlesdyan: Ah, actually the success message uses the pre-encoded values. Querying yields the encoded result... but just for the forward slash

charlesdyan: Ugh, but if we use `AllowEncodedSlashes On`, it's decoded before the route is parsed, so we get a different kind of 404... Will have to explicitly decode after the routing logic

charlesdyan:
retest this please

clio-server/src/main/scala/org/broadinstitute/clio/server/webservice/IndexWebService.scala
```diff
@@ -10,6 +10,12 @@ import org.broadinstitute.clio.transfer.model.ApiConstants._
 import org.broadinstitute.clio.transfer.model._
 import org.broadinstitute.clio.util.model.UpsertId
 
+object IndexWebService {
+
+  def decodeSlashes(s: String): String =
+    s.replace("%2F", "/")
```
pshapiro4broad, line 16: I thought the algorithm would be more complex to support % quoting. Isn’t there a general method that does URL deciding.

charlesdyan: I was worried about cases like this: 
```
Uri("%2520").path.head shouldEqual "%20"
Uri("/%2F%5C").path shouldEqual Path / """/\"""
```
since what we have is a string that has already been Uri decoded (excepting forward slashes) once. I figured just decoding the slash would leave us with fewer unsupported cases than double decoding

pshapiro4broad (approved):

PR #312, opened 2019-11-26
PO-21742 Added ExternallyHosted document status

tlangs:
### Description

Ops has the need for an `ExternallyHosted` document status to track data that is not hosted by us, but we still want to know about. To mark files as `ExternallyHosted`, a note must be provided, and should ideally contain information about what organization is hosting the data. The command prefix for marking files as `ExternallyHosted` is `relinquish-` because `mark-externally-hosted-ubam` is really long and cumbersome. Open to other ideas for command name.

Both unit and integration tests have been added for all data types.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -249,11 +249,10 @@ class ClioWebClient(
     */
   def simpleQuery[I](clioIndex: ClioWebClient.QueryAux[I])(
     input: I,
-    includeDeleted: Boolean,
-    includeAll: Boolean
+    includeAllStatuses: Boolean
   ): Source[Json, NotUsed] = {
     import clioIndex.implicits._
-    if (includeDeleted || includeAll) {
+    if (includeAllStatuses) {
       query(clioIndex)(input.asJson)
     } else {
       query(clioIndex)(
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
@@ -249,10 +249,11 @@ class ClioWebClient(
     */
   def simpleQuery[I](clioIndex: ClioWebClient.QueryAux[I])(
     input: I,
-    includeDeleted: Boolean
+    includeDeleted: Boolean,
+    includeAll: Boolean
   ): Source[Json, NotUsed] = {
     import clioIndex.implicits._
-    if (includeDeleted) {
+    if (includeDeleted || includeAll) {
       query(clioIndex)(input.asJson)
     } else {
       query(clioIndex)(
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
@@ -375,7 +375,7 @@ trait ArraysTests { self: BaseIntegrationSpec =>
     testQueryAll(DocumentStatus.Deleted)
   }
 
-  it should "show relinquised arrays records on queryAll, but not query" in {
+  it should "show relinquished arrays records on queryAll, but not query" in {
     testQueryAll(DocumentStatus.ExternallyHosted)
   }
 
--- a/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
+++ b/clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
@@ -296,9 +296,9 @@ trait ArraysTests { self: BaseIntegrationSpec =>
 
   def testQueryAll(documentStatus: DocumentStatus): Future[Assertion] = {
     val queryArg = documentStatus match {
-      case DocumentStatus.Deleted          => "--include-deleted"
-      case DocumentStatus.ExternallyHosted => "--include-all"
-      case _                               => ""
+      case DocumentStatus.Deleted  => "--include-deleted"
+      case DocumentStatus.External => "--include-all"
+      case _                       => ""
     }
     val barcode = "barcode." + randomId
 
@@ -375,8 +375,8 @@ trait ArraysTests { self: BaseIntegrationSpec =>
     testQueryAll(DocumentStatus.Deleted)
   }
 
-  it should "show relinquished arrays records on queryAll, but not query" in {
-    testQueryAll(DocumentStatus.ExternallyHosted)
+  it should "show External arrays records on queryAll, but not query" in {
+    testQueryAll(DocumentStatus.External)
   }
 
   it should "respect user-set regulatory designation for arrays" in {
@@ -749,11 +749,11 @@ trait ArraysTests { self: BaseIntegrationSpec =>
     }
   }
 
-  def testRelinquishArrays(
+  def testExternalArrays(
     existingNote: Option[String] = None
   ): Future[Assertion] = {
-    val relinquishNote =
-      s"$randomId --- Relinquished by the integration tests --- $randomId"
+    val markExternalNote =
+      s"$randomId --- Marked External by the integration tests --- $randomId"
 
     val chipwellBarcode = s" chip$randomId"
     val version = 3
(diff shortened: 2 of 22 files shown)


clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
```diff
@@ -249,10 +249,11 @@ class ClioWebClient(
     */
   def simpleQuery[I](clioIndex: ClioWebClient.QueryAux[I])(
     input: I,
-    includeDeleted: Boolean
+    includeDeleted: Boolean,
+    includeAll: Boolean
   ): Source[Json, NotUsed] = {
     import clioIndex.implicits._
-    if (includeDeleted) {
+    if (includeDeleted || includeAll) {
```
charlesdyan: So is the idea here to keep includeDeleted to not break things that use it, but really it includes all records, so add an option that is semantically correct?

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -359,12 +365,20 @@ trait ArraysTests { self: BaseIntegrationSpec =>
...
         }
       }
     }
   }
 
+  it should "show deleted arrays records on queryAll, but not query" in {
+    testQueryAll(DocumentStatus.Deleted)
+  }
+
+  it should "show relinquised arrays records on queryAll, but not query" in {
```
charlesdyan: an 'h'

clio-integration-test/src/it/scala/org/broadinstitute/clio/integrationtest/tests/ArraysTests.scala
```diff
@@ -735,6 +749,116 @@ trait ArraysTests { self: BaseIntegrationSpec =>
...
+
+    val chipwellBarcode = s" chip$randomId"
+    val version = 3
+
+    val vcfContents = s"$randomId --- I am a vcf fated for other worlds --- $randomId"
+    val vcfIndexContents =
+      s"$randomId --- I am an index fated for other worlds --- $randomId"
+    val metrics1Contents = s"$randomId --- I am a questing metrics file --- $randomId"
+    val metrics2Contents =
+      s"$randomId --- I am a second questing metrics file --- $randomId"
```
charlesdyan, line 766: cute

charlesdyan (approved):
minor comments/qs. lgtm

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
tlangs: `includeDeleted` always included all results, not just `Normal` and `Deleted` ones. I'd like to replace it with `includeAll` because its much more descriptive of what actually happens, but I'm leaving `includeDeleted` in for compatibility reasons.

tlangs:
retest this please

pshapiro4broad:
Some higher level questions

- What is the requirement for this new document status? That is, what is the behavior that this flag provides that we don't have from other existing flags? I didn't see any behavior change in this checkin.
- Can documents be created with this new status? Or is it expected that all documents are created as `Normal` and then become external?
- Why is this done as a document status, instead of using `Location`? That seems like a more natural place to put this feature, and if the intent is to have this affect how operations work, using `Location` would do this already (although those code paths may be untested).
- If you _do_ decide to do this as a document status operation, it would be better to generalize the document status changing operation instead of duplicating the code that marks a document a deleted.

tlangs:
- This change is necessary because we want to be able to track if a document is hosted by an external provider after we process it. Right now, the only way to do so would be to mark it as `Deleted`, but that is confusing since we would assume the data is just gone instead of hosted somewhere else.
- We wouldn't be tracking data we don't generate in Clio, so all documents enter Clio with a `Normal` status. When we submit to an external provider, we would then update the status to be `ExternallyHosted`. 
- The `Location` field refers to the "Processing Location", like in the `METRICS.AGGREGATION` table. It wouldn't make sense to add `ExternallyHosted` there. Plus, `Location` is a key field, and we can't update key fields of a record.
- I'll look at refactoring the document status change operation.

tlangs:
I think Delete and Relinquish executors differ enough that shoehorning them into common code would be messier than the small amount of current duplication.

pshapiro4broad:
> This change is necessary because we want to be able to track if a document is hosted by an external provider after we process it. Right now, the only way to do so would be to mark it as Deleted, but that is confusing since we would assume the data is just gone instead of hosted somewhere else.

What do you mean by "track"? Are you saying that this status is being stored so some external query can retrieve it, but it has no effect on clio's operations? If a clio document is external, does that mean that its gs paths are still present but are no longer pointing at files?

clio-client/src/main/scala/org/broadinstitute/clio/client/webclient/ClioWebClient.scala
pshapiro4broad: If they have the same meaning, are both needed at the web API layer? I can see supporting both at the command line, but it seems redundant to have two web API options that have the same behavior. We would accept that "deleted" in this case means all non-Normal status.

You could keep both command line options for backwards compatibility with scripts, and map them to the same option as before.

tlangs:
Yes, the status is being stored in Clio so that we have a record of what happened to the file. As with deleting, the paths remain in the record, but the document status tells us that the files are no no longer at the paths.

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/Metadata.scala
```diff
@@ -34,6 +34,12 @@ trait Metadata[M <: Metadata[M]] { self: M =>
     */
   def markDeleted(deletionNote: String): M
 
+  /**
+    * return a copy of this object which has been marked as externally hosted,
+    * with the given notes appended.
+    */
+  def markExternallyHosted(relinquishNote: String): M
```
pshapiro4broad: Could this API be refactored to share code with `markDeleted`, e.g. `changeStatus(note: String, status: DocumentStatus)`?

clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/Metadata.scala
```diff
@@ -34,6 +34,12 @@ trait Metadata[M <: Metadata[M]] { self: M =>
...
+
   /**
     * Return a copy of this object with given document status.
     */
```
pshapiro4broad: While you're here, can you remove `documentStatusFieldName` below? It's wrong and is unused.

pshapiro4broad:
Regarding the command name, I think it's not obvious that you make an object external by relinquishing it. How about 

- make the document status `External` instead of `ExternallyHosted`
- make the command prefix `mark-external-`

tlangs:
Definitely can rename it. Charlotte and I chatted about it and had a hard time coming up with a good name. Naming is hard...

pshapiro4broad:
> Yes, the status is being stored in Clio so that we have a record of what happened to the file.

I understand, my question was who is asking clio about external records vs deleted records? What is the use case for that query? It sounds so far like this could be done by having a `mark-deleted` command that marked an object as deleted but didn't actually do any deletions (or, as an argument to `delete`, e.g. `--mark-only`.)

tlangs:
Ops needs to know if the data no longer exists, or if it is hosted by another service. The new document status allows us to programmatically get this information and is much clearer than making it a sub-status of Deleted.

pshapiro4broad:
> Ops needs to know if the data no longer exists, or if it is hosted by another service. The new document status allows us to programmatically get this information and is much clearer than making it a sub-status of Deleted.

It sounds like what's needed is something that stores the state of external deliveries, which is (conceptually at least) orthogonal to the deleted state of the record. I assume that we don't delete the data until after external delivery occurs, so there is a period of time where an object is both external and not deleted.

If the cloud data is deleted after a delivery, how is the deletion done, e.g., does it use clio delete?

tlangs:
Right now, we don't expect delivered (to Terra) data to be Deleted. Clio continues to track the files location into the workspace bucket. Once we submit data to an external hoster, we relinquish all control to them. If it gets deleted, Clio shouldn't care. In the future, we may use the `mark-external` command to also delete delete our copy of the data, but as of now, we aren't confident enough in our submissions process to actually delete any data.

pshapiro4broad (approved):

PR #313, opened 2019-12-04
PO-21741 track billing project

tlangs:
### Description

It is very useful for Ops to be able to track the billing project of delivered data. This will aid us in keeping costs in check and give us more visibility into who's actually paying for delivered data. The default billing project is "broad-genomics-data", which is what we currently deliver everything into for V2 deliveries.

----

* Review cycle:

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala
@@ -162,7 +162,7 @@ final case class DeleteCram(
 final case class DeliverCram(
   @Recurse key: CramKey,
   workspaceName: String,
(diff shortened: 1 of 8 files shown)


pshapiro4broad (approved):
This looks good but I wonder if there's any benefit to having a default, if it's the same for all records. Is there some use to having this set to the same string vs leaving it empty?

Also, it would be nice to have a test for this in one of the data delivery tests.

tlangs:
The default is nice because it means we can release clio without having to coordinate with updating our delivery pipelines. We need to patch all the current delivered records with "broad-genomics-data" so that we can then backfill the AnVIL stuff that was delivered to a different billing project.

tbl3rd (approved):
I like withWorkspaceName() -=> withWorkspace().

PR #316, opened 2020-01-03
GH-595: Update AkkaVersion to 2.5.20 to avoid TTL warning in response.

tbl3rd:
That is all.


jessicaway (approved):
Thanks for looking into this!

rexwangcc (approved):

PR #317, opened 2020-01-08
GH-597: An empty JSON response should be `[]` instead of ``.

tbl3rd:
https://broadinstitute.atlassian.net/browse/GH-597

### DIFF
--- a/clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
+++ b/clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
@@ -16,7 +16,8 @@ import org.broadinstitute.clio.client.util.IoUtil
 import org.broadinstitute.clio.client.webclient.ClioWebClient
 import org.broadinstitute.clio.util.auth.ClioCredentials
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext}
 import scala.util.{Failure, Success, Try}
 
 /**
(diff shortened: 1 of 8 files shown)


pshapiro4broad (dismissed):
A regression test would be nice, if it's not too difficult to add. Given the number of tests clio already has I was surprised that we didn't already have a test that covered this case.

tbl3rd:
I meant to open this as a draft. Sorry.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
```diff
@@ -64,26 +65,39 @@ object ClioClient extends LazyLogging {
...
+    }
+
+    def complete(done: Try[Done]): Unit = {
+      val status = done match {
+        case Success(_) => 0
+        case Failure(ex) =>
+          logger.error("Failed to execute command", ex)
+          1
+      }
+      Await.result(system.terminate(), 23.seconds)
```
pshapiro4broad, line 91: It looks like this is the only change here, is that correct?

Maybe a comment would be helpful, at least for the magic number.

pshapiro4broad (dismissed):
Looks good, as far as I can tell.

clio-client/src/main/scala/org/broadinstitute/clio/client/ClioClient.scala
tbl3rd: I thought the magic number was magic enough for its value to be obviously arbitrary, but OK.

rhiananthony (approved):
👍

PR #325, opened 2020-03-04
Fixing broken diffson validation and adding unit test

rhiananthony:
### Description

This work is to fix the breaking change introduced by the last merge to prod. Namely that there was a change in how diffson allowed (remembered) changes after doing diffs of two documents. This is in turn part of the effort for fixing GH-624, GH-625, and GH-626

----

* Review cycle:

### DIFF
--- a/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
+++ b/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
@@ -5,7 +5,8 @@ import akka.stream.scaladsl.Source
 import diffson._
 import diffson.circe._
 import diffson.jsonpatch.Replace
-import diffson.jsonpatch.simplediff._
+import diffson.jsonpatch.simplediff.remembering._
+import diffson.lcs.Patience
 import io.circe.{Json, JsonObject}
 import io.circe.syntax._
 import org.broadinstitute.clio.server.dataaccess.{PersistenceDAO, SearchDAO}
@@ -147,6 +148,7 @@ abstract class IndexService[CI <: ClioIndex](
     existingMetadata: clioIndex.MetadataType,
     newMetadata: clioIndex.MetadataType
   ): Either[Exception, clioIndex.MetadataType] = {
+    implicit val foo = new Patience[Json]
     val differences = diff(existingMetadata.asJson, newMetadata.asJson)
     /* Types of differences:
         Replace - the existing value is different from what we want to set. We ignore nulls. This will overwrite.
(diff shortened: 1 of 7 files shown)


clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
```diff
@@ -147,6 +148,7 @@ abstract class IndexService[CI <: ClioIndex](
     existingMetadata: clioIndex.MetadataType,
     newMetadata: clioIndex.MetadataType
   ): Either[Exception, clioIndex.MetadataType] = {
+    implicit val foo = new Patience[Json]
```
rexwangcc, line 151: Debugging line?

rexwangcc (approved):
I have 1 comment otherwise LGTM!

clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala
rhiananthony: I think that this is something which tells the test to wait for the completion of the task which returns the future. Dan suggested that it should be added and I'm not sure whether we can delete it or not. I'll test, and comment with the answer.

rhiananthony: It requires the Patience to be added, if not then the diff fails as this:
[error] /Users/ranthony/GitHub/Broad/clio/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala:152: could not find implicit value for parameter Diff: diffson.Diff[io.circe.Json,P]
[error]     val differences = diff(existingMetadata.asJson, newMetadata.asJson)
[error]                           ^
[error] /Users/ranthony/GitHub/Broad/clio/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala:9: Unused import
[error] import diffson.lcs.Patience
[error]                    ^
[error] /Users/ranthony/GitHub/Broad/clio/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala:152: local val differences in method validateUpsert is never used
[error]     val differences = diff(existingMetadata.asJson, newMetadata.asJson)
[error]         ^
[error] /Users/ranthony/GitHub/Broad/clio/clio-server/src/main/scala/org/broadinstitute/clio/server/service/IndexService.scala:148: parameter value existingMetadata in method validateUpsert is never used
[error]     existingMetadata: clioIndex.MetadataType,
[error]     ^

tbl3rd: Looks like an implicit argument to the following line `val differences = diff(…)`
to specify a "longest common subsequence" algorithm optimized for JSON
for use with the Circe library. I don't know why, but can demonstrate how to
get IntelliJ to show you what I mean.

That implicit was probably provided by one of the other imports before the
library update, but now we have to do it to avoid the `Unused import` error.

PR #326, opened 2020-03-22
Ra develop deploy clio300

rhiananthony:
### Description

The purpose of this change os to fix the deployment so that it will work with the new terraform module in the new network. Part of this work is changing the way that deployments were being done. Instead of having to build a fqdn from the host name and ssh into servers using the dsde-jenkins ssh key we now use a service account and the native gcloud commands for access to the instance. Instead of passing the instance as a parameter, we now leverage the ability for google to return us the instances with specified labels (in this case servers with the labels "clio" and "active"). 

This relates to ticket: https://broadinstitute.atlassian.net/browse/GH-683

QA for this would be to run a deployment and verify that it deploys as expected. Note the deployment and the code are linked, so you may need to build the container first using the clio-build job prior to running this one. 

----

* Review cycle:


rexwangcc (dismissed):
Looks good to me

jenkins/deploy-clio-server.sh
```diff
@@ -263,30 +314,48 @@ main() {
...
     exit 1
   fi
 
-  # Tag the current commit with the name of the environment that was just deployed to.
-  # This is a floating tag, so we have to use -f.
-  #
-  # NOTE: we can't push the tag here because Jenkins has to run this script with its
-  # GCE SSH key, which doesn't match its GitHub key. We push the tag as a follow-up
-  # step using the git publisher.
+#  # Tag the current commit with the name of the environment that was just deployed to.
```
pshapiro4broad: Did you mean to comment these lines twice?

jenkins/deploy-clio-server.sh
```diff
@@ -170,59 +154,126 @@ render_ctmpls() {
...
 }
 
-stop_containers() {
-  local -r clio_fqdn=$1
+activate_service_account() {
+  VAULT_TOKEN=$(cat /etc/vault-token-dsde)
+  SA_VAULT_PATH="secret/dsde/gotc/dev/common/ci-deployer-service-account.json"
+
+  docker run --rm -e VAULT_TOKEN="${VAULT_TOKEN}" \
+    broadinstitute/dsde-toolbox:dev vault read -format=json ${SA_VAULT_PATH} \
```
pshapiro4broad, line 163: I think `vault read -format=json ... | jq '.data'` is the same as `vault read -format=json -field=data ...`

jenkins/deploy-clio-server.sh
```diff
@@ -170,59 +154,126 @@ render_ctmpls() {
...
 
-  ssh ${SSH_OPTS[@]} ${SSH_USER}@${clio_fqdn} "${DOCKER_COMPOSE[*]} pull"
-  ssh ${SSH_OPTS[@]} ${SSH_USER}@${clio_fqdn} "${DOCKER_COMPOSE[*]} up -d"
+  # Get the instance marked with the Clio and Active labels (There should only be one)
+  CLIO_INSTANCE_LIST=$(gcloud --format="table[no-heading](Name)" compute \
+                         --project ${CLIO_PROJECT} instances list \
+                         --filter="labels.app=clio AND labels.state=active")
+
+  # Verify that there is only one instance
+  if [[ ${#CLIO_INSTANCE_LIST[@]} -ne 1 ]];
```
pshapiro4broad, line 179: if you end the line with a `;`, then you can put the `then` after it, e.g. `if [[ ... ]]; then`. If you want to put `then` on its own line, then you don't need a `;` since it's a command separator and there isn't a command after it. Personally I prefer `; then` but either is OK.

jenkins/deploy-clio-server.sh
```diff
@@ -170,59 +154,126 @@ render_ctmpls() {
...
+stop_clio_containers() {
+  CLIO_PROJECT=$1
+  CLIO_INSTANCE=$2
+
+  gcloud compute --project ${CLIO_PROJECT} \
+                 ssh ${CLIO_INSTANCE} \
+                 --zone=us-central1-a \
+                 --command="test -f ${COMPOSE_FILE} &&
+                              ${DOCKER_COMPOSE[*]} stop || echo &&
+                            test -f ${COMPOSE_FILE} &&
```
pshapiro4broad: Why is the second `test` here? If `COMPOSE_FILE` wasn't there when the first `test` runs, then this line wouldn't be executed. Or is the thinking that maybe `docker-compose stop` might delete the compose.yml file, so we need to check after running `stop`?

jenkins/deploy-clio-server.sh
rhiananthony: I think this may be true. Is there a reason to change it?

jenkins/deploy-clio-server.sh
rhiananthony: I know it's not needed, but the linter in my IDE liked it, so I left it.

jenkins/deploy-clio-server.sh
rhiananthony: This was a copy-paste error. It was originally 2 different commands and they got merged into one. I removed the second test.

jenkins/deploy-clio-server.sh
rhiananthony: I tried to use this and got the following error: "Field data not present in secret". It appears that it can get the fields inside of the json object's "data" field, but not the values of the json object itself, which includes metadata for the secret

pshapiro4broad (approved):
Don't forget to squash and merge!

pshapiro4broad:
The pullapprove error can be ignored

PR #330, opened 2020-04-13
updated SSL ctmpls for new paths of new ssl certs

hjfbynara:
### Description

Update the clio server configuration to support the new SSL wildcard certificates.  As part of this change we are moving away from a generic vault path for SSL certs and to a more explicit path so that it is obvious what is being used and allow the migration to new certs "app" at a time as opposed to all at once for all apps using a common vault path.

In the long run - when clio is either migrated to a GCE load balancer or more ideally to Kubernetes - the method of this configuration will be significantly different.

----

* Review cycle:


rhiananthony (approved):
👍

clio-server/config/server.crt.ctmpl
```diff
@@ -1 +1 @@
-{{with $environment := env "ENV"}}{{with $commonSsl := vault (printf "secret/dsde/gotc/%s/common/server.crt" $environment)}}{{$commonSsl.Data.server_crt}}{{end}}{{end}}
+{{with $environment := env "ENV"}}{{if eq $environment "dev"}}{{with $commonSsl := vault (printf "secret/dsp/certs/wildcard.gotc-dev.broadinstitute.org/20220403/server.crt")}}{{$commonSsl.Data.value}}{{end}}{{end}}{{if eq $environment "staging"}}{{with $commonSsl := vault (printf "secret/dsp/certs/wildcard.gotc-staging.broadinstitute.org/20220408/server.crt")}}{{$commonSsl.Data.value}}{{end}}{{end}}{{if eq $environment "prod"}}{{with $commonSsl := vault (printf "secret/dsp/certs/wildcard.gotc-prod.broadinstitute.org/20220408/server.crt")}}{{$commonSsl.Data.value}}{{end}}{{end}}{{end}}
```
pshapiro4broad, line 1: If I'm reading this correctly, it looks like this adds a datestamp to the vault path. This seems like a bad idea, as it now requires manual updates every time a new SSL cert is pushed. Was there a reason why the date was added?

pshapiro4broad (approved):

clio-server/config/server.crt.ctmpl
tbl3rd: Maybe …
```
{{with $environment := env "ENV"}}
    {{if eq $environment "dev"}}
        {{with $commonSsl := vault (printf "secret/dsp/certs/wildcard.gotc-dev.broadinstitute.org/20220403/server.crt")}}
            {{$commonSsl.Data.value}}
        {{end}}
    {{end}}
    {{if eq $environment "staging"}}
        {{with $commonSsl := vault (printf "secret/dsp/certs/wildcard.gotc-staging.broadinstitute.org/20220408/server.crt")}}
            {{$commonSsl.Data.value}}
        {{end}}
    {{end}}
    {{if eq $environment "prod"}}
        {{with $commonSsl := vault (printf "secret/dsp/certs/wildcard.gotc-prod.broadinstitute.org/20220408/server.crt")}}
            {{$commonSsl.Data.value}}
        {{end}}
    {{end}}
{{end}}
```

tbl3rd: Why is the first number `20220403` different?

tbl3rd: It looks like this just maps `$environment` to 3 different certs?

hjfbynara: you are correct that each time the SSL cert is renewed this config file will need updating and a re-deployment will need to happen.  Of course whenever a SSL cert is updated a redeployment will be necessary.  Since a SSL cert is only good for so long - the timestamp is the date of expiration.  Using a timestamp path allows the renewal process to happen completely independent of the various applications adopting the updated cert.   In one sense trying to prevent changes from occurring underneath the covers on applications.  Then the applications can decide at what point they want to adopt the new cert - to fit into their own development/deployment cycle.  With the caveat that it must be done prior to current cert expiration.  

With helm (under k8s) these type of changes are much easier - it would just be running: helm upgrde -f new-values.yaml.  -  Where the new-values.yaml file would include the reference to the new updated (timestamped) cert.  This would update the appropriate references without really needing a full redeployment - since swapping out SSL certs with Ingress configurations under k8s - is a no downtime process.

So I opted for a process that allows the admins to renew certs whenever they need to with the knowledge that deploying the updated certs will have zero impact on any application consuming them.  And leaving the dev teams the responsibility of migrating to the new SSL certs within their own cycle.  Again as long as it is prior to current cert expiration.  Sort of like dev teams decide when to accept patches and updated library versions.

hjfbynara: The main reason I put everything on a few lines has to do with an artifact of ctmpl language.  It adds blank lines for every ctmpl language line it processes.  putting them on a single line means there will not be any errant blank lines.  Newer versions of ctmpl does not have this issue - but we can't at this time adopt the newer version due to breaking changes that were introduced as part of the upgrade.  So its a balance between code readability and introducing blank lines in rendered file - which sometimes breaks things.

hjfbynara: > Why is the first number `20220403` different?

the dev cert was renewed prior to the other environment - so the expiration date is different.    the YYYYMMDD is the expiration date of that cert.

hjfbynara: > It looks like this just maps `$environment` to 3 different certs?

Yes 3 diff environments.  overall the setup we have does not make clio's deployment methodology as clean.   but tries to balance easy management of cert renewal by the admins with the future k8s deployment methodology.  Sacrificing some things on the legacy deployment methodologies like current cromwell and clio.

pshapiro4broad: Thanks for the explanation! 👍

PR #331, opened 2020-05-01
Svc account needed to be specific to environment.

rhiananthony:
### Description

The service account which is activated is environment dependant. There is a non prod and a prod svc account. This fixes the activation of that svc account, which in turn allows deployments. 

----

* Review cycle:

### DIFF
--- a/jenkins/deploy-clio-server.sh
+++ b/jenkins/deploy-clio-server.sh
@@ -156,8 +156,10 @@ render_ctmpls() {
 }
 
 activate_service_account() {
+  ENV=${1}
+
   VAULT_TOKEN=$(cat /etc/vault-token-dsde)
-  SA_VAULT_PATH="secret/dsde/gotc/dev/common/ci-deployer-service-account.json"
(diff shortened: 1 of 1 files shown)


hjfbynara (approved):

jenkins/deploy-clio-server.sh
```diff
@@ -156,8 +156,11 @@ render_ctmpls() {
 }
 
 activate_service_account() {
+  ENV=${1}
```
pshapiro4broad, line 159: `ENV` is already a bound environment variable, so this (I think) will overwrite the environment variable with the parameter (which is the same). So this shouldn't have any effect and the script should work the same if it's removed.

jenkins/deploy-clio-server.sh
```diff
@@ -156,8 +156,11 @@ render_ctmpls() {
...
+
+  echo "Activating Service Account"
```
pshapiro4broad, line 161: quotes have no effect here (but people seem to like them)
```suggestion
  echo Activating Service Account
```

pshapiro4broad (approved):
A few minor comments

PR #336, opened 2020-06-08
Correct behavior for Arrays Undelivery

tlangs:
### Description

For Arrays, we need to distinguish an undelivery from a delivery because of the differences in how we store idats in prod storage and delivered buckets. We had to thread an `undeliver` boolean through a lot of commands just to get this done. This still feels like a better solution rather than imputing whether a delivery is actually an undelivery based on paths that may or may not exist.

----

* Review cycle:


clio-client/src/main/scala/org/broadinstitute/clio/client/metadata/ArrayDeliverer.scala
```diff
@@ -8,11 +8,16 @@ class ArrayDeliverer extends MetadataMover[ArraysMetadata] {
...
     destination: URI,
-    newBasename: Option[String]
+    newBasename: Option[String],
+    undeliver: Boolean
   ): (ArraysMetadata, Iterable[IoOp]) = {
     import MetadataMover.buildFilePath
 
-    val idatDestination = destination.resolve(ArrayDeliverer.IdatsDir)
+    val idatDestination = if (undeliver) {
+      destination.resolve("../").resolve(ArrayDeliverer.IdatsDir)
```
pshapiro4broad: Instead of doing it this way, how about making `IdatsDir` overridable and creating `ArrayUndeliverer` which subclasses `ArrayDeliverer` and overrides this field? That would be a smaller change and limit the scope to `UndeliverArrays`.

https://github.com/broadinstitute/clio/blob/develop/clio-client/src/main/scala/org/broadinstitute/clio/client/commands/ClioCommand.scala#L390

pshapiro4broad (changes requested):

gbggrant (approved):

clio-client/src/main/scala/org/broadinstitute/clio/client/dispatch/DeliverExecutor.scala
```diff
@@ -46,9 +46,10 @@ class DeliverExecutor[CI <: DeliverableIndex](
   }
 
   override protected[dispatch] def buildMove(
-    metadata: moveCommand.index.MetadataType
+    metadata: moveCommand.index.MetadataType,
+    undeliver: Boolean = false
   ): Source[(moveCommand.index.MetadataType, immutable.Seq[IoOp]), NotUsed] = {
-    super.buildMove(metadata).map {
+    super.buildMove(metadata, undeliver).map {
```
pshapiro4broad: If this works, then you should be able to back this change out

tlangs: Yup, you're completely right. Clio is confusing when you're deep in it...

pshapiro4broad (approved):
Looks good!

PR #342, opened 2020-12-03
POD-837 Added insert_size_metrics_path to BAM index

tlangs:
### Description

Adding `insert_size_metrics_path` to the Clio BAM index.

----

* Review cycle:

### DIFF
--- a/clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/bam/BamMetadata.scala
+++ b/clio-transfer-model/src/main/scala/org/broadinstitute/clio/transfer/model/bam/BamMetadata.scala
@@ -11,6 +11,7 @@ case class BamMetadata(
   bamSize: Option[Long] = None,
   bamPath: Option[URI] = None,
   baiPath: Option[URI] = None,
+  insertSizeMetricsPath: Option[URI] = None,
   workspaceName: Option[String] = None,
   billingProject: Option[String] = None,
   notes: Option[String] = None,


gordonwade (approved):

tbl3rd (approved):
Do we want to support move-bam or deliver-bam?

tlangs:
We already do.

pshapiro4broad:
> > Do we want to support move-bam or deliver-bam?

> We already do.

@tlangs I don't see what you're referring to. This change adds a new URI field but makes no change to `BamDeliverer` or `pathsToDelete`. So it looks like this field isn't yet supported by deliver, move, delete or undeliver.

tbl3rd:
> So it looks like this field isn't yet supported by deliver, move, delete or undeliver.

Thanks. That was my point.

tlangs:
@pshapiro4broad BAMs in Clio closely mirror the CRAM implementation. For deletion, we don't delete CRAM metrics files, only sequence data. So, we will also not delete BAM metrics files. As for moving, there is no operational requirement to move the `insert_size_metrics_file`. This metrics file is being added solely to get BAM support in EDDy, and all other uses of the BAM index don't care about the metrics file.

pshapiro4broad:
> @pshapiro4broad BAMs in Clio closely mirror the CRAM implementation. For deletion, we don't delete CRAM metrics files, only sequence data. So, we will also not delete BAM metrics files. As for moving, there is no operational requirement to move the insert_size_metrics_file. This metrics file is being added solely to get BAM support in EDDy, and all other uses of the BAM index don't care about the metrics file.

Ok, that's good to know. This is different from metrics files on other objects, such as Crams, where (I believe) we do something with every URI in the metadata. It may be worth adding a comment to the code to make this clear for a future maintainer.

On the other hand, if there's code in EDDy that's specific to a field in clio, maybe that's something that should be done in a clio-client command so EDDy doesn't need to be updated when a field in clio changes.

